PR #2, opened 2018-05-15
Add tests for components and actions

georgebabey:
Gotta start somewhere...

### DIFF
--- a/src/components/Courseware/CoursewareNav/CoursewareNav.mocks.js
+++ b/src/components/Courseware/CoursewareNav/CoursewareNav.mocks.js
@@ -1,0 +1,135 @@
+const mockCourseOutline = {
+  displayName: 'Demo Course',
+  type: 'course',
+  id: 'DemoCourse',
+  descendants: [
+    {
+      displayName: 'Chapter 1',
+      type: 'chapter',
+      id: 'c1',
+      descendants: [
+        {
+          displayName: 'Section 1.1',
+          type: 'sectional',
+          id: 's1.1',
+          descendants: [
+            {
+              displayName: 'Vertical 1.1.1',
+              type: 'vertical',
+              id: 'v1.1.1',
+              descendants: [],
+            },
+            {
+              displayName: 'Vertical 1.1.2',
+              type: 'vertical',
+              id: 'v1.1.2',
+              descendants: [],
+            },
+            {
+              displayName: 'Vertical 1.1.3',
+              type: 'vertical',
+              id: 'v1.1.3',
+              descendants: [],
+            },
+            {
+              displayName: 'Vertical 1.1.4',
+              type: 'vertical',
+              id: 'v1.1.4',
+              descendants: [],
+            },
+          ],
+        },
+        {
+          displayName: 'Section 1.2',
+          type: 'sectional',
+          id: 's1.2',
+          descendants: [
+            {
+              displayName: 'Vertical 1.2.1',
+              type: 'vertical',
+              id: 'v1.2.1',
+              descendants: [],
+            },
+          ],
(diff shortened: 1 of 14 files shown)


src/components/Courseware/CoursewareNav/CoursewareNav.mocks.js
```diff
@@ -0,0 +1,135 @@
...
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
+
+/* eslint-disable import/prefer-default-export, object-curly-newline */
+export {
```
adamstankiewicz: nit: to avoid having to disable `eslint` here, we could do `export const mockCourseOutline = { ... }` above instead.

src/components/Courseware/CoursewareNav/index.jsx
```diff
@@ -46,8 +46,18 @@ class CoursewareNav extends React.Component {
...
+  id: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  descendants: PropTypes.array.isRequired,
+});
+
 CoursewareNav.propTypes = {
-  courseOutline: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+  courseOutline: PropTypes.shape({
+    displayName: PropTypes.string.isRequired,
+    descendants: PropTypes.arrayOf(nodeShape).isRequired,
```
adamstankiewicz: Is there ever a case where the `courseOutline` prop will have more than one level of `descendants`? If that is possible, can this `courseOutline` prop handle multiple levels?

src/data/actions/index.test.js
```diff
@@ -1,65 +1,104 @@
...
+          'section-1': {
+            id: 'section-1',
+            display_name: 'Section 1',
+            student_view_url: 'http://www.example.com/chapter1/section2',
+            type: 'Section',
+            descendants: [],
+          },
+        },
+      };
+      fetchMock.getOnce(`http://localhost:18000/api/courses/v1/blocks/?course_id=${encodeURIComponent(courseId)}&username=staff&depth=all&nav_depth=3&block_types_filter=course,chapter,sequential,vertical`, {
```
adamstankiewicz: nit: For long URLs with params like this, it might be cleaner to use something like [`querystring`](https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options) to represent the params as an object instead (also would apply to the LMS API request in the `CourseOutline` action).

src/components/Courseware/CoursewareNav/CoursewareNav.mocks.js
georgebabey: that would only eliminate the `object-curly-newline`. It really wants `export default`, but I know we're going to add additional mocks so I left it as-is and once we have another mock we can remove the eslint-disables

src/components/Courseware/CoursewareNav/index.jsx
georgebabey: Yes, that is supported because `nodeShape` also contains descendants. 

Ideally `nodeShape.descendants` would be an `arrayOf(nodeShape)`  Not able to do that in a single declaration above but could do it in 2 lines. Opted to leave it as a generic array because I got scared off by the comment recursive propTypes could be "[arbitrarily slow](https://github.com/facebook/react/issues/5676)"  and didn't see a lot of value but if we think that level of checking is necessary I can update.

src/data/actions/index.test.js
georgebabey: 👍 This will eventually (soon) move into a constant, lets make sure to use `querystring` when it does

PR #3, opened 2018-05-15
Previous/Next buttons!

brittneyexline:
This implements the previous and next buttons for courseware navigation.


src/components/Courseware/CoursewareContent/VerticalNav/index.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Button } from '@edx/paragon';
+
+
+const VerticalNav = function VerticalNav(props) {
+  return (
+    <div>
+      <Link to={{
```
georgebabey: Dug in a little - while this works it is semantically wrong as a `<Link>` becomes an anchor and anchors shouldn't contain buttons...

Will also have accessibility issues as you'd tab to the anchor first, then the button. Instead we can use the withRouter HOC to get history, and then use `history.push` and an onClick for the button...

See: https://stackoverflow.com/questions/45666288/how-to-use-reacts-router-v4-history-push

src/data/actions/index.js
```diff
@@ -21,12 +21,28 @@ const createTreeNode = (node, blocks) => (
...
+  }
+
+  return node.descendants.reduce(reducer, []);
+}
 
 const buildOutlineTree = (blockData) => {
   const rootBlock = blockData.blocks[blockData.root];
-  const outline = createTreeNode(rootBlock, blockData.blocks);
+  let outline = createTreeNode(rootBlock, blockData.blocks);
+  outline['verticalNodeList'] = getVerticalNodeList(rootBlock, blockData.blocks);
```
georgebabey: I'd move this out of the `buildOutlineTree` method and into its own method that we can call from the action block

PR #8, opened 2018-06-05
WIP enterprise data api to /admin page for course enrollments

adamstankiewicz:
Adds an `/admin` page that pulls in course enrollment data for a given enterprise and displays the results in a table:

image

**Notes**
1. Still need to add tests before this can be merged.
2. The API url for the `edx-analytics-data-api` is hardcoded with a specific enterprise UUID (the one used in the `enterprise_data` fixtures) and a JWT token.
3. We might want to discuss what data we should include in the table in the UI. For now, I pulled in things that seem like should be in there.


src/containers/App/index.jsx
```diff
@@ -23,6 +24,7 @@ export default function App() {
         <Route exact path="/" component={HomePage} />
         <Route path="/courses/:courseId" component={CoursewarePage} />
         <Route path="/faq" component={FAQSupportPage} />
+        <Route path="/admin" component={CourseEnrollmentsPage} />
```
georgebabey: Should probably be called the AdminPage, with the CourseEnrollmentsContainer as an element of the page? Just thinking the administrative section may grow and it would be good to have that separation defined up front.

src/containers/CourseEnrollmentsPage/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+      ...enrollment,
+      enrollment_created_timestamp: this.formatTimestamp(enrollment.enrollment_created_timestamp),
+      passed_timestamp: this.formatTimestamp(enrollment.passed_timestamp),
+      course_duration_weeks: `${enrollment.course_duration_weeks} weeks`,
+    }));
+  }
+
+  render() {
+    return (
+      <div className="container">
```
georgebabey: Should break this out into a dumb component

src/containers/App/index.jsx
adamstankiewicz: My thought process was that because, for now, the admin section is only enrollments, the `CourseEnrollmentsPage` would be the container for the `/admin` route. However, when we have more than one admin page, the route for the `CourseEnrollmentsPage` would change to be `/admin/enrollments`. I was thinking we'd likely create an `AdminPage` as a dashboard type of page that displays important info and links to the specific admin pages, like `CourseEnrollmentsPage`.

The suggestion you made above seems to make most sense if we are likely going to have a single administrative route (i.e., `/admin`) that contains all of the admin tables/visualization on a single page. With that, it would make sense to rename `CourseEnrollmentsPage` to `CourseEnrollmentsContainer` and move it into a new `AdminPage` container like you suggested.

Maybe this can be a talking point at the upcoming design session with product/UX this week to determine how the information architecture of the administrative section will be structured/organized?

georgebabey: 👍  I'm fine withe keeping this as is, we can always change it in the future. I was thinking a default `/admin` route but maybe that is just `/admin/dashboard`  Its too early to say for sure so lets just merge and move forward.

PR #10, opened 2018-06-19
Added Login Page w/ authentication

jonalexvasquez:
Made changes to the Login Component, going to add tests, and make a change to the logic of displaying footer/header once George's brand branch merges


src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+        <Fieldset
+          legend={<font color="grey">Login to edX from Enterprise</font>}
+          invalidMessage="This is invalid!"
+          variant={{ status: 'DANGER' }}
+          variantIconDescription="Error"
+        >
+          <InputText name="email" value={this.state.email} label="Email" type="text" onChange={(event) => { this.handleEmailChange(event); }} />
+          <InputText name="password" value={this.state.password} label="Password" type="password" onChange={(event) => { this.handlePasswordChange(event); }} />
+          <div className="form-addons">
+            <div className="submit-button">
```
adamstankiewicz: nit: Is the `submit-button` class name used anywhere else? If not, you could probably remove it.

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+      password: '',
+    };
+    this.handleEmailChange = this.handleEmailChange.bind(this);
+    this.handlePasswordChange = this.handlePasswordChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleEmailChange(event) {
+    this.setState({
+      email: event.trim(),
```
adamstankiewicz: Nice use of `.trim()`!

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+          legend={<font color="grey">Login to edX from Enterprise</font>}
+          invalidMessage="This is invalid!"
+          variant={{ status: 'DANGER' }}
+          variantIconDescription="Error"
+        >
+          <InputText name="email" value={this.state.email} label="Email" type="text" onChange={(event) => { this.handleEmailChange(event); }} />
+          <InputText name="password" value={this.state.password} label="Password" type="password" onChange={(event) => { this.handlePasswordChange(event); }} />
+          <div className="form-addons">
+            <div className="submit-button">
+              <Button label="Submit" onClick={this.handleSubmit} />
```
adamstankiewicz: nit: The UI mocks show the button having a label of `Sign In`.

src/containers/LoginPage/index.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import LoginForm from '../../components/LoginForm';
+import './LoginPage.scss';
+
+const LoginPage = () => (
+  <div className="container-fluid">
+    <h1> This is the Login Page </h1>
+    <p>Our company is great, here is more info about us</p>
+    <div className="form-container">
+      <LoginForm />
```
adamstankiewicz: We could probably get away with only including the login form on the page, horizontally centered using Bootstrap (i.e., disregard the messaging on the left) since it's relatively unclear what messaging will go there for now.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,64 @@
...
+        callback(new Error(error))
+      ));
+  }, 2000);
+};
+
+const login = (email, password) => (
+  (dispatch) => {
+    dispatch(setLoginPending(true));
+    // follow authentication process for this
+    authenticate(email, password, (error) => {
```
adamstankiewicz: Not really necessary to use a separate method `authenticate` with a callback here to make this work. Refer to how the [`fetchCourseEnrollments` ](https://github.com/edx/edx-portal/blob/master/src/data/actions/courseEnrollments.js#L18) action does this.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,64 @@
...
+        dispatch(setLoginSuccess(true));
+      } else {
+        dispatch(setLoginSuccess(false));
+        dispatch(setLoginError(error));
+      }
+    });
+  }
+);
+
+
```
adamstankiewicz: nit: Unnecessary line break.

src/containers/App/index.jsx
```diff
@@ -24,6 +26,7 @@ const App = () => (
       <Route path="/courses/:courseId" component={CoursewarePage} />
       <Route path="/faq" component={FAQSupportPage} />
       <Route path="/admin" component={CourseEnrollmentsPage} />
+      <Route path="/login" component={LoginPage} />
```
georgebabey: We need to think about if this should live outside of `<App />` since the enterprise branding in the header/footer are here. Lets sync up on this to make sure we don't clobber this PR with the branding changes in https://github.com/edx/edx-portal/tree/feature/branding

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+import {
+  LOGIN_FAILURE,
+  LOGIN_REQUEST,
+  LOGIN_SUCCESS,
+} from '../constants/loginForm';
+
+const setLoginPending = loading => ({
+  type: LOGIN_REQUEST,
+  payload: { loading },
```
adamstankiewicz: You don't need to pass the `loading` variable into this function as the `loading: true` can be set within the reducer for the `LOGIN_REQUEST` action.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+      } else {
+        dispatch(setLoginSuccess(false));
+        dispatch(seterror(error.message));
+      }
+    });
+  }
+);
+
+export default login;
+
```
adamstankiewicz: nit: Extra line

src/data/actions/loginForm.js
adamstankiewicz: @jonalexvasquez ^^ Let's move the `authenticate` method to the API services layer like `fetchCourseEnrollments` does (see above link).

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+const setLoginPending = loading => ({
+  type: LOGIN_REQUEST,
+  payload: { loading },
+});
+
+const setLoginSuccess = isLoginSuccess => ({
+  type: LOGIN_SUCCESS,
+  payload: { isLoginSuccess },
+});
```
adamstankiewicz: You don't need to pass the `isLoginSuccess` variable into this function as `isLoginSuccess: true` can be set within the reducer for the `LOGIN_SUCCESS` action.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+  type: LOGIN_FAILURE,
+  payload: { error },
+});
+
+const authenticate = (email, passphrase, callback) => {
+  const dataToSend = {
+    grant_type: 'password',
+    username: email,
+    password: passphrase,
+    client_id: 'c8OHQhvyy1paUgrmVyNX18lC2AkgXPQdIF5xE7EY',
```
adamstankiewicz: nit: Maybe add a comment explaining where the `client_id` value is coming from?

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+    // follow authentication process for this
+    authenticate(email, password, (error) => {
+      dispatch(setLoginPending(false));
+      if (!error) {
+        dispatch(setLoginSuccess(true));
+        dispatch(seterror(''));
+      } else {
+        dispatch(setLoginSuccess(false));
+        dispatch(seterror(error.message));
+      }
```
adamstankiewicz: This logic can be cleaned up a bit once you refactor to use the API service layer. For example, you should only dispatch `setLoginSuccess` if the request was successful. Here, you're dispatching even if there was an error. Likewise for dispatching the `seterror` function. Refer to the actions/courseEnrollments.js file for an example.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
+import axios from 'axios';
```
adamstankiewicz: nit: I think it's a little cleaner if the functions (e.g., `setLoginPending`, `setLoginSuccess`, `seterror`) are more in line with the action types (e.g., `LOGIN_REQUEST`, `LOGIN_SUCCESS`, `LOGIN_FAILURE`) kind of like how we're doing it in `actions/courseEnrollments.js`.  Maybe `fetchLoginRequest`, `fetchLoginSuccess`, `fetchLoginFailure`.

nit: You could also rename the actions from `LOGIN_REQUEST` -> `FETCH_LOGIN_REQUEST` to be consistent with the other action names in the app.

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,63 @@
...
+const seterror = error => ({
+  type: LOGIN_FAILURE,
+  payload: { error },
+});
+
+const authenticate = (email, passphrase, callback) => {
+  const dataToSend = {
+    grant_type: 'password',
+    username: email,
+    password: passphrase,
```
adamstankiewicz: nit: Maybe rename `passphrase` to `password`.

src/data/reducers/loginForm.js
```diff
@@ -0,0 +1,35 @@
+import {
+  LOGIN_FAILURE,
+  LOGIN_REQUEST,
+  LOGIN_SUCCESS,
+} from '../constants/loginForm';
+
+const loginForm = (state = {
+  isAuthenticated: localStorage.getItem('access_token') !== false,
```
adamstankiewicz: In Javascript, the `===` (versus `==`) looks for the same type. In this case, your logic is assuming that `localStorage.getItem('access_token')` will return a boolean if access_token does not exist. If `access_token` doesn't exist, it will return `null`. Even though `null` is falsey, it is not explicitly boolean so `localStorage.getItem('access_token') !== false` will actually return `true` even if `access_token` does not exist.

You could use `!!localStorage.getItem('access_token')` instead, which will force the resulting value to be boolean.

src/data/reducers/loginForm.js
```diff
@@ -0,0 +1,35 @@
...
+  isLoginSuccess: false,
```
adamstankiewicz: I think having just `isAuthenticated` is fine (no need for `isLoginSuccess` as you're setting both keys to the same value anyways).

src/data/reducers/loginForm.js
```diff
@@ -0,0 +1,35 @@
...
+  loading: false,
+  error: '',
```
adamstankiewicz: nit: `error` should be `null` instead of an empty string.

src/data/reducers/loginForm.js
```diff
@@ -0,0 +1,35 @@
...
+}, action) => {
+  switch (action.type) {
+    case LOGIN_REQUEST:
+      return {
+        ...state,
+        loading: action.payload.loading,
```
adamstankiewicz: You can just set `loading: true,` here instead of passing in a boolean through the action payload (see the `courseEnrollments.js` reducer).

src/data/reducers/loginForm.js
```diff
@@ -0,0 +1,35 @@
...
+      };
+    case LOGIN_SUCCESS:
+      return {
+        ...state,
+        isLoginSuccess: action.payload.isLoginSuccess,
+        isAuthenticated: action.payload.isLoginSuccess,
```
adamstankiewicz: Same as above; you could just set `isAuthenticated: true,` here.

src/index.jsx
```diff
@@ -17,7 +18,10 @@ const AppWrapper = () => (
     <ConnectedRouter history={history}>
       <CookiesProvider>
         <Router>
-          <App />
+          <Switch>
+            <Route path="/login" component={LoginPage} />
+            <App />
```
adamstankiewicz: I believe `<App />` should have a wildcard path here (i.e., `<Route path="" component={App} />`) as I believe `App` would otherwise be rendered regardless of the route.

src/index.jsx
```diff
@@ -17,7 +18,10 @@ const AppWrapper = () => (
     <ConnectedRouter history={history}>
       <CookiesProvider>
         <Router>
-          <App />
+          <Switch>
```
adamstankiewicz: As I mentioned in another comment, I think it makes sense to move the `Header` and `Footer` into this component under `Router`. Right now, you're using the `Footer` within the `LoginPage` component. The `LoginPage` component should only know about things related to the login page (i.e., only the form).

If you move the `Header` and `Footer` into `AppWrapper` you'd get the header and footer for free on the login page.

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+  redirectToAdminPage() {
+    this.props.history.push('/admin');
+  }
+
+  render() {
+    const { loading, error, isAuthenticated } = this.props;
+    if (isAuthenticated) { this.redirectToAdminPage(); }
+    return (
+      <div className="container">
+        <div className="row justify-content-center">
```
adamstankiewicz: You can update the page title by using the `Helmet` component nested under the `.container` div:
```jsx
<Helmet>
  <title>Login</title>
</Helmet>
```

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+          <div className="col-4">
+            <form onSubmit={this.handleSubmit}>
```
adamstankiewicz: It looks like you're calling `this.handleSubmit` twice, once here `onSubmit` and once `onClick` of the "Sign In" button. Is there a reason for that?

src/components/Footer/index.jsx
```diff
@@ -15,9 +16,11 @@ class Header extends React.Component { // eslint-disable-line react/prefer-state
             <Link className="logo" to="/">
               <Img src={EdxLogo} alt="edX logo" />
             </Link>
-            <Link className="logo" to="/">
-              <Img src={EnterpriseLogo} alt="Enterprise logo" />
-            </Link>
+            {(this.props.location.pathname.indexOf('/login') === -1) &&
```
georgebabey: I'd rather see this conditional based on the `props.logo` passed in rather than making the footer know about the routes. This just feels like too much knowledge for a footer.   Note: the image will come from props when the feature/branding branch lands: https://github.com/edx/edx-portal/compare/feature/branding#diff-60564e5c08a485e77598ff91f67eb3d3R18

src/components/Footer/index.jsx
```diff
@@ -1,12 +1,13 @@
...
+import PropTypes from 'prop-types';
+import { Link, withRouter } from 'react-router-dom';
 
 import Img from '../Img';
 import EdxLogo from '../../images/edx-logo.png';
 import EnterpriseLogo from '../../images/enterprise-logo.png';
 import './Footer.scss';
 
-class Header extends React.Component { // eslint-disable-line react/prefer-stateless-function
+class Footer extends React.Component { // eslint-disable-line react/prefer-stateless-function
```
georgebabey: 👍 good spot

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  }
+
+  handleSubmit(event) {
+    event.preventDefault();
+    const { email, password } = this.state;
+    // used to send email and pass to api
+    this.props.login(email, password);
+    // clear form
+    this.state = {
```
georgebabey: use `setState` here

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    const { loading, error, isAuthenticated } = this.props;
+    if (isAuthenticated) { this.redirectToAdminPage(); }
+    return (
+      <div className="container">
+        <div className="row justify-content-center">
+          <div className="col-4">
+            <form onSubmit={this.handleSubmit}>
+              <Fieldset
+                legend={<p className="text-secondary">Login to edX from Enterprise</p>}
+                invalidMessage="Invalid username and/or password!"
```
georgebabey: We should get the actual copy for the error messages (from UX?).  For example the `!` should likely be changed here

src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    };
+  }
+
+  redirectToAdminPage() {
+    this.props.history.push('/admin');
+  }
+
+  render() {
+    const { loading, error, isAuthenticated } = this.props;
+    if (isAuthenticated) { this.redirectToAdminPage(); }
```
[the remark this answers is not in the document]
jonalexvasquez: I agree that we should make this check in the constructor but won't we have to keep this in the render function as well? After the user logs in `isAuthenticated` updates and they should be redirected to the admin page.

src/data/reducers/loginForm.js
jonalexvasquez: If loading is set to true then how will the spinner stop? I believe it should stay like this unless I'm missing something

src/data/actions/loginForm.js
jonalexvasquez: My reasoning behind this was to make debugging more verbose. If there was no error I believe we should dispatch that it was successful and dispatch a `null` error. If a user fails to login the first time, and successfully the second time then the error from the first attempt will remain. I agree that we should only dispatch only `seterror` if there was an error found though.

adamstankiewicz: @jonalexvasquez That logic can be handled by the reducer without needing to dispatch `LOGIN_SUCCESS` and `LOGIN_REQUEST` actions that take parameters. On `LOGIN_REQUEST`, you can reset `error` to `null`. See the [course enrollments reducer](https://github.com/edx/edx-portal/blob/master/src/data/reducers/courseEnrollments.js#L15) for an example.

src/data/reducers/loginForm.js
adamstankiewicz: @jonalexvasquez On `LOGIN_REQUEST`, loading will always be true. Then, in the reducers for `LOGIN_SUCCESS` and `LOGIN_FAILURE` actions, you would set `loading: false`.

src/containers/Header/index.jsx
```diff
@@ -21,15 +21,17 @@ const Header = (props) => {
...
-        <Dropdown
-          title="edx@example.com"
-          menuItems={[
-            {
-              label: 'Logout',
-              href: '/',
-            },
-          ]}
-        />
+        {props.location.pathname.indexOf('/login') === -1 &&
```
georgebabey: This might look cleaner by using `mapStateToProps` to pass in `state.login.isAuthenticated` rather than using the routers location.

src/data/services/LmsApiService.js
```diff
@@ -15,24 +12,40 @@ class LmsApiService {
...
     return axios.get(portalConfigurationUrl, {
       withCredentials: true,
       headers: {
-        Authorization: `JWT ${this.jwtToken}`,
+        Authorization: `JWT ${jwtToken}`,
       },
     });
   }
+
+  static authenticate = (email, passWord) => {
```
georgebabey: `password` not `passWord`

src/components/LoadingSpinner/index.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { FadeLoader } from 'react-spinners';
```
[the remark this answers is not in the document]
adamstankiewicz: @georgebabey This component is an actual spinner that can be positioned inline next to the login button. My component is not a spinner, but rather what was in Roya's mocks:

image

The component I added would not work for the login form.

src/containers/Header/index.jsx
```diff
@@ -21,15 +21,17 @@ const Header = (props) => {
...
+          <Dropdown
+            title="edx@example.com"
```
adamstankiewicz: We'll probably need to store the user's email address in the Redux store so that the `Dropdown` knows what email address to use here.

src/data/services/LmsApiService.js
```diff
@@ -16,24 +14,41 @@ class LmsApiService {
...
       },
     });
   }
+
+  static authenticate(email, password) {
+    const loginData = {
+      grant_type: 'password',
+      username: email,
+      password,
+      client_id: 'd7cad6d795f7235e8a3c', // client_id configured from django( localhost admin)
```
georgebabey: Now that the config PR has landed can you update this to pull from the ./config and add the client_id to the webpack.dev.config and webpack.prod.config files?

src/data/services/LmsApiService.js
```diff
@@ -16,24 +14,41 @@ class LmsApiService {
...
+      token_type: 'jwt',
+    };
+    const authUrl = 'http://localhost:18000/oauth2/access_token';
```
georgebabey: This should use `${this.baseUrl}`

src/data/actions/loginForm.js
```diff
@@ -0,0 +1,37 @@
...
+  payload: { error },
+});
+
+const login = (email, password) => (
+  (dispatch) => {
+    dispatch(fetchLoginRequest());
+    return LmsApiService.authenticate(email, password)
+      .then((response) => {
+        const cookies = new Cookies();
+        cookies.set('access_token', response.data.access_token);
```
georgebabey: We're https everywhere so lets set the secure option on this cookie.

src/config/index.js
```diff
@@ -1,4 +1,6 @@
 export default {
   LMS_BASE_URL: process.env.LMS_BASE_URL,
   DATA_API_BASE_URL: process.env.DATA_API_BASE_URL,
+  LMS_CLIENT_ID: process.env.LMS_CLIENT_ID,
+  SECURE_COOKIES: process.env.NODE_ENV !== 'development',
```
georgebabey: 👍

PR #11, opened 2018-06-19
Separate API calls to their own services layer

adamstankiewicz:
[ENT-1050](https://openedx.atlassian.net/browse/ENT-1050)

Moves the API calls to the Enterprise data api and the LMS api to their own services layer to create better separation of concerns.

I also updated tests to increase coverage 📈. Still not fully covered but an improvement.

 Also, now that we have Travis CI running for the repo, I noticed our builds were failing. This was due to using Webpack 4 and a compatibility issue with `extract-text-webpack-plugin`. In fact, it's no longer recommended to use `extract-text-webpack-plugin` with Webpack 4 anymore; it's now recommended to use `mini-css-extract-plugin`. I needed to  upgrade `Webpack` to the latest version for `mini-css-extract-plugin` to not error during build. Then, the `is-es5` check was failing due to `const` being in a the compiled vendor.min.js file. I had to add `node_modules` to the `include` paths in webpack.prod.config.js for `babel` to transpile vendor packages from ES6 to ES5. 

After all that, we now have a passing build on Travis! 😃🎆🍾🎉 I'm going to file an issue in the `front-end-cookie-cutter-application` repo to fix this upstream.


src/components/CourseEnrollments/index.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
 import qs from 'query-string';
 import moment from 'moment';
+import _ from 'lodash';
```
georgebabey: Should import just what we need from method packages, since we're only using 2 methods

PR #14, opened 2018-07-10
Fix Docker start up.

douglashall:
This brings this repo inline with https://github.com/edx/front-end-cookie-cutter-application/blob/master/Dockerfile#L29 and enables `docker up` to work.


douglashall:
@georgebabey @adamstankiewicz Can you review/merge?

PR #15, opened 2018-07-16
Admin dashboard sorting & styles

adamstankiewicz:
#### Description
Adds sorting functionality to the course enrollments table, styles to match the mocks, updates to tests, and adds edX favicon. This PR covers these two tickets:
* [ENT-1037](https://openedx.atlassian.net/browse/ENT-1037)
* [ENT-1071](https://openedx.atlassian.net/browse/ENT-1071)

#### UI
image


src/components/Admin/index.jsx
```diff
@@ -51,14 +42,31 @@ class Admin extends React.Component {
   }
 
   renderErrorMessage() {
-    const message = `There was a problem fetching dashboard data: ${this.props.error.message}`;
-    return <StatusMessage alertType="danger" message={message} />;
+    return (
+      <StatusAlert
```
georgebabey: Lets wrap all of this into a ErrorAlert component that just takes the message as props so we can reuse it here and in CourseEnrollments

src/components/Admin/index.jsx
```diff
@@ -51,14 +42,31 @@ class Admin extends React.Component {
...
+        open
+      />
+    );
   }
 
   renderLoadingMessage() {
     return (
-      <div className="loading">
-        <p>Loading...</p>
+      <div className="overview loading mt-3 d-flex align-items-center justify-content-center">
```
georgebabey: May should wrap this as a LoadingMessage component as well. I assume we'll have loading messages on other pages so having a single place would be good - then if we ever change how we're displaying loading messages its in a single place.

If you want to leave this for now and refactor once we create the next LoadingMessage

src/components/CourseEnrollments/index.jsx
```diff
@@ -1,18 +1,20 @@
...
+import qs from 'query-string';
+import { withRouter } from 'react-router';
+import { Pagination, Icon, StatusAlert, Table } from '@edx/paragon';
+import { isNaN, omitBy } from 'lodash';
 
+import 'font-awesome/css/font-awesome.css';
 import './CourseEnrollments.scss';
 
-const StatusMessage = props => (
+const StatusAlertWrapper = props => (
```
georgebabey: Recommend pulling this out into a new file - this file is already getting quite long and this wrapper + the proptype definition for it could be pulled used elsewhere

src/components/Admin/index.jsx
adamstankiewicz: @georgebabey The `StatusAlert` here is no longer the Paragon `StatusAlert`, but rather the new wrapped component in `components/StatusAlert` that is reused here and in `CourseEnrollments`.

PR #16, opened 2018-07-20
ENT-1122. Error and 404 page

adamstankiewicz:


src/containers/ErrorPage/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+const ErrorPage = (props) => {
+  const { location } = props;
+  if (location.state && location.state.error) {
+    const { error } = location.state;
+
+    return (
+      <div className="container">
+        {error.status === 404 ? (
+          <Redirect to="/404" />
```
georgebabey: I don't love the double redirect here, first to /error and then to /404 but it is probably okay for now...

src/index.jsx
```diff
@@ -31,9 +37,11 @@ const AppWrapper = () => (
           <Header />
           <Switch>
             <Route exact path="/support" component={SupportPage} />
-            <Route path="/login" component={LoginPage} />
```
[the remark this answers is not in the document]
brittneyexline: fine with me

src/containers/ErrorPage/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+              dismissible={false}
+              dialog={location.state.error.message}
+              open
+            />
+          </div>
+        )}
+      </div>
+    );
+  }
+  return <Redirect to="/" />;
```
brittneyexline: won't this end up on the 404 page anyway?

PR #18, opened 2018-07-20
Support logout

adamstankiewicz:
Updates the paragon version to the latest to include the Dropdown changes that can accept elements. Passes `<Link>` component into the Dropdown for the logout link.

Note: This won't work until the Paragon release 3.3.3 is actually released (not sure why it didn't go through). Also, snapshot tests may need to be updated.


georgebabey:
@brittneyexline This is ready for review now

src/components/Header/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+        <Link
+          to={enterpriseSlug ? `/${enterpriseSlug}` : '/'}
+          className="navbar-brand"
+        >
+          <Img src={enterpriseLogo || EdxLogo} alt="" />
+        </Link>
+        {email && <Dropdown
+          title={email}
+          menuItems={[
+            <Link to="/login" onClick={logout}>Logout</Link>,
```
georgebabey: I kept this as a link to `/login` but with the onClick removing the cookie.  Other alternatives we could consider:

- Redirect in the action handler for logout
- Have a /logout route that removes the cookie and redirects

I was a worried with the currently implementation we'd get to the login page and still have isAuthenticated to true and bypass the page, before the action handler resolved. It seems to work in testing it, but it still feels a bit messy.

src/data/actions/authentication.test.js
```diff
@@ -2,17 +2,29 @@ import configureMockStore from 'redux-mock-store';
...
 const axiosMock = new MockAdapter(axios);
 
+jest.genMockFromModule('universal-cookie');
+jest.mock('universal-cookie');
+
+const mockCookies = {
+  set: jest.fn(),
+  remove: jest.fn(),
+};
+Cookies.mockImplementation(() => mockCookies);
```
georgebabey: This sets the Cookies constructor with our mock, so we can spy on set/remove below

src/components/Header/index.jsx
brittneyexline: yea, I would have expected a /logout route - that would make it easier to reuse the functionality if we need to trigger logout in other situations.

PR #19, opened 2018-07-23
ENT-1051 Use webpack EnvironmentPlugin for configuration variables

georgebabey:
This PR leverages the webpack EnvironmentPlugin to set configuration variables. Currently
only two values are needed:
- LMS_BASE_URL
- DATA_API_BASE_URL

By default values for development are set but the production values are left undefined.
Webpack will throw a warning if the variables are not defined. They can be passed in
through process.env:
```
LMS_BASE_URL=https://courses.edx.org DATA_API_BASE_URL=https://prod-edx-analyticsapi.edx.org npm run build
```

It is worth noting that the create-react-app project does support more complex scenarios with [custom environment variables])https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-custom-environment-variables) however this project does not include any of the necessary react-scripts to support this. We may want to consider adding in react-scripts or at least [dotenv-webpack](https://github.com/mrsteele/dotenv-webpack) support in the future.

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -2,6 +2,7 @@
 // time at the expense of creating larger, unoptimized bundles.
 const Merge = require('webpack-merge');
 const path = require('path');
+const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 const commonConfig = require('./webpack.common.config.js');
@@ -94,6 +95,11 @@ module.exports = Merge.smart(commonConfig, {
       inject: true, // Appends script tags linking to the webpack bundles at the end of the body
       template: path.resolve(__dirname, '../public/index.html'),
       favicon: path.resolve(__dirname, '../src/images/favicon.ico'),
+    }),
+    new webpack.EnvironmentPlugin({
+      NODE_ENV: 'development',
+      LMS_BASE_URL: 'http://localhost:18000',
+      DATA_API_BASE_URL: 'http://localhost:8000',
     }),
   ],
   // This configures webpack-dev-server which serves bundles from memory and provides live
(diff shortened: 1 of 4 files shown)


georgebabey:
@douglashall This will let us pass configuration in through the pipeline job for the production (stage & prod) builds.

douglashall:
@georgebabey Can the variables be read from a separate JSON file so it is easy for the pipeline to override? Also, can you add this to the frontend app cookie cutter?

georgebabey:
@douglashall As we discussed the pipeline can pass along the config when running `npm run build` This puts the responsibility for how to handle the configuration to the pipeline, not the repo, which feels right to me. 

We will need to add the portal configuration to edx-internal, I'm imagining something like:
```
EDX_INTERNAL_CONFIG:
  LMS_BASE_URL: 'https://courses.edx.org'
  DATA_API_BASE_URL: 'https://prod-edx-analyticsapi.edx.org
```
The pipeline will read that yaml and pass it along any configuration to`npm run build`

Lets try this out - if we like this approach I can add it back to the cookiecutter. Working on fixing the tests for this PR

douglashall:
@georgebabey I'm adding this to my edx-internal PR: https://github.com/edx/edx-internal/pull/500.

package.json
```diff
@@ -93,9 +93,6 @@
     ],
     "transformIgnorePatterns": [
       "/node_modules/(?!(@edx/paragon)/).*/"
-    ],
```
georgebabey: These globals in the package.json did not work, and courseContext was not used. I removed it to avoid confusion.

Plus setting up globals in src/setupTest.js seems better than polluting the package.json

PR #21, opened 2018-07-27
ENT-1128 Add enterprise list page

brittneyexline:
This introduces a new page which lists all the enterprises a user has access to, with each listed being a link to the enterprise's data dashboard page. If the user only has access to one enterprise, it will redirect to the enterprise's data dashboard.


src/components/EnterpriseList/index.jsx
```diff
@@ -0,0 +1,166 @@
...
+      />
+    );
+  }
+
+  renderLoadingMessage() {
+    return <LoadingMessage className="enterprise-list" />;
+  }
+
+  render() {
+    const { className, loading, error } = this.props;
```
georgebabey: I dont think className is needed as a props here. Doesn't look like it gets set and doesn't seem necessary

src/data/reducers/enterpriseList.test.js
```diff
@@ -0,0 +1,65 @@
...
+  loading: false,
+  error: null,
+};
+
+describe('enterpriseList reducer', () => {
+  it('has initial state', () => {
+    expect(enterpriseList(undefined, {})).toEqual(initialState);
+  });
+
+  it('updates fetch enterpriseList request state', () => {
```
georgebabey: I found this wording confusing and changed it in the  authentication reducer here: https://github.com/edx/edx-portal/pull/22/files#diff-e7c432af33f88589df966bdc2261a45aR22

Up to you if you want to change it here

PR #22, opened 2018-07-30
Reads the username from the jwt in the access_token cookie on app load

georgebabey:
Sets the store with the username if it exists in the cookie. This ensures the dropdown menu loads so you can properly logout


src/data/actions/authentication.js
```diff
@@ -48,14 +51,45 @@ const logoutEvent = () => ({
...
+// When loading user details from the JWT stored on the local browser
+const localUserLoaded = email => ({
+  type: LOCAL_USER_LOADED,
+  payload: { email },
+});
+
+// This is used when the local browser does not have a JWT with the user details. This is
+// not an error condition, the user may not have logged in yet. We still capture this event
+// for tracking purposes.
+const localUserMissing = () => ({
```
georgebabey: I'm not sure this is necessary, since it doesn't change the store, but it seemed like a useful event to capture

src/data/reducers/authentication.js
```diff
@@ -43,6 +44,12 @@ const authentication = (state = {
         error: null,
         email: null,
       };
+    case LOCAL_USER_LOADED:
+      return {
```
georgebabey: This is basically duplicated from FETCH_LOGIN_SUCCESS, other than the loading flag.

We could combine the two cases but I didn't know if the fall-through would be more confusing to read.

src/data/actions/authentication.js
brittneyexline: i'm ok with leaving it in

PR #24, opened 2018-08-01
Added support for csv downloads

georgebabey:
This PR adds support for downloading a csv report of the learner enrollment data.  Downloads are handled through redux actions so we can easily capture those events (and someday soon send them to segment)

This requires a change to edx-enterprise-data/edx-analytics-data-api (https://github.com/edx/edx-enterprise-data/pull/48) to support pulling down the entire dataset without pagination.

screen shot 2018-08-01 at 2 51 04 pm


package.json
```diff
@@ -54,6 +55,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
+    "babel-preset-minify": "^0.2.0",
```
georgebabey: For some reason this dev-dependency is needed when webpack bundles files-saver. It's listed as a preset in file-savers' babelrc file.  Seems odd to have to include it here but I couldn't find a way around `Error: Couldn't find preset "minify" relative to directory "/Users/georgebabey/dev/edx-portal"`

src/data/services/EnterpriseDataApiService.js
```diff
@@ -5,35 +5,39 @@ import { getAccessToken } from '../../utils';
 
 class EnterpriseDataApiService {
   // TODO: This should access the data-api through the gateway instead of direct
-  static enterpiseBaseUrl = `${config.DATA_API_BASE_URL}/enterprise/api/v0/enterprise/`;
+  static enterpriseBaseUrl = `${config.DATA_API_BASE_URL}/enterprise/api/v0/enterprise/`;
```
georgebabey: typo fix

src/data/services/EnterpriseDataApiService.js
```diff
@@ -5,35 +5,39 @@ import { getAccessToken } from '../../utils';
...
 
   static fetchCourseEnrollments(enterpriseId, options) {
     const queryParams = {
       page: 1,
       page_size: 50,
       ...options,
     };
-    let enrollmentsUrl = `${this.enterpiseBaseUrl}${enterpriseId}/enrollments/`;
+    const enrollmentsUrl = `${this.enterpriseBaseUrl}${enterpriseId}/enrollments?${qs.stringify(queryParams)}`;
```
georgebabey: Changed this to always pass in the paging options, even on initial load. Otherwise we rely on the backend to send us the correct `page_size` but the backend is currently configured to only send 10. This caused an error when we tried to page to the second page if we had > 10 but < 50.

src/data/services/EnterpriseDataApiService.js
```diff
@@ -5,35 +5,39 @@ import { getAccessToken } from '../../utils';
...
     const jwtToken = getAccessToken();
 
-    if (options && typeof options === 'object') {
-      enrollmentsUrl += `?${qs.stringify(queryParams)}`;
-    }
-
     return axios.get(enrollmentsUrl, {
-      withCredentials: true,
```
georgebabey: Because we're using token authentication in the header we don't need to pass along any cookies, so withCredentials wasn't needed and would have required extra CORS configurations on the data-api if we kept it.

src/components/Admin/index.jsx
```diff
@@ -117,10 +118,24 @@ class Admin extends React.Component {
...
+              <div className="row">
+                <div className="col-12 col-md-6 py-3 invisible">Showing data as of Month DD, YYYY</div>
+                <div className="col-12 col-md-6 text-md-right">
+                  <Button
+                    label={
+                      <span>
+                        <span className="fa fa-download" /> Download full report (CSV)
+                      </span>
+                    }
+                    onClick={() => downloadCsv(this.props.enterpriseId)}
```
brittneyexline: why not pull enterpriseId out of props above too?

PR #25, opened 2018-08-01
return total learners/users for dashboard

jonalexvasquez:
Need to check if my changes worked once portal works for me locally to test it out


jonalexvasquez:
The reason it displays number of users is because I was running the enterprise data server locally which picks up my changes but wouldn't work on yours since it is not merged yet!

brittneyexline:
Right - I am wondering if it will look acceptable without your data api changes being released yet, or if we need to wait to merge this until your other work is merged.

PR #29, opened 2018-08-06
ENT-1112 Use Segment to track events on admin dashboard

brittneyexline:


src/data/store.js
```diff
@@ -2,14 +2,35 @@ import { applyMiddleware, createStore } from 'redux';
...
+import Segment, { identifyUser, trackEvent, trackPageView } from '@redux-beacon/segment';
+import { LOCATION_CHANGE } from 'react-router-redux';
+import { FETCH_LOGIN_SUCCESS } from './constants/authentication';
+import { FETCH_CSV_REQUEST } from './constants/courseEnrollments';
 
 import reducers from './reducers';
 
 const loggerMiddleware = createLogger();
 
+const eventsMap = {
```
adamstankiewicz: I wonder if we created actions that could be dispatched for sorting/pagination so that we can track those here as well?

config/webpack.dev.config.js
```diff
@@ -101,6 +102,7 @@ module.exports = Merge.smart(commonConfig, {
       LMS_BASE_URL: 'http://localhost:18000',
       DATA_API_BASE_URL: 'http://localhost:8000',
       LMS_CLIENT_ID: '6cb72dc5675a1e7efe81',
+      SEGMENT_KEY: 'GANzwyDLm1mSKeqZab6apWRHZW9TxQNe',
```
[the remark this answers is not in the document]
brittneyexline: I guess we would only need it if we are specifically debugging something related to Segment, so I am ok with removing this.

src/data/store.js
brittneyexline: I think that is what we would have to do in order to track them - I decided not to take on that lift in this PR.

adamstankiewicz: Gotcha. Should we create another ticket for the tracking of sorting/pagination then? Or did we decide it's not important to track.

brittneyexline: I'll create a ticket.

PR #35, opened 2018-08-09
Add a default route that redirects to /enterprises

georgebabey:
This allows the user to visit just portal.edx.org after logging in and
not end up on a 404 page. Also renamed `App` to `EnterpriseApp` to make it
more explicit that it is not the entrypoint for the entire App, as that
is served by the index.js and the `<AppWrapper>`


src/index.jsx
```diff
@@ -44,7 +44,8 @@ const AppWrapper = () => (
             <Route exact path="/404" component={NotFoundPage} />
             <Route exact path="/error" component={ErrorPage} />
             <PrivateRoute path="/enterprises" component={EnterpriseIndexPage} />
-            <PrivateRoute path="/:enterpriseSlug" component={App} />
+            <PrivateRoute path="/:enterpriseSlug" component={EnterpriseApp} />
+            <Route exact path="/" component={EnterpriseIndexPage} />
```
georgebabey: Other than the rename, this is the only change. We use `exact` here so that other, non-matching URLs still 404 below.

PR #38, opened 2018-08-09
Dispatch logout action when API calls return 401 or 403 errors

adamstankiewicz:
When an API response returns an error with a status code of 401 (unauthorized) or 403 (forbidden), we should force logout the user so they return to the login page to reset their session.

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -100,7 +100,7 @@ module.exports = Merge.smart(commonConfig, {
       NODE_ENV: 'development',
       LMS_BASE_URL: 'http://localhost:18000',
       DATA_API_BASE_URL: 'http://localhost:8000',
-      LMS_CLIENT_ID: '6cb72dc5675a1e7efe81',
+      LMS_CLIENT_ID: 'CMehRRNqfiBRVJKnPOkjBDjAvurtnHpELoehKAvZ',
     }),
   ],
   // This configures webpack-dev-server which serves bundles from memory and provides live
--- a/src/data/services/EnterpriseDataApiService.js
+++ b/src/data/services/EnterpriseDataApiService.js
@@ -1,7 +1,7 @@
-import axios from 'axios';
 import qs from 'query-string';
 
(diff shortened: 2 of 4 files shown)


src/utils.js
```diff
@@ -24,7 +28,17 @@ const getAccessToken = () => {
...
+axios.interceptors.response.use(response => response, (error) => {
+  const errorStatus = error.response.status;
+  if (errorStatus === 401 || errorStatus === 403) {
+    store.dispatch(logout());
+  }
+  return Promise.reject(error);
+});
+
 export {
+  axios,
```
georgebabey: I think we should call this something different, maybe `httpClient` so we know its not stock axios

config/webpack.dev.config.js
```diff
@@ -100,7 +100,7 @@ module.exports = Merge.smart(commonConfig, {
       NODE_ENV: 'development',
       LMS_BASE_URL: 'http://localhost:18000',
       DATA_API_BASE_URL: 'http://localhost:8000',
-      LMS_CLIENT_ID: '6cb72dc5675a1e7efe81',
+      LMS_CLIENT_ID: 'CMehRRNqfiBRVJKnPOkjBDjAvurtnHpELoehKAvZ',
```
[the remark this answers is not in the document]
adamstankiewicz: Haha yeah, I shook my head when I accidentally committed that after just having that conversation :P

src/utils.js
adamstankiewicz: Good idea!

src/utils.js
```diff
@@ -30,15 +30,15 @@ const getAccessToken = () => {
...
+  const errorStatus = error && error.response && error.response.status;
   if (errorStatus === 401 || errorStatus === 403) {
     store.dispatch(logout());
   }
   return Promise.reject(error);
 });
 
 export {
-  axios,
+  axios as httpClient,
```
georgebabey: I feel like this should be moved into its own file.

PR #42, opened 2018-08-13
Adds error and loading state for CSV downloading

georgebabey:
This also upgrades edx-bootstrap to 1.0.0 which includes bootstrap 4.0.0 rather than the 4 beta.
The bootstrap 4 beta had removed cursor:pointer on buttons, we no longer need to handle this ourselves.

### DIFF
--- a/src/components/Admin/index.jsx
+++ b/src/components/Admin/index.jsx
@@ -58,6 +58,18 @@ class Admin extends React.Component {
     );
   }
 
+  renderCsvErrorMessage() {
+    return (
+      <StatusAlert
+        className={['mt-3']}
+        alertType="danger"
+        iconClassName={['fa', 'fa-times-circle']}
+        title="Unable to Generate CSV Report"
+        message={`Please try again. (${this.props.csvError.message})`}
+      />
+    );
+  }
+
   renderLoadingMessage() {
     return <LoadingMessage className="overview mt-3" />;
   }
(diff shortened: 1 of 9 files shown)


src/components/Admin/index.jsx
```diff
@@ -58,6 +58,18 @@ class Admin extends React.Component {
     );
   }
 
+  renderCsvErrorMessage() {
```
georgebabey: CSV Button could probably be its own component, but for now I kept it in the admin page.

src/components/Admin/index.jsx
```diff
@@ -135,9 +151,10 @@ class Admin extends React.Component {
                   </div>
                   <div className="col-12 col-md-6 text-md-right">
                     <Button
+                      disabled={csvLoading}
                       label={
                         <span>
-                          <Icon className={['fa', 'fa-download', 'mr-1']} /> Download full report (CSV)
+                          <Icon className={['fa', 'mr-1'].concat(downloadButtonIconClasses)} /> Download full report (CSV)
```
adamstankiewicz: Nice!

src/index.scss
```diff
@@ -3,7 +3,3 @@
 
 $fa-font-path: "~font-awesome/fonts";
 @import "~font-awesome/scss/font-awesome";
-
-button:hover {
```
adamstankiewicz: Good catch on the version of `edx-bootstrap` so that we can remove this.

src/components/Admin/index.jsx
adamstankiewicz: Agreed. Should we create a ticket for this?

src/index.scss
georgebabey: Followup PR to frontend-cookiecutter coming

PR #43, opened 2018-08-13
ENT-1159 Handle 404s from data api more gracefully

brittneyexline:


src/components/Admin/index.jsx
```diff
@@ -46,6 +46,21 @@ class Admin extends React.Component {
...
+      courseCompletions,
+      enrolledLearners,
+    } = this.props;
+    let hasEmptyData = true;
+
+    [courseCompletions, enrolledLearners, numberOfUsers].forEach((item) => {
+      hasEmptyData = item === 0 && hasEmptyData;
+    });
+
+    return hasEmptyData;
```
georgebabey: might read better with `.every` (mdn)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every]

```
return [courseCompletions, enrolledLearners, numberOfUsers].every((item) => item === 0);
```

src/data/actions/courseEnrollments.js
```diff
@@ -28,6 +35,10 @@ const fetchCourseEnrollments = (enterpriseId, options) => (
         dispatch(fetchCourseEnrollmentsSuccess(response.data));
       })
       .catch((error) => {
+        if (error.response.status === 404) {
+          dispatch(fetchCourseEnrollmentsSuccess(emptyEnrollments));
```
georgebabey: I could see someone coming back to this code and being confused why we're doing this - might want to add a comment explaining that the API returns a 404 if there are no enrollments found and this prevents the page from treating that as an error?

src/components/Admin/index.jsx
adamstankiewicz: If we change to `.every` here, we may want to do it [here](https://github.com/edx/edx-portal/blob/fef1012df6974412c3f1063841a5215110aded51/src/components/Admin/index.jsx#L42) as well. #consistency

brittneyexline: i'm not sure we can use .every there because it is or'ing instead of and'ing.

georgebabey: If thats the case we could use `.some` instead of `.every` but it looks to me like we're checking that every one of those values is zero?

brittneyexline: ah i meant for the other place adam linked - for this case it makes sense

adamstankiewicz: Yeah, you're right. Sorry about that :) That being said, perhaps `.some` would be of use there. Feel free to change it if you'd like.

PR #45, opened 2018-08-16
Handle when no branding config exists, enterprise logo image refactor

adamstankiewicz:
Refactors how the enterprise branding configuration is set up for the admin dashboard. Before, we were making a call to the `enterprise_branding_configuration` endpoint to get the enterprise's logo. However, if an enterprise is not configured with a logo, a branding configuration object does not exist for that `EnterpriseCustomer`, resulting in a 404 response and causing the dashboard to render the `NotFoundPage`.

The branding configuration data is part of the response of the `with_access_to` endpoint on `EnterpriseList`. So, we're now using the branding data returned by that endpoint instead to set up the `portalConfiguration` Redux state. This changes means we are actually removing an API call from the dashboard.  💯 🎉

Also included in this PR are minor refactors and some additional logic to handle when an enterprise logo does exist but doesn't load for whatever reason (e.g., logo file no longer exists, network error). When this happens: in the `Header`, we would show the edX logo instead of no logo; and in the `Footer`, we would hide the entire placeholder where the enterprise logo usually shows (doesn't make sense to show two edX logos in the footer 😉).

Note: I've added some review comments for clarifications about specific code instead of outlining them here.


src/components/Footer/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  renderEnterpriseLogo() {
+    const { enterpriseLogo, enterpriseSlug, enterpriseName } = this.props;
+    return (
+      <Link className="logo" to={`/${enterpriseSlug}`}>
+        <Img
+          src={enterpriseLogo}
+          alt={`${enterpriseName} logo`}
+          onError={() => this.setState({ enterpriseLogoNotFound: true })}
```
adamstankiewicz: The `onerror` attribute on an `img` handles the case when the browser tries to load an image, but the image fails to load (e.g., the image no longer exists at that URL, network issues, etc.).

Here, when the image can't load, we don't want to show any image at all. I handle this by setting state that the image can't be loaded which then prevents the `renderEnterpriseLogo` function was being called (see the logic in the `render` function on line 50).

src/components/Header/index.jsx
```diff
@@ -8,42 +8,53 @@ import LinkWrapper from '../LinkWrapper';
...
-    email,
-  } = props;
+class Header extends React.Component {
+  renderLogo() {
+    const { enterpriseLogo, enterpriseName } = this.props;
+    return (
+      <Img
+        src={enterpriseLogo || EdxLogo}
+        alt={`${enterpriseName || 'edX'} logo`}
+        onError={(e) => { e.target.src = EdxLogo; }}
```
adamstankiewicz: Similar to the logic in the `Footer` above, but in the callback function for `onerror` we _do_ want to show a logo (otherwise, the `Header` would be logoless!). In this case, if the enterprise logo can't be loaded, we swap out the `img` src with the edX logo.

src/containers/EnterpriseApp/index.jsx
```diff
@@ -9,17 +9,38 @@ import AdminPage from '../AdminPage';
...
-import { fetchPortalConfiguration } from '../../data/actions/portalConfiguration';
+import fetchEnterpriseList from '../../data/actions/enterpriseList';
+import { setPortalConfiguration } from '../../data/actions/portalConfiguration';
 import { getLocalUser } from '../../data/actions/authentication';
 
 class EnterpriseApp extends React.Component {
   componentDidMount() {
     const { enterpriseSlug } = this.props.match.params;
+    if (!this.props.enterprises) {
+      this.props.fetchEnterpriseList();
```
adamstankiewicz: With this refactor, we are getting the enterprise branding configuration from the API response returned in the `with_access_to` endpoint. On a hard refresh of the dashboard page, we never call that endpoint so we need to do that here if no enterprises are loaded yet.

src/containers/EnterpriseApp/index.jsx
```diff
@@ -9,17 +9,38 @@ import AdminPage from '../AdminPage';
...
+    } else {
+      const enterprise = this.findEnterprise(enterpriseSlug);
+      this.props.setPortalConfiguration(enterprise);
+    }
 
-    this.props.getPortalConfiguration(enterpriseSlug);
     this.props.getLocalUser();
   }
 
+  componentDidUpdate(prevProps) {
```
adamstankiewicz: The logic here handles the case when the enterprises already exist (i.e., user coming from the `EnterpriseList` page), so we just need to dispatch the `setPortalConfiguration` action.

src/containers/EnterpriseApp/index.jsx
```diff
@@ -42,26 +63,25 @@ class EnterpriseApp extends React.Component {
 
     return (
       <div>
-        {enterpriseId &&
```
adamstankiewicz: We no longer need to verify that the `enterpriseId` exists before rendering the redirect/routes as it always will exist.

src/data/reducers/portalConfiguration.js
```diff
@@ -1,49 +1,32 @@
...
-        ...state,
-        loading: false,
-        error: action.payload.error,
-        enterpriseId: null,
-        enterpriseSlug: null,
-        enterpriseLogo: null,
+        enterpriseId: action.payload.data.uuid,
+        enterpriseName: action.payload.data.name,
+        enterpriseSlug: action.payload.data.slug,
+        enterpriseLogo: action.payload.data.branding_configuration
```
adamstankiewicz: This is split onto multiple lines for linting purposes (line exceeds max length otherwise).

PR #46, opened 2018-08-16
ENT-1179. Fix react router / redux to dispatch LOCATION_CHANGE events

georgebabey:
Two issues:
- Missing routerMiddleware
- Had two routers defined, one from react-router-redux as well as the default BrowserRouter from react-router-dom


adamstankiewicz:
Good catch on noticing we had two routers defined.

PR #48, opened 2018-08-17
Searchbar implementation

jonalexvasquez:
Created the UI for the searchbar. When user searches for enterprises and only one enterprise is returned it will still list that enterprise instead of directly redirecting them


jonalexvasquez:
Need to implement a way to clear and reset a search to return to the full list of enterprises

adamstankiewicz:
This PR can be closed due to #60 merging.

PR #51, opened 2018-08-21
ENT-1154. Support loading demo data bookmarklet with new script

georgebabey:
This PR adds a new standalone script to the prod build called demoDataLoader.js.  The intention is for this script to be loaded manually from a bookmarklet, at which point the calls to the data-service will return static data instead of hitting the endpoint.

This will allow sales to demo with a more representative demo company without having to pollute the production database with this data.

A nice side-effect of this PR is that we can also leverage this in development by uncommenting the entrypoint in the webpack.dev.config allowing the option of quick frontend development without having to run the data-api.

The bookmarklet is available here: https://codepen.io/edx/live/e5f46af8f39968b9693c8414091f6cc3/


config/webpack.prod.config.js
```diff
@@ -113,6 +116,10 @@ module.exports = Merge.smart(commonConfig, {
   },
   // Specify additional processing or side-effects done on the Webpack output bundles as a whole.
   plugins: [
+    // Cleans the dist directory before each build
+    new CleanWebpackPlugin(['dist'], {
```
adamstankiewicz: Woo, no more old build files in `dist`! 😃

src/demo/data.js
```diff
@@ -0,0 +1,51 @@
...
+  const lastName = faker.name.lastName();
+
+  const course = courses[faker.random.number({ max: courses.length - 1 })];
+  const courseStart = faker.date.past();
+  const courseEnd = faker.date.future(null, courseStart);
+  const hasPassed = Math.random() <= percentPassed;
+  const grade = hasPassed ? faker.random.number({ min: 70, max: 100 }) :
+    faker.random.number({ max: 70 });
+
+  // TODO: we could generate all columns so we could easily fake the csv call as well
```
adamstankiewicz: Do we know if the csv download is something sales/product would want to demo? If so, we should probably include it in this PR.

georgebabey: Good question - I talked to product and they feel opening the downloaded CSV would not be part of the demo so I left it off for now. Could always add that in later if we/they find it useful.

PR #55, opened 2018-08-22
Add support for sorting of demo data

georgebabey:
This adds support for sorting to the demoDataLoader.js script. Also a small change to only pull
in the english locale for faker, as that library is quite large

### DIFF
--- a/src/demo/rewireEnterpriseDataApiService.js
+++ b/src/demo/rewireEnterpriseDataApiService.js
@@ -6,7 +6,20 @@ const rewire = () => {
     const pageSize = options.page_size || 50;
     const page = options.page || 1;
     const start = (page - 1) * pageSize;
-    const results = enrollments.slice(start, start + pageSize);
+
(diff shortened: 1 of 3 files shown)


src/demo/rewireEnterpriseDataApiService.js
```diff
@@ -6,7 +6,20 @@ const rewire = () => {
     const pageSize = options.page_size || 50;
     const page = options.page || 1;
     const start = (page - 1) * pageSize;
-    const results = enrollments.slice(start, start + pageSize);
+
+    // A comparator function that uses the options.ordering field to determine sort order
+    const sortByOptions = (a, b) => {
+      const ordering = options.ordering || 'user_email';
+      const isDesc = ordering.startsWith('-');
```
adamstankiewicz: Confirmed that we are using the `babel-polyfill` so `startsWith` should be safe to use: https://github.com/edx/edx-portal/blob/682a4d72d181ea129c0c693f8bcc0d4e67dddb31/src/index.jsx#L1

PR #56, opened 2018-08-22
WIP JWT Cookie Authentication

douglashall:
**Sandboxes:**

- tygra.sandbox.edx.org - Serves the edx-portal frontend app via nginx
- panthro.sandbox.edx.org - LMS
- analytics-api-jaga.sandbox.edx.org - edx-analytics-data-api

### DIFF
--- a/src/data/actions/authentication.js
+++ b/src/data/actions/authentication.js
@@ -1,7 +1,6 @@
 import Cookies from 'universal-cookie';
 import jwtDecode from 'jwt-decode';
 
-import configuration from '../../config';
 import {
   FETCH_LOGIN_FAILURE,
   FETCH_LOGIN_REQUEST,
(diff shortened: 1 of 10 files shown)


src/data/actions/authentication.js
```diff
@@ -31,17 +30,8 @@ const fetchLoginFailure = error => ({
 const login = (email, password) => (
   (dispatch) => {
     dispatch(fetchLoginRequest());
-    return LmsApiService.authenticate(email, password)
+    return AuthService.login(email, password)
```
robrap: Can you explain what when wrong with the original plan for the single login page?  Is this what made us have to use cross domain csrf?  I think it might expand the concerned pages when it comes to click-jacking as well.  I know you will be discussing with Nimisha as well when she returns.

PR #60, opened 2018-08-27
Support searching for enterprises from EnterpriseList

adamstankiewicz:
This PR adds support for searching for enterprise(s) from the `EnterpriseList` component, using Paragon's `SearchField`. 

On desktop, the search box is positioned to the right; on mobile, it drops below the "Enterprise List" heading and expands to full width.

image


src/components/Admin/Admin.scss
```diff
@@ -10,7 +10,3 @@
         flex: 1;
     }
 }
-
-.download-btn div {
```
adamstankiewicz: With the most recent Paragon release, the `div` in the `Icon` component was changed to a `span` so this CSS is no longer needed.

src/components/EnterpriseList/index.jsx
```diff
@@ -126,23 +173,33 @@ class EnterpriseList extends React.Component {
...
+          </div>
+          <div className="row mt-2">
+            <div className="col">
               {error && this.renderErrorMessage()}
               {loading && !enterprises && this.renderLoadingMessage()}
               {!loading && !error && enterprises && enterprises.length === 0 &&
                 this.renderEmptyEnterpriseListMessage()
               }
-              {!loading && !error && enterprises && enterprises.length === 1 && pageCount === 1 &&
+              {this.shouldRenderRedirectToEnterpriseAdminPage() &&
```
adamstankiewicz: Moving this conditional to its own function to reduce some complexity in the `render` function.

src/index.scss
```diff
@@ -3,3 +3,6 @@
 
 $fa-font-path: "~font-awesome/fonts";
 @import "~font-awesome/scss/font-awesome";
+
+@import "~@edx/paragon/src/SearchField/SearchField";
```
adamstankiewicz: To make Paragon styles work properly, we need to include the `scss` file for any component that has non-native Bootstrap here.

src/components/EnterpriseList/EnterpriseList.test.jsx
```diff
@@ -159,11 +178,55 @@ describe('<EnterpriseList />', () => {
...
+
+      // onSubmit
+      wrapper.find('SearchBar').find('input[type=\'search\']').simulate('change', { target: { value: searchQuery } });
+      expect(wrapper.find('SearchBar').find('input[type=\'search\']').prop('value')).toEqual(searchQuery);
+      wrapper.find('SearchBar').find('.input-group-append').find('button').last()
+        .simulate('click');
+      expect(wrapper.find('EnterpriseList').instance().state.searchQuery).toEqual(searchQuery);
+      expect(wrapper.find('EnterpriseList').instance().state.searchSubmitted).toBeTruthy();
+
+      // // onClear
```
georgebabey: nit: double `//`s

src/components/EnterpriseList/EnterpriseList.test.jsx
```diff
@@ -159,11 +178,55 @@ describe('<EnterpriseList />', () => {
...
         enterprises={null}
       />
-    ));
-    expect(wrapper.find('EnterpriseList').instance().state.pageCount).toEqual(null);
+    )).find('EnterpriseList');
+    expect(wrapper.dive().state('pageCount')).toEqual(null);
+  });
+
+  describe('enterprise list search', () => {
+    it('searchQuery state changes onSearch and onClear', () => {
```
georgebabey: nit: I'd break this into 2 tests

PR #62, opened 2018-08-30
JWT Cookie Auth Login Redirect

douglashall:


src/components/Header/index.jsx
```diff
@@ -37,7 +37,7 @@ class Header extends React.Component {
           {email && <Dropdown
             title={email}
             menuItems={[
-              <LinkWrapper to="/logout">Logout</LinkWrapper>,
```
adamstankiewicz: Now that we no longer need to use the `LinkWrapper` component, we might want to remove the component entirely; this was the only place it was being used.

src/data/reducers/index.js
```diff
@@ -3,16 +3,21 @@ import { routerReducer } from 'react-router-redux';
 import courseOutline from './courseOutline';
 import dashboardAnalytics from './dashboardAnalytics';
 import portalConfiguration from './portalConfiguration';
-import authentication from './authentication';
 import table from './table';
 import csv from './csv';
 
+const identityReducer = (state) => {
```
adamstankiewicz: Is there a reason why you're not having the `identityReducer` in its own file, i.e. `identityReducer.js`?

src/components/Header/index.jsx
douglashall: Does `LinkWrapper` provide something that we don't get from an `<a>` tag?

src/data/reducers/index.js
douglashall: Seemed like overkill in this case.

src/index.jsx
```diff
@@ -46,4 +52,8 @@ const AppWrapper = () => (
   </Provider>
 );
 
-ReactDOM.render(<AppWrapper />, document.getElementById('root'));
+if (routeConfig.isCurrentRoutePublic() || apiClient.isAuthenticated()) {
```
georgebabey: Can we add a comment here indicating this is outside the rendering of react in order to short-circuit loading the app just to redirect.

src/index.jsx
```diff
@@ -8,20 +8,31 @@ import {
...
+const routeConfig = getRouteConfig();
+routeConfig.publicRoutes.forEach((publicRoute) => {
+  routes.push(<Route exact path={publicRoute.path} component={publicRoute.component} />);
+});
+routeConfig.privateRoutes.forEach((privateRoute) => {
+  routes.push(<PrivateRoute
+    path={privateRoute.path}
+    component={privateRoute.component}
+    authenticatedAPIClient={apiClient}
+    redirect={process.env.BASE_URL}
```
georgebabey: This should redirect back to the privateRoute.path instead of BASE_URL.

src/components/Header/index.jsx
adamstankiewicz: @douglashall The links within the Paragon `Dropdown` component must be react-router `Link` components for routing purposes. However, there was an issue with just using the standard `Link` since the `ref` gets set with an `innerRef` prop, not the usual `ref` prop. The `LinkWrapper` component was just that -- a wrapper around `Link` that a `ref` can be set on with an associated `focus` method.

douglashall: Okay cool, so since we are not linking to another page within the SPA we don't need `Link` or `LinkWrapper`.

douglashall:
@georgebabey I think this is good to go now. You can play with it at https://tygra.sandbox.edx.org. Can you give an approval?

src/data/store.js
```diff
@@ -45,8 +42,15 @@ const segmentMiddleware = createMiddleware(eventsMap, Segment());
 
 const middleware = [thunkMiddleware, loggerMiddleware, routerHistoryMiddleware, segmentMiddleware];
 
+const initialState = apiClient.getAuthenticationState();
+if (initialState.authentication) {
+  // eslint-disable-next-line no-undef
+  analytics.identify(initialState.authentication.email);
```
georgebabey: Rather than calling `analytics` directly can we change this to something like `identifyUser(email => { userId: email})(initialState.authentication.email);`

Since the rest of the tracking is going through beacon it seems to make sense to have this also use beacons identifyUser event, even if we just call it immediately.

src/utils.js
```diff
@@ -72,12 +66,14 @@ const getPageOptionsFromUrl = () => {
 
 const removeTrailingSlash = path => path.replace(/\/$/, '');
 
+const isRoutePublic = path => /^\/public.*$/.test(path);
```
georgebabey: 👍

PR #64, opened 2018-09-04
Generic table component w/ redux for paging/sort

georgebabey:
This PR introduces generic `TableContainer` and `TableComponent` components. The `TableContainer` handles all of the pagination and sorting through the redux store, loading indicators and passing pagination and sorting to segment is also included.

Table state is now stored in:
```
{
  table: {
    table-id: {
      loading: false,
      error: null,
      data: {
        count: ###,
        current_page: #,
        num_pages: ##
        results: [{...}, {...}, ....]
     }
  }
```

This does not currently address clearing the data, a followup PR will handle that.

The `EnrollmentsTable` has an example usage of the `TableContainer`:
```
    <TableContainer
      id="enrollments"
      className="enrollments"
      fetchMethod={EnterpriseDataApiService.fetchCourseEnrollments}
      columns={enrollmentTableColumns}
      formatData={formatEnrollmentData}
      tableSortable
    />
```

  The `EnterpriseList` was also refactored to use this new component


src/components/EnterpriseList/index.jsx
```diff
@@ -3,178 +3,96 @@ import PropTypes from 'prop-types';
...
 
   handleSearch(query) {
-    this.props.setSearchQuery(query);
     this.setState({
       searchQuery: query,
       searchSubmitted: true,
     });
+    this.props.searchEnterpriseList({
+      // Ternary used to turn empty string ('') into undefined so it does not get set in qs
+      search: query ? query : undefined, // eslint-disable-line no-unneeded-ternary
```
adamstankiewicz: Could this actually be `search: query || undefined,` to avoid the eslint disable?

src/components/EnterpriseList/index.jsx
```diff
@@ -3,178 +3,96 @@ import PropTypes from 'prop-types';
...
-        alertType="warning"
-        iconClassName={['fa', 'fa-exclamation-circle']}
-        message={emptyMessage}
-      />
-    );
+  fetchEnterprisesWithSearch = (options) => {
+    const optionsWithSearch = {
+      ...options,
+      // Ternary used to turn empty string ('') into undefined so it does not get set in qs
+      search: this.state.searchQuery ? this.state.searchQuery : undefined,
```
adamstankiewicz: Could this actually be `search: this.state.searchQuery || undefined`, to avoid the eslint disable?

src/containers/EnterpriseIndexPage/EnterpriseIndexPage.test.jsx
```diff
@@ -43,11 +45,12 @@ describe('<EnterpriseIndexPage />', () => {
...
     expect(wrapper.props().error).toEqual(null);
-    expect(wrapper.props().enterprises).toEqual(initialState.enterpriseList.enterprises);
+    expect(wrapper.props().enterprises).toEqual(initialState.table['enterprise-list'].data);
   });
 
-  it('getEnterpriseList dispatches fetchEnterpriseList action', () => {
-    wrapper.props().getEnterpriseList();
+  it('searchEnterpriseList dispatches searchEnterpriseList action', () => {
+    wrapper.props().searchEnterpriseList();
+    // TODO: tohavebeencalledwith?
```
adamstankiewicz: Can this comment be more explicit about the reasoning for using `toHaveBeenCalledWith` if it's still a todo?

src/data/store.js
```diff
@@ -24,6 +25,20 @@ const eventsMap = {
...
+    properties: {
+      table: action.payload.tableId,
+      page: action.payload.page,
+    },
+  })),
+  [SORT_REQUEST]: trackEvent(action => ({
+    name: 'Table Sort',
+    properties: {
+      table: action.payload.tableId,
+      column: action.payload.column,
```
adamstankiewicz: Do we also want the sort direction here?

src/components/EnterpriseList/index.jsx
```diff
@@ -3,178 +3,87 @@ import PropTypes from 'prop-types';
...
+    name: enterprise.name,
+    slug: enterprise.slug,
+    uuid: enterprise.uuid,
+  }));
 
-  renderEmptyEnterpriseListMessage() {
-    const { searchQuery } = this.props;
-    let emptyMessage = 'You do not have permission to view any enterprise customer data.';
-    if (searchQuery) {
-      emptyMessage = 'There are no enterprise customers that match your search.';
```
adamstankiewicz: Are we ok with not having custom empty state messages within `TableComponent`? With the refactor, it would always say "There are no results." In some scenarios, we may want to change that message.

src/components/TableComponent/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+  renderLoadingMessage() {
+    return <LoadingMessage className="table-loading" />;
+  }
+
+  renderErrorMessage() {
+    return (
+      <StatusAlert
+        alertType="danger"
+        iconClassName={['fa', 'fa-times-circle']}
+        title="Unable to load full report"
```
adamstankiewicz: Similar to the comment above -- do we want to be able to customize error messages/titles in `TableComponent`? Right now, every table using this component would have the title "Unable to load full report."

src/components/TableComponent/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+          this.renderEmptyDataMessage()
+        }
+        {data && data.length > 0 && this.renderTableContent()}
+      </div>
+    );
+  }
+}
+
+TableComponent.propTypes = {
+  // Props expected from consumer
```
adamstankiewicz: I like the way you're separating the props by consumer/container here 👍

src/components/TableLoadingOverlay/TableLoadingOverlay.scss
```diff
@@ -0,0 +1,19 @@
+.table-loading-overlay {
+  position: absolute;
+  z-index: 10;
+  background:  #000000;
```
adamstankiewicz: nit: Extra space before the color

src/components/TableLoadingOverlay/TableLoadingOverlay.scss
```diff
@@ -0,0 +1,19 @@
...
+.table-loading-message {
+  position: absolute;
+  z-index: 11;
+  width: 100%;
+  height: 100%;
+
+  span {
+    padding: 20px;
+  }
+}
```
adamstankiewicz: nit: Missing blank line at the end

src/components/EnterpriseList/EnterpriseList.test.jsx
```diff
@@ -116,76 +158,24 @@ describe('<EnterpriseList />', () => {
...
       wrapper = mount((
         <MemoryRouter initialEntries={['/test']}>
-          <EnterpriseList
-            enterprises={oneEnterpriseListData}
-            getEnterpriseList={() => {}}
-            clearPortalConfiguration={() => {}}
-            getLocalUser={() => {}}
-            setSearchQuery={() => {}}
-          />
+          <ContextProvider>
```
adamstankiewicz: Similar comment as above regarding just killing `ContextProvider` and going with `<Provider>`.

src/components/EnrollmentsTable/EnrollmentsTable.test.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import { Provider } from 'react-redux';
+
+import EnrollmentsTable from './index';
+// import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
+// import { mockEnrollmentFetchResponse } from './EnrollmentsTable.mocks';
```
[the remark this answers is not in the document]
georgebabey: Yeah these are for the commented out test that I haven't given up on, but don't want to continue to block anyone...

src/components/EnterpriseList/EnterpriseList.test.jsx
```diff
@@ -1,20 +1,65 @@
...
 import { shallow, mount } from 'enzyme';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
 
 import EnterpriseList from './index';
 import mockEnterpriseList from './EnterpriseList.mocks';
 
+const mockStore = configureMockStore([thunk]);
+
+class ContextProvider extends React.Component {
```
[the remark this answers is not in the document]
georgebabey: I was going to try to refactor that in a separate PR

src/components/EnterpriseList/index.jsx
georgebabey: We could always add in a props for overriding if we need to... I figured `There are no results.` was generic enough to work in both use-cases we have now (the enterprise list and the data tables) but if we want I could add the `emptyMessage` props now

src/components/TableComponent/index.jsx
georgebabey: Will make this more generic. As above, we can follow up with custom messages as props if we decide to

PR #69, opened 2018-09-18
Clears out table data when the component unmounts

georgebabey:
There is no reason to keep this data around as we fetch it again
whenever a TableComponent is loaded. This clears out that data from
the redux store on `componentWillUnmount`


src/data/reducers/table.js
```diff
@@ -52,6 +53,13 @@ const tableReducer = (state = {}, action) => {
         loading: false,
         error: action.payload.error,
       });
+    case CLEAR_TABLE:
+      return updateTable(state, action.payload.tableId, {
+        loading: false,
+        error: null,
+        ordering: undefined,
+        data: undefined,
```
adamstankiewicz: We've typically been using `null` as the default value when no data exists in our reducers. Both `ordering` and `data` could be set to `null` here. The [ordering value](https://github.com/edx/edx-portal/blob/master/src/data/actions/table.js#L87) in `actions/table.js` would ultimately get set to `undefined` regardless.

src/data/reducers/table.test.js
```diff
@@ -0,0 +1,90 @@
...
+      'table-1': {
+        data: [{}, {}],
+      },
+    };
+    const expected = {
+      'table-1': {
+        loading: false,
+        error: null,
+        ordering: undefined,
+        data: undefined,
```
adamstankiewicz: If we change the values to `null` per the comment above, we'd also need to change these to `null` here as well.

PR #70, opened 2018-09-18
WIP. This fixes the loading of the paging/sorting from the url

georgebabey:
This however introduces an issue that needs resolving before merging:
On the EnterpriseApp/index.jsx we are calling fetchEnterpriseList()
so that we can then look up the enterpriseId based on slug later on...

I propose we move this data out such that:
EnterpriseList fetches the full list as it does now
Admin component should fetch the portal configuration based on the slug

To support this we will need to update the with_access_to call in the backend
to accept a slug. Then we will need to move around the calls in the frontend to
support this flow


georgebabey:
@brittneyexline @christopappas @adamstankiewicz This is now ready for review

adamstankiewicz:
Also, I noticed that the URL query parameters (e.g., `page_size`, `features`) disappear from the URL on pages with a `TableComponent` (unrelated to this PR; it's like this on stage right now).

I think this is because the [`updateUrl()` function in utils.js](https://github.com/edx/edx-portal/blob/master/src/utils.js#L30) does not include `page_size` or any other query parameters that might exist in the url.

I don't know if you would want to address that in this PR or save it for another PR. We should probably fix that, though.

src/data/services/LmsApiService.js
```diff
@@ -56,6 +45,14 @@ class LmsApiService {
     });
   }
 
+  static fetchEnterpriseBySlug(slug) {
+    return this.fetchEnterpriseList({ slug })
+      // Because we expect only one enterprise by slug we return only the first result
+      .then(data => ({
+        data: data.data.results[0],
```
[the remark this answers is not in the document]
georgebabey: Good call. Worth defensively guarding for. Although [] is truthy so `data.data && data.data.results && data.data.results.length && data.data.results[0]`  What a mouthful, need that `?.` operator already

georgebabey:
@adamstankiewicz Updated, can you take another look?

src/data/actions/portalConfiguration.test.js
```diff
@@ -23,15 +33,33 @@ describe('actions', () => {
...
-        payload: { data: enterpriseData },
-      },
+      { type: FETCH_PORTAL_CONFIGURATION_REQUEST },
+      { type: FETCH_PORTAL_CONFIGURATION_SUCCESS, payload: { data: responseData } },
     ];
     const store = mockStore();
+    axiosMock.onGet(`http://localhost:18000/enterprise/api/v1/enterprise-customer/with_access_to/?page=1&page_size=50&permissions=enterprise_data_api_access&slug=${slug}`)
+      .replyOnce(200, JSON.stringify({ results: [responseData] }));
+
+
```
adamstankiewicz: nit: Extra line

src/containers/EnterpriseApp/index.jsx
```diff
@@ -71,9 +46,17 @@ class EnterpriseApp extends React.Component {
           />
           <Route exact path={`${baseUrl}/courses/:courseId`} component={CoursewarePage} />
           {features.DASHBOARD_V2 ? (
-            <Route exact path={`${baseUrl}/admin/learners/:slug?`} component={AdminPage} />
+            <Route
+              exact
+              path={`${baseUrl}/admin/learners/:actionSlug?`}
+              render={routeProps => <AdminPage {...routeProps} enterpriseSlug={enterpriseSlug} />}
```
adamstankiewicz: Would wrapping `AdminPage` with `withRouter` avoid needing to pass in `routeProps` and `enterpriseSlug`?

src/containers/EnterpriseApp/index.jsx
```diff
@@ -71,9 +46,17 @@ class EnterpriseApp extends React.Component {
...
+            />
           ) : (
-            <Route exact path={`${baseUrl}/admin/learners`} component={AdminPage} />
+            <Route
+              exact
+              path={`${baseUrl}/admin/learners`}
+              render={routeProps => <AdminPage {...routeProps} enterpriseSlug={enterpriseSlug} />}
```
adamstankiewicz: Same as above.

PR #71, opened 2018-09-19
ENT-1228

irfanuddinahmad:


src/components/Admin/index.jsx
```diff
@@ -36,46 +37,64 @@ class Admin extends React.Component {
     const defaultData = {
       title: 'Full Report',
       component: <EnrollmentsTable />,
+      csvFetchMethod: 'downloadCsv',
```
georgebabey: Is the thought this would be the endpoint such as `EnterpriseApiDataService.fetchCourseEnrollmentsCsv`?  I think that would make the most sense, and if so I'd probably remove the `csvOptions` argument and just create multiple methods in the fetchCourseEnrollmentsCsv.

It will mean the top of this file will have a bunch of imports:
```
import {
  fetchCourseEnrollmentsCsv,
  fetchLearnersNotInCourseCsv,
  fetchLearnersEnrolledCsv,
....
} from '../../data/services/EnterpriseApiDataService';

This would keep all of the "logic" around the endpoints in the data service layer.

File: src/components/DownloadCsv/index.jsx [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+      <Button
+        className={['btn-outline-primary', 'download-btn']}
+        disabled={this.state.csvLoading}
+        label={
+          <span>
+            <Icon className={['fa', 'mr-2'].concat(downloadButtonIconClasses)} />
+            Download report (CSV)
+          </span>
+        }
+        onClick={() => csvFetchMethod(enterpriseId, csvOptions)}
```
georgebabey: I'd recommend we wire this up to redux by changing the current actions and reducers in `/actions/courseEnrollments.js` and `/reducers/courseEnrollments.js` to be generic CSV, similar to how the table.js files are generic handlers for the table.

The `/actions/courseEnrollments.js` handler would then take an additional parameter of the fetchMethod and call that:

```
const fetchCsv = (enterpriseId, fetchCsvMethod) => (
  (dispatch) => {
    dispatch(fetchCsvRequest());
    return fetchCsvMethod(enterpriseId)
      .then((response) => {
        saveAs(new Blob([response.data]), `${enterpriseId}_progress_report.csv`);
        dispatch(fetchCsvSuccess());
      })
      .catch((error) => {
        dispatch(fetchCsvFailure(error));
      });
  }
);
```

src/containers/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,19 @@
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+
+import fetchCsv from '../../data/actions/courseEnrollments';
+import DownloadCsv from '../../components/DownloadCsv';
+
+const mapStateToProps = state => ({
+  enterpriseId: state.portalConfiguration.enterpriseId,
```
adamstankiewicz: I don't think the `enterpriseId` is necessary for the `DownloadCsv` container. The service layer needs it but you can get the enterprise id like [how it's done here](https://github.com/edx/edx-portal/blob/902d80ef9d2fdc99590fadcfb7b3c4223566b2dd/src/data/services/EnterpriseDataApiService.js#L14).

src/containers/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,19 @@
...
+  csvError: state.courseEnrollments.csvError,
+});
+
+const mapDispatchToProps = dispatch => ({
+  fetchCsvForMethod: (enterpriseId, csvFetchMethod) => {
+    dispatch(fetchCsv(enterpriseId, csvFetchMethod));
+  },
+});
+
+export default withRouter(connect(mapStateToProps, mapDispatchToProps)(DownloadCsv));
```
adamstankiewicz: Does the `DownloadCsv` component need the router props? If not, we shouldn't include `withRouter` here.

src/containers/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,19 @@
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+
+import fetchCsv from '../../data/actions/courseEnrollments';
+import DownloadCsv from '../../components/DownloadCsv';
+
+const mapStateToProps = state => ({
```
adamstankiewicz: The `mapStateToProps` for this container should have a similar implementation to [how it's done for the table](https://github.com/edx/edx-portal/blob/8b672d5975aeeb94237cf1ea7f46f6bba4fef4c9/src/containers/TableContainer/index.jsx#L6), where all download buttons' loading/error states will be contained in a single place in the Redux store. 

The resulting store should look something like:
```javascript
{
  ...
  table: {
    enrollments: { ... },
   ...
  },
  csv: {
    enrollments: {
      error: null,
      loading: false,
    }
    ...
  }
}
```

src/containers/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,19 @@
...
+  enterpriseId: state.portalConfiguration.enterpriseId,
+  csvLoading: state.courseEnrollments.csvLoading,
+  csvError: state.courseEnrollments.csvError,
+});
+
+const mapDispatchToProps = dispatch => ({
+  fetchCsvForMethod: (enterpriseId, csvFetchMethod) => {
```
adamstankiewicz: The `mapDispatchToProps` for this container should have a similar implementation to [how it's done for the table](https://github.com/edx/edx-portal/blob/8b672d5975aeeb94237cf1ea7f46f6bba4fef4c9/src/containers/TableContainer/index.jsx#L18). 

We should be passing `ownProps.id` and `ownProps.fetchMethod` into the `fetchCsv` method.

nit: Rename `fetchCsvForMethod` to `fetchCsv`.

src/components/Admin/index.jsx
```diff
@@ -11,6 +11,8 @@ import StatusAlert from '../../components/StatusAlert';
 import LoadingMessage from '../../components/LoadingMessage';
 import EnrollmentsTable from '../EnrollmentsTable';
 import AdminCards from '../../containers/AdminCards';
+import DownloadCsv from '../../components/DownloadCsv';
+
```
adamstankiewicz: nit: Extra line

src/components/Admin/index.jsx
```diff
@@ -230,6 +241,7 @@ class Admin extends React.Component {
                   </div>
                   <div className="col-12 col-md-6 text-md-right">
                     {this.renderDownloadCsvButton()}
```
adamstankiewicz: We should remove `renderDownloadCsvButton` and the call to it here now that it's in its own component.

src/components/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+DownloadCsv.defaultProps = {
+  enterpriseId: null,
+  csvError: null,
+  csvLoading: false,
+};
+
+DownloadCsv.propTypes = {
+  fetchCsvForMethod: PropTypes.func.isRequired,
+  csvFetchMethod: PropTypes.string.isRequired,
+  enterpriseId: PropTypes.string,
```
adamstankiewicz: As mentioned in another comment, we can probably do away with enterpriseId prop here.

src/components/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+  render() {
+    const {
+      csvError,
+    } = this.props;
+
+    return (
+      <div className="col">
+        <div className="row">
+          <div className="col-12 col-md-6 text-md-right">
+            {this.renderDownloadCsvButton()}
```
adamstankiewicz: We should only return the JSX that's in `renderDownloadCsvButton` here (i.e., strip out all the surrounding `div`s as we don't want this component to assume any particular layouts).

src/components/DownloadCsv/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+          </div>
+        </div>
+        {csvError && this.renderCsvErrorMessage()}
```
adamstankiewicz: Similar to the comment above, the `DownloadCsv` component shouldn't render out the error message. Layout and the rendering of the CSV error message should be handled by the `Admin` component like it does now.

src/components/Admin/index.jsx
```diff
@@ -42,52 +44,78 @@ class Admin extends React.Component {
...
         component: <EnrolledLearnersTable />,
+        csvFetchMethod: EnterpriseDataApiService.fetchCourseEnrollmentsCsv,
+        csvButtonId: 'enrolled-learners',
       },
       unenrolled: {
         title: 'Learners Not Enrolled in an Active Course',
         description: 'Learners who have completed all of their courses and/or courses have ended.',
         component: <EnrollmentsTable />,
+        csvFetchMethod: EnterpriseDataApiService.fetchCourseEnrollmentsCsv,
+        csvButtonId: 'un-enrollments',
```
adamstankiewicz: nit: `unenrollments`

src/components/Admin/index.jsx
```diff
@@ -42,52 +44,78 @@ class Admin extends React.Component {
...
         component: <LearnerActivityTable id="active-week" activity="active_past_week" />,
+        csvFetchMethod: EnterpriseDataApiService.fetchCourseEnrollmentsCsv,
+        csvButtonId: 'learner-activity',
       },
       'inactive-week': {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Not Active in Past Week',
         component: <LearnerActivityTable id="inactive-week" activity="inactive_past_week" />,
+        csvFetchMethod: EnterpriseDataApiService.fetchCourseEnrollmentsCsv,
+        csvButtonId: 'in-active-enrollments',
```
adamstankiewicz: nit: `inactive-enrollments`

src/components/DownloadCsvButton/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+DownloadCsvButton.defaultProps = {
+  csvLoading: false,
+  fetchMethod: () => {},
+};
+
+DownloadCsvButton.propTypes = {
+  fetchCsv: PropTypes.func.isRequired,
+  fetchMethod: PropTypes.func,
+  csvLoading: PropTypes.bool,
+  id: PropTypes.string.isRequired, // eslint-disable-line react/no-unused-prop-types
```
adamstankiewicz: Remove `id` prop type if it's unused.

src/containers/DownloadCsvButton/DownloadCsvButton.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import DownloadCsvButton from './index';
+import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
+
+const mockStore = configureMockStore([thunk]);
+const store = mockStore({
+  portalConfiguration: {
+    enterpriseId: 'test-enterprise',
+  },
+  dashboardAnalytics: {
```
adamstankiewicz: Do we need `dashboardAnalytics` to test `DownloadCsvButton`?

PR #73, opened 2018-09-20
Adding EnrolledLearners and RegisteredLearners table

adamstankiewicz:
* Adds 2 new tables (and associated tests) for:
  * Unenrolled, but registered learners ([ENT-1164](https://openedx.atlassian.net/browse/ENT-1164));
  * Learners enrolled in 1+ courses ([ENT-1165](https://openedx.atlassian.net/browse/ENT-1165));
* Some refactoring around tests to rid our use of `ContextProvider`;
* Logic in the `paginateTable` action to handle a 404 API response to return empty results instead of the error.


src/data/services/EnterpriseDataApiService.js
```diff
@@ -48,6 +48,45 @@ class EnterpriseDataApiService {
       },
     });
   }
+
+  static fetchUnerolledRegisteredLearners(options) {
+    const { enterpriseId } = store.getState().portalConfiguration;
+    const queryParams = {
+      page: 1,
+      page_size: 50,
```
brittneyexline: nit: should we make a constant for a default page_size?

adamstankiewicz: @brittneyexline Potentially, though, I think that could go in it's own PR.

PR #74, opened 2018-09-24
Learner activity details

mushtaqak:
This PR adds front-end functionality for Learner Activity cards. They include:
1. Active learners in past week.
2. Inactive learners in past week.
3. Inactive learners in past month.

Related enterprise-data PR: https://github.com/edx/edx-enterprise-data/pull/64

[ENT-1166](https://openedx.atlassian.net/browse/ENT-1166)
[ENT-1218](https://openedx.atlassian.net/browse/ENT-1218)

### DIFF
--- a/src/components/Admin/index.jsx
+++ b/src/components/Admin/index.jsx
@@ -14,6 +14,8 @@ import RegisteredLearnersTable from '../RegisteredLearnersTable';
 import EnrolledLearnersTable from '../EnrolledLearnersTable';
 import CompletedLearnersTable from '../CompletedLearnersTable';
 import PastWeekPassedLearnersTable from '../PastWeekPassedLearnersTable';
+import LearnerActivityTable from '../LearnerActivityTable';
+
 import AdminCards from '../../containers/AdminCards';
 
 import { formatTimestamp } from '../../utils';
@@ -60,17 +62,17 @@ class Admin extends React.Component {
       active: {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Top Active Learners',
-        component: <EnrollmentsTable />,
+        component: <LearnerActivityTable id="active-week" activity="active_past_week" />,
       },
       'inactive-week': {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Not Active in Past Week',
-        component: <EnrollmentsTable />,
+        component: <LearnerActivityTable id="inactive-week" activity="inactive_past_week" />,
       },
       'inactive-month': {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Not Active in Past Month',
-        component: <EnrollmentsTable />,
+        component: <LearnerActivityTable id="inactive-month" activity="inactive_past_month" />,
       },
       completed: {
         title: 'Number of Courses Completed by Learner',
--- a/src/components/LearnerActivityTable/index.jsx
+++ b/src/components/LearnerActivityTable/index.jsx
@@ -1,11 +1,13 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import TableContainer from '../../containers/TableContainer';
 import { formatTimestamp, formatPassedTimestamp, formatPercentage } from '../../utils';
 import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
 
-const EnrollmentsTable = () => {
-  const enrollmentTableColumns = [
+const LearnerActivityTable = (props) => {
+  const { activity, id } = props;
+  const tableColumns = [
     {
       label: 'Email',
       key: 'user_email',
@@ -48,7 +50,14 @@ const EnrollmentsTable = () => {
     },
   ];
 
-  const formatEnrollmentData = enrollments => enrollments.map(enrollment => ({
+  const getTableColumns = () => {
+    if (activity !== 'active_past_week') {
+      return tableColumns.filter(column => column.key !== 'passed_timestamp');
+    }
+    return tableColumns;
+  };
+
+  const formatTableData = enrollments => enrollments.map(enrollment => ({
     ...enrollment,
     last_activity_date: formatTimestamp({ timestamp: enrollment.last_activity_date }),
     course_start: formatTimestamp({ timestamp: enrollment.course_start }),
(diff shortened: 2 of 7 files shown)


src/components/ActiveLearnersTable/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    {
+      label: 'Last Activity Date',
+      key: 'last_activity_date',
+      columnSortable: true,
+    },
+  ];
+
+
+  const formatPassedTimestamp = (timestamp) => {
```
adamstankiewicz: Because this function is now shared between more than one table, it may be a good idea to move it to `utils.js` as a utility function.

src/components/Admin/index.jsx
```diff
@@ -59,17 +63,17 @@ class Admin extends React.Component {
...
         title: 'Learners Enrolled in a Course',
         subtitle: 'Top Active Learners',
-        component: <EnrollmentsTable />,
+        component: <ActiveLearnersTable />,
       },
       'inactive-week': {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Not Active in Past Week',
-        component: <EnrollmentsTable />,
+        component: <InactivePastWeekLearnersTable />,
```
adamstankiewicz: Because the `InactivePastWeekLearnersTable` and `InactivePastMonthLearnersTable` share the same columns, I wonder if we might want to make a single component `InactiveLearnersTable` that takes in props to determine which version (e.g., past week or month) to use. The table could default to fetching past week's inactive learners, and have a boolean prop for `pastMonth`. For example,

```javascript
{
  'inactive-week': {
    ...,
    component: <InactiveLearnersTable />, // fetch inactive learners in past week by default
  },
  'inactive-month': {
    ...,
    component: <InactiveLearnersTable pastMonth />, // fetch inactive learners in past month due to pastMonth prop
  },
}
```

If we do this, the `InactiveLearnersTable` component would simply use the appropriate `fetchMethod` supplied to `TableContainer` based on the `pastMonth` prop, e.g.:

```javascript
getFetchMethod() {
  if (this.props.pastMonth) {
    return EnterpriseDataApiService.fetchInactivePastMonthLearners;
  }
  return  EnterpriseDataApiService.fetchInactivePastWeekLearners;
}
```

```jsx
<TableContainer
  ...
  fetchMethod: this.getFetchMethod(),
/>
```

src/components/InactivePastMonthLearnersTable/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+      key: 'course_start',
+      columnSortable: true,
+    },
+    {
+      label: 'End Date',
+      key: 'course_end',
+      columnSortable: true,
+    },
+    {
+      label: 'Passed Date',
```
adamstankiewicz: I believe the inactive tables (both past week & past month) should not have the `Passed Date` table column per the "No has passed column" in the ticket.

src/components/InactivePastMonthLearnersTable/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+      timestamp: enrollment.user_account_creation_timestamp,
+    }),
+    has_passed: enrollment.has_passed ? 'Yes' : 'No',
+    course_price: enrollment.course_price ? `$${enrollment.course_price}` : '',
+    current_grade: formatPercentage({ decimal: enrollment.current_grade }),
+  }));
+
+  return (
+    <TableContainer
+      id="inactive-past-month-learners"
```
adamstankiewicz: If we combine the table components for the inactive learners for past week & past month into a single `InactiveLearnersTable`, we'll need to change the `TableContainer`'s id and className here depending on the `pastMonth` prop passed in (assuming we default to past week).

src/components/Admin/index.jsx
mushtaqak: So I guess, active and inactive past week and past Month can just be merged into componen, with different fetch method. I will let you know Tommorow when I address this feedback ✌️

adamstankiewicz: @mushtaqak The inactive past week/month can be merged into single component but not the active week table. the columns for those tables are not exactly the same as the active learners per the ticket. The inactive week/month tables don't show any data related to passing (no "Passed Date" column).

Though, I suppose you could have all three tables derived from the same component and modify the columns based on the prop passed in.

src/components/Admin/index.jsx
```diff
@@ -63,17 +61,17 @@ class Admin extends React.Component {
       active: {
         title: 'Learners Enrolled in a Course',
         subtitle: 'Top Active Learners',
-        component: <ActiveLearnersTable />,
+        component: <LearnerActivityTable id='active-week' activity='active_past_week'/>,
```
adamstankiewicz: nit: Props with strings should have double quotes, e.g., `id="active-week"`.

src/components/LearnerActivityTable/index.jsx
```diff
@@ -75,14 +74,14 @@ const InactivePastMonthLearnersTable = () => {
 
   return (
     <TableContainer
-      id="inactive-past-month-learners"
-      className="inactive-past-month-learners"
-      fetchMethod={EnterpriseDataApiService.fetchInactivePastMonthLearners}
-      columns={tableColumns}
+      id={props.id}
```
adamstankiewicz: nit: You could use es6 object deconstruction here by pulling out `id` and `activity` from the props above where we create the `tableColumns` variable:

```javascript
const { activity, id } = props;

const tableColumns = [...];
```

src/components/LearnerActivityTable/index.jsx
```diff
@@ -48,12 +48,11 @@ const InactivePastMonthLearnersTable = () => {
     },
   ];
 
-
-  const formatPassedTimestamp = (timestamp) => {
-    if (timestamp) {
-      return formatTimestamp({ timestamp });
+  const getTableColumns = props => {
```
adamstankiewicz: You don't actually need to pass in `props` to this function as you should already have them in the outer scope ([the props passed into the component itself](https://github.com/edx/edx-portal/pull/74/commits/aede8b733ad14b747f807f7502718eaced71b69a#diff-9510153a701fb5abeb9028052670ee72R7)).

src/components/LearnerActivityTable/index.jsx
```diff
@@ -48,12 +48,11 @@ const InactivePastMonthLearnersTable = () => {
...
+    if (props.activity !== 'active_past_week') {
+      return tableColumns.filter((column)=> (column.key != 'passed_timestamp'))
```
adamstankiewicz: Nice!

src/components/LearnerActivityTable/index.jsx
```diff
@@ -75,14 +74,14 @@ const InactivePastMonthLearnersTable = () => {
...
+      fetchMethod={() => EnterpriseDataApiService.fetchCourseEnrollments({learner_activity: props.activity})}
+      columns={getTableColumns(props)}
       formatData={formatTableData}
       tableSortable
     />
   );
 };
 
-export default InactivePastMonthLearnersTable;
+export default LearnerActivityTable;
```
adamstankiewicz: Because this component accepts the `id` and `activity` props, we should include the `PropTypes` validation for this component:

```javascript
LearnerActivityTable.propTypes = {
  id: PropTypes.string.isRequired,
  activity: PropTypes.string.isRequired,
};
```

PR #75, opened 2018-09-24
Support temporarily enabling feature flags with query parameter

adamstankiewicz:
If some features are behind a feature flag (e.g., the collapsible cards), it's impossible to see and verify changes on stage and prod. This PR supports temporarily enabling one or more feature flags by passing a `features` URL query parameter. To have more than one feature enabled, simply put them in a comma-separated list:

```
http://localhost:1991/enterprise-slug/admin/learners?features=DASHBOARD_V2

http://localhost:1991/enterprise-slug/admin/learners?features=DASHBOARD_V2,OTHER_FEATURE
```

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -29,12 +29,13 @@ const updateUrl = (data) => {
     })}`);
   }
 };
+
 const removeTrailingSlash = path => path.replace(/\/$/, '');
 
 export {
-  formatTimestamp,
   formatPercentage,
+  formatTimestamp,
   getAccessToken,
-  updateUrl,
   removeTrailingSlash,
+  updateUrl,
 };
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -1,3 +1,10 @@
+import qs from 'query-string';
+
+const hasFeatureFlagEnabled = (featureFlag) => {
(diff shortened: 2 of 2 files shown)


src/utils.js
```diff
@@ -29,12 +29,19 @@ const updateUrl = (data) => {
...
+
 const removeTrailingSlash = path => path.replace(/\/$/, '');
 
+const hasFeatureFlagEnabled = (featureFlag) => {
+  const { features } = qs.parse(window.location.search);
+  return features && features.split(',').includes(featureFlag);
+};
+
 export {
-  formatTimestamp,
```
adamstankiewicz: Note to reviewers: reordering the exports to be in alphabetical order.

src/utils.js
```diff
@@ -29,12 +29,19 @@ const updateUrl = (data) => {
     })}`);
   }
 };
+
 const removeTrailingSlash = path => path.replace(/\/$/, '');
 
+const hasFeatureFlagEnabled = (featureFlag) => {
```
georgebabey: utils feels like a dumping ground and I don't think you'd want others to use this outside of config. I would recommend keeping this private `config/index.js`

PR #78, opened 2018-10-01
Demo data for M2

mushtaqak:
This PR adds demo data for M2.

[ENT-1227](https://openedx.atlassian.net/browse/ENT-1227)

### DIFF
--- a/src/demo/data.js
+++ b/src/demo/data.js
@@ -2,51 +2,178 @@ import faker from 'faker/locale/en';
 import courses from './courses';
 
 const numEnrollments = 330;
-const percentPassed = 0.70;
-const percentActive = 0.35;
+const numPassed = 120;
 
-const enrollments = [...Array(numEnrollments)].map((_, idx) => {
-  // Although we don't pass first & lastname, this makes our emails more consistent
-  const firstName = faker.name.firstName();
-  const lastName = faker.name.lastName();
-  const provider = 'bestrun.com';
+const getEnrollments = () => {
+  const fromPastWeekDate = new Date();
+  const fromPastMonthDate = new Date();
+  const toPastMonthDate = new Date();
+  fromPastWeekDate.setDate(fromPastWeekDate.getDate() - 7);
+  fromPastMonthDate.setDate(fromPastMonthDate.getDate() - 30);
+  toPastMonthDate.setDate(toPastMonthDate.getDate() - 60);
+  const dateInPastMonths = faker.date.between(fromPastMonthDate, toPastMonthDate);
+  const dateInPastMonth = faker.date.between(fromPastWeekDate, fromPastMonthDate);
 
-  const course = courses[faker.random.number({ max: courses.length - 1 })];
-  const courseStart = faker.date.past();
-  const courseEnd = faker.date.future(null, courseStart);
-  const hasPassed = Math.random() <= percentPassed;
-  const grade = hasPassed ? faker.random.number({ min: 70, max: 100 }) :
-    faker.random.number({ max: 70 });
+  let enrollmentsData = [...Array(numEnrollments)].map((_, idx) => {
+    // Although we don't pass first & lastname, this makes our emails more consistent
+    const firstName = faker.name.firstName();
+    const lastName = faker.name.lastName();
+    const provider = 'bestrun.com';
 
-  // TODO: we could generate all columns so we could easily fake the csv call as well
-  return {
-    id: idx,
-    user_email: faker.internet.email(firstName, lastName, provider),
-    course_title: course.title,
-    course_price: course.price,
-    course_start: courseStart,
-    course_end: courseEnd,
-    passed_timestamp: Math.random() <= hasPassed ? faker.date.between(courseStart, courseEnd)
-      : null,
-    has_passed: hasPassed,
-    current_grade: grade / 100,
-    last_activity_date: faker.date.recent(Math.random() <= percentActive ? 7 : 45),
-  };
-});
+    const course = courses[faker.random.number({ max: courses.length - 1 })];
+    const courseStart = faker.date.past();
+    const courseEnd = faker.date.future();
+    const hasPassed = idx < numPassed;
+    const grade = hasPassed ? faker.random.number({ min: 70, max: 100 }) :
+      faker.random.number({ max: 70 });
+    const lastActivityDate = hasPassed ? faker.date.recent(7) : dateInPastMonths;
 
-const overview = {
-  // We could derive this data from the enrollments but for now just using static data
-  active_learners: {
-    past_month: 111,
-    past_week: 48,
-  },
-  course_completions: 85,
-  enrolled_learners: 203,
-  last_update_date: new Date(),
-  number_of_users: 268,
+    return {
+      id: idx,
+      user_email: faker.internet.email(firstName, lastName, provider),
+      consent_granted: true,
+      course_title: course.title,
+      course_price: course.price,
+      course_start: courseStart,
+      course_end: courseEnd,
+      passed_timestamp: hasPassed ? faker.date.recent(7)
+        : null,
+      has_passed: hasPassed,
+      current_grade: grade / 100,
+      last_activity_date: lastActivityDate,
+    };
+  });
+
+  // Add more records to generate data for different tables.
+  const enrollmentsSample = enrollmentsData.slice(0, 5);
+  enrollmentsData = [
+    ...enrollmentsData,
+    ...enrollmentsSample.map(enrollment => ({
+      ...enrollment,
+      last_activity_date: dateInPastMonth,
+    })),
+    ...enrollmentsSample.map(enrollment => ({
+      ...enrollment,
+      consent_granted: false,
+    })),
+    ...enrollmentsSample.map(enrollment => ({
+      ...enrollment,
+      passed_timestamp: dateInPastMonths,
+      last_activity_date: dateInPastMonths,
+    })),
+    ...enrollmentsSample.map(enrollment => ({
+      ...enrollment,
+      course_end: dateInPastMonth,
+    })),
+  ];
+  return enrollmentsData;
 };
 
+const allEnrollments = getEnrollments();
+const enrollments = allEnrollments.filter(enrollment => enrollment.consent_granted === true);
+const completedEnrollments = enrollments.filter(enrollment => enrollment.has_passed === true);
+
+const getEnrollmentsForUser = email => (
+  enrollments.filter(enrollment => enrollment.user_email === email)
+);
+
+const getCompletedEnrollmentsForUser = email => (
+  completedEnrollments.filter(enrollment => enrollment.user_email === email)
+);
+
+const filterEnrolledLearners = () => enrollments.map(enrollment => ({
+  user_email: enrollment.user_email,
+  user_account_creation_timestamp: faker.date.past(),
+  enrollment_count: getEnrollmentsForUser(enrollment.user_email).length,
+}));
+
+const filterEnrolledLearnersForInactiveCourses = () => enrollments.filter(enrollment => (
+  enrollment.course_end.getTime() < new Date().getTime()
+)).map(enrollment => ({
+  user_email: enrollment.user_email,
+  enrollment_count: getEnrollmentsForUser(enrollment.user_email).length,
+  course_completion_count: getCompletedEnrollmentsForUser(enrollment.user_email).length,
+  last_activity_date: enrollment.last_activity_date,
+}));
+
+const filterUnenrolledRegisteredLearners = () => enrollments.map(enrollment => ({
+  user_email: enrollment.user_email,
+  user_account_creation_timestamp: faker.date.past(),
+}));
+
+const filterLearnerActivity = (filterOption) => {
+  let filteredLearnerEnrollments = enrollments;
+  const lastActivityDate = new Date();
+  if (filterOption === 'active_past_week') {
+    lastActivityDate.setDate(lastActivityDate.getDate() - 7);
+    filteredLearnerEnrollments = enrollments.filter(enrollment => (
+      enrollment.last_activity_date.getTime() >= lastActivityDate.getTime()
+    ));
+  } else if (filterOption === 'inactive_past_week') {
+    lastActivityDate.setDate(lastActivityDate.getDate() - 7);
+    filteredLearnerEnrollments = enrollments.filter(enrollment => (
+      enrollment.last_activity_date.getTime() <= lastActivityDate.getTime()
+    ));
+  } else if (filterOption === 'inactive_past_month') {
+    lastActivityDate.setDate(lastActivityDate.getDate() - 30);
+    filteredLearnerEnrollments = enrollments.filter(enrollment => (
+      enrollment.last_activity_date.getTime() <= lastActivityDate.getTime()
+    ));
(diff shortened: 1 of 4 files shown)


src/demo/data.js
```diff
@@ -34,19 +34,66 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
+  return enrollments.filter((enrollment) => {
+    return enrollment.has_passed &&
+            enrollment.passed_timestamp.getTime() >= passedDate.getTime() &&
+            enrollment.passed_timestamp.getTime() <= today.getTime();
+  });
+};
+
+const filterLearnerActivity = (enrollments, filterOption) => {
+  if (filterOption === 'active_past_week') {
+    let lastActivityDate = faker.date.recent(7);
```
georgebabey: @mushtaqak This is your problem - you shouldn't use `faker.date.recent` as that will return a random date sometime in the last 7 days. Instead use `let weekAgo = new Date(); weekAgo.setDate(weekAgo.getDate() - 2);`

mushtaqak: Exactly, I found the issue as well. Couldn't push the code due to some other mess I had created 😌

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
     course_price: course.price,
     course_start: courseStart,
     course_end: courseEnd,
-    passed_timestamp: Math.random() <= hasPassed ? faker.date.between(courseStart, courseEnd)
+    passed_timestamp: hasPassed ? faker.date.recent(7)
```
adamstankiewicz: Does it make sense that the `passed_timestamp` is restricted to the past 7 days? Technically, the passed timestamp could be anywhere from course start to course end dates. I imagine you're setting it to the past 7 days such that the data contains at least some learners who have passed in the past week for the reports that filter in the past week. Would something like this make more sense?:

```javascript
const percentPassedPastWeek = 0.25;

return {
  ...
  passed_timestamp: Math.random() <= percentPassedPastWeek ? fake.date.recent(7) : faker.date.between(courseStart, courseEnd),
  ...
}
```

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
+    last_activity_date: lastActivityDate,
   };
 });
 
+const filterEnrolledLearners = () => {
+  return enrollments.map((enrollment) => {
+    return {
+      user_email: enrollment.user_email,
+      user_account_creation_timestamp: faker.date.past(),
+      enrollment_count: faker.random.number(1, 10), // TODO: another way to populate data?
```
adamstankiewicz: Should we have the enrollment count range include `0` to represent when some users don't have any enrollments?:

```javascript
faker.random.number(0, 10),
```

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
+    lastActivityDate.setDate(lastActivityDate.getDate() - 30);
+    filteredLearnerEnrollments = learnerEnrollments.filter(enrollment => enrollment.last_activity_date.getTime() <= lastActivityDate.getTime());
+  }
+  return filteredLearnerEnrollments;
+};
+
+const filterPastWeekCompletions = (pastWeekEnrollments) => {
+  const today = new Date();
+  const passedDate = new Date();
+  passedDate.setDate(today.getDate() - 7);
```
adamstankiewicz: This could be reduced to a single line:
```javascript
const passedDate = new Date().setDate(today.getDate() - 7);
```

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
+  return pastWeekEnrollments.filter((enrollment) => {
+    return enrollment.has_passed &&
+            enrollment.passed_timestamp.getTime() >= passedDate.getTime() &&
+            enrollment.passed_timestamp.getTime() <= today.getTime();
```
adamstankiewicz: nit: Extra indentation on the last 2 parts of this conditional:

```javascript
return pastWeekEnrollments.filter((enrollment) => {
  return enrollment.has_passed &&
    enrollment.passed_timestamp.getTime() >= passedDate.getTime() &&
    enrollment.passed_timestamp.getTime() >= today.getTime();
});
```

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
+  }
+  return filterEnrollments;
+};
+
 const overview = {
-  // We could derive this data from the enrollments but for now just using static data
   active_learners: {
-    past_month: 111,
-    past_week: 48,
+    past_month: filterPastWeekCompletions(enrollments).length,
```
adamstankiewicz: Nice to derive the data from the enrollments data! 👍

src/demo/data.js
```diff
@@ -26,27 +30,112 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
-  number_of_users: 268,
+  number_of_users: enrollments.length,
+};
+
+const getEnrollmentsCsv = () => {
+  // Get column names.
+  let csvData = `${Object.keys(enrollments[0]).toString()}\n`;
+
+  // Get data
+  for (const enrollment of enrollments) { // TODO:
```
adamstankiewicz: Is there still a `TODO` here?

src/demo/rewireEnterpriseDataApiService.js
```diff
@@ -1,7 +1,37 @@
 import EnterpriseDataApiService from '../../src/data/services/EnterpriseDataApiService';
-import { enrollments, overview } from './data';
+import { filterCompletedLearnerCourses, filteredEnrollments, filterEnrolledLearners, filterUnenrolledRegisteredLearners, getEnrollmentsCsv, overview } from './data';
```
adamstankiewicz: nit: This could be cleaned up a bit by having each import on its own line:

```javascript
import {
  filterCompletedLearnerCourses,
  filteredEnrollments,
  filteredEnrolledLearners,
  ...
} from './data';
```

src/demo/data.js
```diff
@@ -14,9 +13,14 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
   const course = courses[faker.random.number({ max: courses.length - 1 })];
   const courseStart = faker.date.past();
   const courseEnd = faker.date.future(null, courseStart);
-  const hasPassed = Math.random() <= percentPassed;
+  const hasPassed = idx < numPassed;
```
adamstankiewicz: Out of curiosity, is there a reason why you're changing `hasPassed` to be based on `numPassed` instead of the following?:

``` javascript
Math.random() <= percentPassed
```

src/demo/data.js
```diff
@@ -14,9 +13,14 @@ const enrollments = [...Array(numEnrollments)].map((_, idx) => {
...
   const grade = hasPassed ? faker.random.number({ min: 70, max: 100 }) :
     faker.random.number({ max: 70 });
+  const fromPastMonthDate = new Date();
+  const toPastMonthDate = new Date();
+  fromPastMonthDate.setDate(fromPastMonthDate.getDate() - 30);
+  toPastMonthDate.setDate(toPastMonthDate.getDate() - 60);
+  const lastActivityDate = hasPassed ? faker.date.recent(7) : faker.date.between(fromPastMonthDate, toPastMonthDate);
```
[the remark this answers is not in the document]
mushtaqak: It doesn't actually makes sense. I just made like this so that we get at-least same data overtime. Previously, date received was not the same.

src/demo/data.js
mushtaqak: I did this for the same reason so that data is consistent.

src/demo/data.js
mushtaqak: I did this for the same reason so that data is consistent. First 120 enrolments get passed :)

mushtaqak:
@adamstankiewicz  @georgebabey This is ready for review now.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
-  course_completions: 85,
-  enrolled_learners: 203,
+  course_completions: filterCompletedLearnerCourses().length,
+  enrolled_learners: filterEnrolledLearners().length,
   last_update_date: new Date(),
-  number_of_users: 268,
+  number_of_users: enrollments.length,
 };
 
+const getEnrollmentsCsv = () => enrollments.reduce((csvData, enrollment) => `${csvData}${Object.values(enrollment).toString()}\n`, `${Object.keys(enrollments[0]).toString()}\n`); // eslint-disable-line max-len
```
adamstankiewicz: Can this line be formatted to avoid using `// eslint-disable-line max-len`? It would probably also make it easier to read.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
+    ...enrollmentsSample.map(enrollment => ({ ...enrollment, consent_granted: false })), // eslint-disable-line max-len
+    ...enrollmentsSample.map(enrollment => ({ ...enrollment, passed_timestamp: dateInPastMonths, last_activity_date: dateInPastMonths })), // eslint-disable-line max-len
+    ...enrollmentsSample.map(enrollment => ({ ...enrollment, course_end: dateInPastMonth })), // eslint-disable-line max-len
+  ];
+  return enrollmentsData;
+};
+
+const enrollments = getEnrollments();
+
+const filterEnrolledLearners = () => enrollments.filter(enrollment => enrollment.consent_granted === true).map((enrollment) => { // eslint-disable-line max-len
```
adamstankiewicz: Can this line be formatted to avoid using `// eslint-disable-line max-len`? It would probably also make it easier to read.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
+  const totalCurrentEnrollements = enrollments.filter(currentEnrollment => currentEnrollment.user_email === enrollment.user_email).length; // eslint-disable-line max-len
+  return {
+    user_email: enrollment.user_email,
+    user_account_creation_timestamp: faker.date.past(),
+    enrollment_count: totalCurrentEnrollements,
+  };
+});
+
+const filterEnrolledLearnersForInactiveCourses = () => enrollments.filter(enrollment => enrollment.course_end.getTime() < new Date().getTime()).map((enrollment) => { // eslint-disable-line max-len
```
adamstankiewicz: Can this line be formatted to avoid using `// eslint-disable-line max-len`? It would probably also make it easier to read.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
+  user_email: enrollment.user_email,
+  user_account_creation_timestamp: faker.date.past(),
+}));
+
+const filterLearnerActivity = (learnerEnrollments, filterOption) => {
+  let filteredLearnerEnrollments = learnerEnrollments;
+  const lastActivityDate = new Date();
+  if (filterOption === 'active_past_week') {
+    lastActivityDate.setDate(lastActivityDate.getDate() - 7);
+    filteredLearnerEnrollments = learnerEnrollments.filter(enrollment => enrollment.last_activity_date.getTime() >= lastActivityDate.getTime()); // eslint-disable-line max-len
```
adamstankiewicz: Can this line (and the others in this function) be formatted to avoid using `// eslint-disable-line max-len`? It would probably also make it easier to read.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
+
+const filterPastWeekCompletions = (pastWeekEnrollments) => {
+  const today = new Date();
+  const passedDateTime = new Date().setDate(today.getDate() - 7);
+  return pastWeekEnrollments.filter(enrollment => enrollment.has_passed &&
+    enrollment.passed_timestamp.getTime() >= passedDateTime &&
+    enrollment.passed_timestamp.getTime() <= today.getTime());
+};
+
+const filterCompletedLearnerCourses = () => enrollments.filter(enrollment => enrollment.has_passed).map((enrollment) => { // eslint-disable-line max-len
```
adamstankiewicz: Can this line be formatted to avoid using `// eslint-disable-line max-len`? It would probably also make it easier to read.

src/demo/data.js
```diff
@@ -2,51 +2,153 @@ import faker from 'faker/locale/en';
...
+    ...enrollmentsSample.map(enrollment => ({ ...enrollment, passed_timestamp: dateInPastMonths, last_activity_date: dateInPastMonths })), // eslint-disable-line max-len
+    ...enrollmentsSample.map(enrollment => ({ ...enrollment, course_end: dateInPastMonth })), // eslint-disable-line max-len
+  ];
+  return enrollmentsData;
+};
+
+const enrollments = getEnrollments();
+
+const filterEnrolledLearners = () => enrollments.filter(enrollment => enrollment.consent_granted === true).map((enrollment) => { // eslint-disable-line max-len
+  const totalCurrentEnrollements = enrollments.filter(currentEnrollment => currentEnrollment.user_email === enrollment.user_email).length; // eslint-disable-line max-len
```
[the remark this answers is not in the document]
mushtaqak: bad typo 😆

mushtaqak:
@adamstankiewicz Fixed all `max-len` violations and addressed your feedback.

src/demo/data.js
```diff
@@ -2,51 +2,178 @@ import faker from 'faker/locale/en';
...
+    past_month: filterPastWeekCompletions().length,
+    past_week: filterLearnerActivity('active_past_week').length,
   },
-  course_completions: 85,
-  enrolled_learners: 203,
-  last_update_date: new Date(),
-  number_of_users: 268,
+  course_completions: filterCompletedLearnerCourses().length,
+  enrolled_learners: filterEnrolledLearners().length,
+  last_update_date: new Date().toString(),
```
adamstankiewicz: Thanks for adding the `toString()` here :) Per my previous comment, though, `last_update_date` should be `last_updated_date` as that is the key the API returns.

src/demo/rewireEnterpriseDataApiService.test.js
```diff
@@ -2,29 +2,68 @@ import EnterpriseDataApiService from '../../src/data/services/EnterpriseDataApiS
...
 
   afterEach(() => {
     // rewire in the tests overrides the methods on EnterpriseDataApiService so we restore
     // them to their original method calls after each test. Ideally we would just save/restore
     // EnterpriseDataApiService but imports are considered read-only.
+    EnterpriseDataApiService.fetchCompletedLearners = fetchCompletedLearners;
     EnterpriseDataApiService.fetchCourseEnrollments = fetchCourseEnrollments;
+    EnterpriseDataApiService.fetchEnrolledLearners = fetchEnrolledLearners;
+    EnterpriseDataApiService.fetchEnrolledLearnersForInactiveCourses = fetchEnrolledLearnersForInactiveCourses; // eslint-disable-line max-len
+    EnterpriseDataApiService.fetchUnerolledRegisteredLearners = fetchUnerolledRegisteredLearners;
```
adamstankiewicz: Have you rebased on `master` recently? The typo in this method name (`fetchUnerolledRegisteredLearners`) is fixed on `master`.

PR #82, opened 2018-10-09
Ensure all card action tables have a CSV fetch method

adamstankiewicz:
1. Ensures each card action table has an associated csv fetch method;
2. Refactors the `EnterpriseDataApiService` so that we don't have two methods (i.e., one for normal and one for csv) for every request; instead, supports passing a `csv` boolean flag to the fetch method for it to fetch the csv.
3. Refactors the logic around when the csv button is disabled:
    - Disables whenever the table data is loading (to prevent accidentally downloading a csv while a different table is loading);
    - Disables if there are no results in the table (no reason to let users download if there's no data to download).
4. Updates the `LearnerActivityTable` to clear it's associated Redux state whenever the `id` prop changes.
    - This is necessary because this component is used for 3 of the card action tables. When switching between these 3 tables, the state never gets cleared so the loading flags were not accurate.

### DIFF
--- a/src/components/Admin/Admin.test.jsx
+++ b/src/components/Admin/Admin.test.jsx
@@ -6,6 +6,7 @@ import { MemoryRouter } from 'react-router-dom';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
+import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
 import Admin from './index';
 
 import { features } from '../../config';
@@ -94,7 +95,7 @@ describe('<Admin />', () => {
               match={{
                 url: '/',
                 params: {
-                  actionSlug: 'registered',
+                  actionSlug: 'registered-unenrolled-learners',
                 },
               }}
             />
@@ -111,7 +112,7 @@ describe('<Admin />', () => {
               match={{
                 url: '/',
                 params: {
-                  actionSlug: 'enrolled',
+                  actionSlug: 'enrolled-learners',
                 },
               }}
             />
@@ -128,7 +129,7 @@ describe('<Admin />', () => {
               match={{
                 url: '/',
                 params: {
-                  actionSlug: 'no-current-courses',
+                  actionSlug: 'enrolled-learners-inactive-courses',
                 },
               }}
             />
@@ -145,7 +146,7 @@ describe('<Admin />', () => {
               match={{
                 url: '/',
                 params: {
-                  actionSlug: 'active',
+                  actionSlug: 'learners-active-week',
                 },
               }}
             />
@@ -162,7 +163,7 @@ describe('<Admin />', () => {
               match={{
                 url: '/',
                 params: {
-                  actionSlug: 'inactive-week',
+                  actionSlug: 'learners-inactive-week',
                 },
               }}
             />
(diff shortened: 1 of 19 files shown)


src/components/Admin/Admin.test.jsx
```diff
@@ -294,4 +295,222 @@ describe('<Admin />', () => {
...
   });
+
+  describe('calls download csv fetch method for table', () => {
+    let spy;
+
+    afterEach(() => {
+      spy.mockRestore();
+    });
+
+    it('enrollments', () => {
```
georgebabey: You could consider looping over fetchUrl/actionSlug pairs to avoid a lot of the duplication here. Might make it easier to read since theres a lot of boilerplate creating the AdminWrapper

src/components/Admin/Admin.test.jsx
```diff
@@ -294,4 +295,222 @@ describe('<Admin />', () => {
...
+          match={{
+            url: '/',
+            params: {
+              actionSlug: 'learners-inactive-month',
+            },
+          }}
+        />
+      ));
+      wrapper.find('.download-btn').simulate('click');
+      expect(spy).toHaveBeenCalled();
```
georgebabey: May want to make sure these are calledWith the proper parameters as fetchCourseEnrollments gets called multiple times... Could also make sure we're passing `{ csv: true }`

src/components/Admin/index.jsx
```diff
@@ -112,10 +129,24 @@ class Admin extends React.Component {
...
+  }
+
+  shouldDisableCsvButton(id) {
+    const tableData = this.getTableData(id);
+    if (!tableData) {
+      return true;
     }
-    return '';
+    const isTableLoading = tableData.loading;
+    const isTableEmpty = tableData.results && !tableData.results.length;
```
georgebabey: Was the empty table check necessary? Looks like it is already captured on the row with `this.hasEmptyData()` https://github.com/edx/edx-portal/pull/82/files#diff-a3ca57fe250c904aa8b89703ffc82c38R262

Doesn't hurt to have it here too, just curious.

src/components/Admin/index.jsx
```diff
@@ -194,8 +225,8 @@ class Admin extends React.Component {
     } = this.props;
 
     const { params: { actionSlug } } = match;
-    const tableData = this.getMetadataForAction(actionSlug);
-    const csvErrorMessage = this.getCsvErrorMessage(tableData.csvButtonId);
+    const tableMetadata = this.getMetadataForAction(actionSlug);
```
georgebabey: 👍

src/components/Admin/Admin.test.jsx
adamstankiewicz: That's probably a good idea to check that the proper parameters are being passed. Though, I think because we have `spy.mockRestore()` in the `afterEach()`, checking `toHaveBeenCalled()` for `fetchCourseEnrollments` each time it's called should check that it's being called each time.

I will change this to `toHaveBeenCalledWith` and check the params.

src/components/Admin/index.jsx
adamstankiewicz: `this.hasEmptyData()` is checking if the overview data is empty, not the table. Some of the table data could be empty even if the overview data is not empty. Therefore, it's necessary to check if the table data is empty on its own to determine whether the CSV button should be disabled.

src/components/Admin/Admin.test.jsx
georgebabey: Sorry, I meant its called from different tests/scenarios with different parameters. You got it :)

PR #84, opened 2018-10-15
Fix failing test for demo data

adamstankiewicz:
I noticed this test on `master` was failing after @mushtaqak merged his demo data changes. The reason why it was failing is because the `last_updated_date` was a string, but we were checking to verify it as a `Date` object. `last_updated_date` changed to be a string when I had Mushtaq add `.toString()` to the date as the last updated date wasn't showing in the UI otherwise.

Also, this test had the key `last_update_date`, not `last_updated_date`.

Still TBD how the test was passing on @mushtaqak branch.

### DIFF
--- a/src/demo/rewireEnterpriseDataApiService.test.js
+++ b/src/demo/rewireEnterpriseDataApiService.test.js
@@ -102,7 +102,7 @@ describe('rewireEnterpriseDataApiService', () => {
         },
         course_completions: expect.any(Number),
         enrolled_learners: expect.any(Number),
-        last_update_date: expect.any(Date),
+        last_updated_date: expect.any(String),
         number_of_users: expect.any(Number),
       };
       expect(results.data).toEqual(expectedResults);


georgebabey:
Update: looks like the tests didn't actually pass but are still existing with passing status code. I'll work on updating.. https://travis-ci.org/edx/edx-portal/builds/441691708?utm_source=github_status&utm_medium=notification

src/demo/rewireEnterpriseDataApiService.test.js
```diff
@@ -102,7 +102,7 @@ describe('rewireEnterpriseDataApiService', () => {
         },
         course_completions: expect.any(Number),
         enrolled_learners: expect.any(Number),
-        last_update_date: expect.any(Date),
+        last_updated_date: expect.any(String),
```
mushtaqak: my bad, missed this file :grimacing: 

However, strange how did this test pass :confused: ?

georgebabey: @mushtaqak Test didn't pass, but was reported incorrectly. See the link I posted above. Will look into why these errors aren't failing the build 😠

PR #87, opened 2018-10-15
Fixes routing / backbutton issue

georgebabey:


src/components/TableComponent/index.jsx
```diff
@@ -16,10 +19,26 @@ class TableComponent extends React.Component {
...
   componentDidUpdate(prevProps) {
-    const { id } = this.props;
+    const { id, location } = this.props;
+
+    // Handle the case where the table has changed by paginating the new table
     if (id && id !== prevProps.id) {
       this.props.paginateTable();
     }
+
+    // Handle the case where the router has changed query params in the store. This is used when
```
adamstankiewicz: nit: This comment is a little confusing. It says, `"...changed query params in the store"` but we're not actually looking at the query params in the store in the below logic.

georgebabey: Good catch. Carry over from when I was mapping router state to props.

src/components/TableComponent/index.jsx
```diff
@@ -16,10 +19,25 @@ class TableComponent extends React.Component {
...
   componentDidUpdate(prevProps) {
-    const { id } = this.props;
+    const { id, location } = this.props;
+
+    // Handle the case where the table has changed by paginating the new table
     if (id && id !== prevProps.id) {
       this.props.paginateTable();
     }
+
+    // Handle the case where the queryparams have changed. This is used whensorting & paging, but
```
adamstankiewicz: nit: `queryparams` to `query params`; `whensorting` to `when sorting`.

src/components/TableComponent/index.jsx
```diff
@@ -16,10 +19,25 @@ class TableComponent extends React.Component {
...
+    // also when the back button is used. We need to determine if this is a pagination or sorting
+    // request as we handle these as slightly differently actions in the action handlers.
```
adamstankiewicz: nit: `differently` to `different`

src/utils.js
```diff
@@ -27,23 +27,47 @@ const getAccessToken = () => {
...
 
-  if (newQuery !== window.location.search) {
-    history.push(`?${newQuery}`);
+  const newQueryString = `?${qs.stringify(newQuery)}`;
+  if (newQueryString !== window.location.search) {
+    history.push(newQueryString);
   }
 };
 
+const getPageOptionsFromQS = () => {
```
adamstankiewicz: nit: It might be worthwhile to add a comment above describing this function to make it clearer what its purpose is and that the `QS` in the function name refers to getting options from the URL query string.

src/data/actions/table.js
```diff
@@ -132,17 +99,13 @@ const sortData = (data, ordering) => {
...
-    // Our Api is expecting single orderField
-    const ordering = sortOptions.direction === 'desc' ? `-${sortOptions.key}` : sortOptions.key;
     const tableState = getState().table[tableId];
     const options = {
-      ...fetchOptions(tableState),
+      ...getPageOptionsFromQS(),
       ordering,
     };
-    updateUrl(options);
     dispatch(sortRequest(tableId, ordering));
```
adamstankiewicz: Unrelated to the changes in this PR, but do you know why we are passing `ordering` to the `sortRequest` action when the associated reducer ([`SORT_REQUEST`](https://github.com/edx/edx-portal/blob/master/src/data/reducers/table.js#L39)) does not set `ordering` in the Redux store?

georgebabey: At one point we were setting the ordering in the `SORT_REQUEST`, put for `PAGINATION` actions we not setting it until we successfully got the data in a `PAGATION_REQUEST`.  It seems when I made the change to the reducer I missed the action handler. Will update

georgebabey: We do this so that the events we publish to segment have the ordering in them.
screen shot 2018-10-16 at 10 17 03 pm

adamstankiewicz: Oh, right. Thanks for the reminder! 😄 I wonder if we should add comments when we pass values to the actions that aren't used in the reducer as a reminder for why we're doing that, e.g., `// pass ordering for segment tracking`.

adamstankiewicz:
@georgebabey Totally cool with that. It's a minor enough of an issue to not warrant blocking this PR. Should we add another JIRA story to update the Paragon `Table` to not block ENT-1283?

georgebabey:
@adamstankiewicz renamed the function and added comments about the extra data in the actions. I think this is good to go!

adamstankiewicz:
:shipit:

PR #106, opened 2018-11-02
Avatar icon for profile

muhammad-ammar:
[ENT-1104](https://openedx.atlassian.net/browse/ENT-1104)


muhammad-ammar:
@georgebabey @adamstankiewicz This is ready for review but test are failing due to snapshot. I am not sure why they are failing as I haven't touched the failing components?

src/components/Header/index.jsx
```diff
@@ -36,6 +38,7 @@ class Header extends React.Component {
           </div>
           {email && <Dropdown
             title={email}
+            iconElement={<img src={avatarIcon} alt="Avatar icon for profile" />}
```
adamstankiewicz: The image used here should be the profile image from the LMS. However, I don't believe we currently have a way to retrieve the authenticated user's profile image as it doesn't come back in the JWT token and I am unaware if there's currently an endpoint in LMS where we can fetch the user's profile image.

@georgebabey Thoughts on how we can/should get the profile image url?

muhammad-ammar: @adamstankiewicz @georgebabey We can use `api/user/v1/accounts/{username}` endpoint to get  profile images but the problem is that we don't have `username` available in edx-portal. There is another `api/user/v1/me` endpoint which can return username of logged in user but to me it doesn't look right to make a request just to get the logged in user's username.

Can not we get username inside the response returned by oauth2 authentication?

muhammad-ammar:
screen shot 2018-11-05 at 7 13 13 pm

@georgebabey @adamstankiewicz This is how profile image looks like on portal. 

Note: If you want to test this on devstack, profile image `upload` works on LMS but image access is not working due to incorrect mapping of `MEDIA_ROOT` [here](https://github.com/edx/edx-platform/blob/master/lms/envs/devstack.py#L10). So in order to make it work on devstack, you need change it to `MEDIA_ROOT = "/edx/var/edxapp/media"` from `MEDIA_ROOT = "/edx/var/edxapp/uploads"`

muhammad-ammar:
@adamstankiewicz @georgebabey Please review the changes. I have made the changes to get the profile image from LMS but username is hardcoded to `staff` for now.

src/components/Header/index.jsx
```diff
@@ -50,12 +60,14 @@ Header.propTypes = {
   enterpriseLogo: PropTypes.string,
   enterpriseName: PropTypes.string,
   email: PropTypes.string,
+  userProfile: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
adamstankiewicz: To get around needing this `eslint-disable-line` comment, you could do:
```javascript
{
  userProfile: PropTypes.shape({}),
}
```

src/components/LoginForm/index.jsx
```diff
@@ -26,6 +26,8 @@ class LoginForm extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (this.props.isAuthenticated && !prevProps.isAuthenticated) {
+      // TODO! add logic to get the username of currently loggedin user
+      this.props.fetchUserProfile('staff');
```
adamstankiewicz: Until the username is in the JWT token and we determine how we're going to fetch the profile image (putting it in the JWT token, or the API request), we should probably not call `fetchUserProfile` here.

@georgebabey Do you think we should ship the actions/reducers/constants associated with `fetchUserProfile` if we're always going to use the default avatar for the time being?

georgebabey: Knowing that we'll fast follow when the JWT token ships (and I confirmed with @douglashall the username will be there) I'd just leave this in for now. Maybe update the actions so it returns the default instead of calling the service?

muhammad-ammar:
@georgebabey Build is green now. There were two issues
* Some untouched components were complaining for snapshots
* And some snapshots were failing due to mismatch in dates

Fixed the snapshot issues with the following
* I updated the paragon version due to which those snapshots needs to be updated because `<Icon />` component is updated.
* I changed my macbook timezone to Eastern Standard Time and date mismatch issues were gone.

muhammad-ammar:
@georgebabey @adamstankiewicz I have addressed the feedback. Please continue the review.

src/components/LoginForm/index.jsx
adamstankiewicz: @muhammad-ammar @georgebabey Now that the @douglashall's auth change has merged, the username should exist in the `authentication` Redux state.

@muhammad-ammar If you rebase on `master`, are you able to access the username in the `authentication` state? However, with the authentication changes, we no longer have a `LoginPage` component/container so this call would need to moved elsewhere.

PR #109, opened 2018-11-07
Coupon management and code details

adamstankiewicz:
* Code management page with coupons overview
* Coupon details with the codes associated with each coupon
* "Request More Codes" button and page structure
    * When the form on this new page is submitted, it redirects back to the "Code Management" page, displaying a success `StatusAlert` informing of the success.
* Demo data for coupons and coupon details

Still left to do:
* Pagination
    * coupon overview & coupon details
    * demo data
* Filtering for coupon details
    * demo data
* Tests

### Screenshots

**Code Management**
image

**Request More Codes**
image

image


georgebabey:
We should consider bundling coupon management separately as there are some customers that will not use this feature at all

src/components/CodeManagement/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+      const state = { ...history.location.state };
+      delete state.hasRequestedCodes;
+      history.replace({ ...history.location, state });
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { enterpriseId } = this.props;
+    if (enterpriseId && enterpriseId !== prevProps.enterpriseId) {
+      this.props.fetchCouponOrders();
```
georgebabey: do you need to `fetchPortalConfiguration` here as well or is that covered up the stack?

src/components/CodeManagement/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+        title="Unable to load coupons"
+        message={`Try refreshing your screen (${this.props.error.message})`}
+      />
+    );
+  }
+
+  renderCoupons() {
+    const { coupons } = this.props;
+    return (
+      <React.Fragment>
```
georgebabey: 👍

src/components/Coupon/Coupon.scss
```diff
@@ -0,0 +1,86 @@
...
+        &:hover,
+        &:focus {
+            opacity: 1;
+        }
+    }
+
+    &:not(.expanded) {
+        &:hover,
+        &:focus {
+            border-color: $gray-400 !important; // `!important` needed to override .border Bootstrap class
```
georgebabey: 😢

src/components/Coupon/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+      this.toggleCouponDetails();
+    }
+  }
+
+  renderExpandCollapseIcon() {
+    const { detailsExpanded } = this.state;
+    const iconClass = detailsExpanded ? 'fa-chevron-up' : 'fa-chevron-down';
+    const screenReaderText = detailsExpanded ? 'Close' : 'Open';
+    return (
+      <Button
```
georgebabey: Why is this a button if it has no action and is not accessible (`tabIndex="-1"`)

adamstankiewicz:
> We should consider bundling coupon management separately as there are some customers that will not use this feature at all

Makes sense. I would suggest we figure out how to do the bundling of coupon management in a separate PR. Thoughts?

src/components/CodeManagement/index.jsx
adamstankiewicz: We should only need it in the initial `componentDidMount` as the `enterpriseSlug` will change up the stack in the prop that gets passed to the route for code management.

For the `Admin` page, we only call it in `componentDidMount` as well.

georgebabey:
@adamstankiewicz lets tackle the `componentDidMount` and bundling as separate PRs

src/components/CodeManagement/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+    super(props);
+
+    this.couponRefs = [];
+
+    this.handleCouponExpand = this.handleCouponExpand.bind(this);
+  }
+
+  componentDidMount() {
+    const { enterpriseId, enterpriseSlug } = this.props;
+    const history = createHistory();
```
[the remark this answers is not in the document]
adamstankiewicz: I'm still trying to figure out why, but importing from `data/history.js` does not actually work the same as creating new history here.

src/components/Coupon/index.jsx
adamstankiewicz: Yeah, I will be making this just an icon without the button. Same thing with the error icon which is still to come.

src/components/CodeManagement/index.jsx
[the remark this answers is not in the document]
georgebabey: Hmm - it did work well in `utils.js` so I'm unclear why it would not here.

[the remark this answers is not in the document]
adamstankiewicz: Need to do a little more testing but in the `<Redirect />` in `RequestMoreCodes` component, I was including the `push` prop. Removing that prop seems to make the `history.location.state` not persist, solving the problem.

src/components/CodeManagement/index.jsx
```diff
@@ -0,0 +1,240 @@
...
+  componentDidUpdate(prevProps) {
+    const { enterpriseId, location, history } = this.props;
+
+    if (enterpriseId && enterpriseId !== prevProps.enterpriseId) {
+      this.props.fetchCouponOrders();
+    }
+
+    if (location.state && location.state.hasRequestedCodes) {
+      this.setState({ // eslint-disable-line react/no-did-update-set-state
+        hasRequestedCodes:
```
georgebabey: nit: newline here makes this hard to read

src/components/CodeManagement/index.jsx
```diff
@@ -0,0 +1,240 @@
...
+    );
+  }
+
+  renderRequestCodesSuccessMessage() {
+    return (
+      <StatusAlert
+        className={['mt-3']}
+        alertType="success"
+        iconClassNames={['fa', 'fa-check-circle']}
+        title="Request for more codes received."
```
georgebabey: Nit: titles shouldn't have a `.` ?  We don't include it elsewhere.

src/components/CodeManagement/index.jsx
adamstankiewicz: Already fixed :)

PR #112, opened 2018-11-09
Add buttonType prop to NumberCard details button

adamstankiewicz:
With the refactor of the `NumberCard` expand/collapse button, we're not setting the `buttonType` prop which causes the background to appear gray on Safari and Firefox.

This PR adds `buttonType="link"` to the button which resolves this issue. It also adds a hover/focus state of underline the text in the button which is better for a11y.

### DIFF
--- a/src/components/Admin/__snapshots__/Admin.test.jsx.snap
+++ b/src/components/Admin/__snapshots__/Admin.test.jsx.snap
@@ -227,7 +227,7 @@ Array [
                   className="footer-title"
                 >
                   <button
-                    className="btn toggle-collapse btn-block"
+                    className="btn toggle-collapse btn-block btn-link"
                     onBlur={[Function]}
                     onClick={[Function]}
                     onKeyDown={[Function]}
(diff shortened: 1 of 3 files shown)


georgebabey:
Should we also make buttonType a required prop in paragon? You don't need to do so right now but maybe file a ticket.

adamstankiewicz:
> Should we also make buttonType a required prop in paragon? You don't need to do so right now but maybe file a ticket.

Might not be a bad idea. I don't think having only a `.btn` class is technically valid Bootstrap. All the Bootstrap docs include a type as well. I'll file a ticket in Paragon.

Related, I just realized everywhere we have an `btn-outline-primary` class on a `<Button />` should actually be the `buttonType` prop instead of adding it to `className` if we want to use the `<Button />` API properly. I will open another PR to clean that up (should just be on the csv download button & "reset to full report" button).

Example:

```jsx
<Button buttonType="outline-primary" label="Hello world!" />
```
instead of

```jsx 
<Button className={['btn-outline-primary']} label="Hello world!" />
```

PR #114, opened 2018-11-12
added request codes action-reducer-service

irfanuddinahmad:
Description: This PR adds the necessary action, reducer and LMSApiService updates for the "request codes" functionality

JIRA: https://openedx.atlassian.net/browse/ENT-1169


adamstankiewicz:
@irfanuddinahmad The base branch of this PR (`astankiewicz/orders-overview`) is now merged to `master`. You should be able to update the base branch to `master` now so the only changes in this PR are yours.

src/containers/RequestCodesPage/index.jsx
```diff
@@ -1,18 +1,26 @@
 import { connect } from 'react-redux';
 
 import RequestCodesPage from '../../components/RequestCodesPage';
-
```
adamstankiewicz: nit: We should keep this blank line to separate the component imports from action imports. Makes it a little more clear what things we're importing where.

src/data/services/LmsApiService.js
```diff
@@ -30,6 +30,17 @@ class LmsApiService {
     return apiClient.get(enterpriseListUrl);
   }
 
+  static requestCodes(enterpriseName, emailAddress, numberOfCodes) {
```
adamstankiewicz: For functions that have more than one argument like this, it's usually a little cleaner to pass in an object with keys for the three variables, e.g.:

```javascript
static requestCodes({ enterpriseName, emailAddress, numberOfCodes }) {
  ...
}
```

That way, there's no confusion around the order of arguments passed into the function and it becomes more explicit.

src/containers/RequestCodesPage/index.jsx
```diff
@@ -1,18 +1,26 @@
...
   emailAddress: state.authentication.email,
+  error: state.codes.error,
+  success: state.codes.success,
 });
 
 const mapDispatchToProps = dispatch => ({
   fetchPortalConfiguration: (slug) => {
     dispatch(fetchPortalConfiguration(slug));
   },
+  requestCodes: (enterpriseName, emailAddress, numberOfCodes) => {
```
adamstankiewicz: As described in another comment, we should pass in a single object here instead of separate arguments. The `requestCodes` action handler can then use object destructuring to extract the 3 values as shown in the example in the other comment.

```javascript
requestCodes: (options) => {
  dispatch(requestCodes(options));
}
```

src/data/actions/codes.js
```diff
@@ -0,0 +1,53 @@
...
+const requestCodesFailure = error => ({
+  type: REQUEST_CODES_FAILURE,
+  payload: {
+    error,
+  },
+});
+
+const clearRequestCodesEvent = () => ({ type: CLEAR_REQUEST_CODES });
+
+const requestCodes = (enterpriseName, emailAddress, numberOfCodes) => (
```
adamstankiewicz: Related to the other comments, this function could accept just a single object:

```javascript
const requestCodes = (options) => (
  (dispatch) => {
    dispatch(requestCodesRequest());
    return LmsApiService.requestCodes(options)
      .then(() => { ... })
      .catch(() => {... });
  }
);
```
The `LmsApiService.requestCodes()` function would extract the values out using object destructuring.

PR #126, opened 2018-11-20
Update coupon demo data api response to reflect actual api fields

adamstankiewicz:
Now that we have know the format of the coupon overview API response, we can update the field names in the `Coupon` component and demo data.

Also, after discussing with @brittneyexline and Jennifer about how the `max_uses` field can potentially be `null` (i.e., "unlimited" uses), the UI should only show "X of Y" in the "Enrollments Redeemed" column of the coupons if `max_uses` is set. If it isn't, we will only show `num_uses` field, e.g.:

image

### DIFF
--- a/src/components/Coupon/Coupon.test.jsx
+++ b/src/components/Coupon/Coupon.test.jsx
@@ -25,8 +25,8 @@ const store = mockStore({
 const initialCouponData = {
   id: 1,
   title: 'Test Coupon',
-  start_date: '',
-  end_date: '',
+  start_date: '2018-06-01T12:00:00Z',
+  end_date: '2019-02-01T12:00:00Z',
   has_error: false,
   num_unassigned: 0,
   num_uses: 0,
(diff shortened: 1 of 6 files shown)


adamstankiewicz:
@georgebabey I pushed up a few tests for the `Coupon` component, including snapshots for both variations of `max_uses` (i.e., `null` or otherwise).

adamstankiewicz:
@georgebabey Pushed another commit to fix dates:
- `Coupon` was only rendering whatever date was given to it. It looked formatted before because the demo data formatted it correctly. Now, the formatting takes place inside the component.
- Dates in the demo data that are simulated API responses were not in the same format as the API actually returns; it should be returned as an ISO string. Now, every date in the demo data is returned as an ISO string, which mimics our API responses.
  - Had to refactor a bit because we filter using date comparisons in the demo data (e.g., filter for past week completions), and we can do those comparisons with ISO strings. Instead, we parse the date strings using `moment` during the filtering.

Let me know if you have questions!

PR #130, opened 2018-12-07
Swap out filter buttons with dropdown select

adamstankiewicz:
This PR removes the previous filter buttons and, in their place, puts in an `InputSelect` component. I also rename a few variables/props to be `isExpanded`. This PR should be probably be merged before @mushtaqak's WIP PR lands as the filter UI has changed here.

image

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -17,31 +17,28 @@ class CouponDetails extends React.Component {
   constructor(props) {
     super(props);
 
-    this.tableColumns = [
-      {
-        label: <CheckBox />,
-        key: 'select',
-      },
-      {
-        label: 'Assigned To',
-        key: 'assigned_to',
-      },
-      {
-        label: 'Redemptions',
-        key: 'redemptions',
-      },
-      {
-        label: 'Code',
-        key: 'title',
-      },
-      {
-        label: 'Actions',
-        key: 'actions',
-      },
-    ];
+    this.defaultToggle = 'not-assigned';
 
     this.state = {
-      selectedToggle: 'not-assigned',
+      selectedToggle: this.defaultToggle,
+      tableColumns: [
+        {
+          label: <CheckBox />,
+          key: 'select',
+        },
+        {
+          label: 'Redemptions',
+          key: 'redemptions',
+        },
+        {
+          label: 'Code',
+          key: 'title',
+        },
+        {
+          label: 'Actions',
+          key: 'actions',
+        },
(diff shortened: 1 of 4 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -41,16 +41,13 @@ class CouponDetails extends React.Component {
     ];
 
     this.state = {
-      selectedFilter: null,
+      selectedToggle: 'not-assigned',
```
mushtaqak: [ENT-1345](https://openedx.atlassian.net/browse/ENT-1345) mentions to default filter (toggle now) to fetch all coupon codes records. Here we will be defaulting to 'not-assigned'?

src/components/CouponDetails/index.jsx
```diff
@@ -151,37 +158,28 @@ class CouponDetails extends React.Component {
...
-                        buttonType={selectedFilter === 'not-redeemed' ? 'primary' : 'outline-primary'}
-                        label="Not Redeemed"
-                        onClick={() => this.setSelectedFilter('not-redeemed')}
+                        className={['ml-2']}
+                        buttonType="primary"
+                        label="Go"
+                        onClick={this.handleToggleSelect}
                         disabled={this.isTableLoading()}
                       />
-                      {selectedFilter &&
```
mushtaqak: @adamstankiewicz about my question above, If we are going to keep 'default to all records', we may need this. If not, you can ignore this.

src/components/CouponDetails/index.jsx
adamstankiewicz: In recent discussions with UX and Product, we decided to treat this as toggles instead of filters, where the initial table data will be unassigned codes. I will reconfirm this with UX/Product today, though.

adamstankiewicz:
@georgebabey @mushtaqak I pushed up a commit to this PR where we add the "Assigned To" table column only when the selected toggle is something other than "not-assigned". When the selected toggle is "not-assigned", we ensure the "Assigned To" table column is removed.

adamstankiewicz:
Also, just FYI, still not quite sure why the snapshots are failing when everything passes locally.

adamstankiewicz:
Got both checks to pass; just had to rebase this branch on `master`.

src/components/CouponDetails/index.jsx
adamstankiewicz: @mushtaqak Chatted with UX about this again to confirm that initial table data will be only unassigned codes.

PR #141, opened 2018-12-14
Adding in actions/reducers/constants as a first pass

christopappas:


src/data/actions/individualAssignment.js
```diff
@@ -0,0 +1,42 @@
+import {
+  INDIVIDUAL_ASSIGNMENT_REQUEST,
+  INDIVIDUAL_ASSIGNMENT_SUCCESS,
+  INDIVIDUAL_ASSIGNMENT_FAILURE,
+} from '../constants/individualAssignment';
+
+
```
adamstankiewicz: nit: Extra line

src/data/constants/individualAssignment.js
```diff
@@ -0,0 +1,10 @@
+const INDIVIDUAL_ASSIGNMENT_REQUEST = 'COUPONS_REQUEST';
+const INDIVIDUAL_ASSIGNMENT_SUCCESS = 'COUPONS_SUCCESS';
+const INDIVIDUAL_ASSIGNMENT_FAILURE = 'COUPONS_FAILURE';
+
+
```
adamstankiewicz: nit: Extra line

src/data/actions/individualAssignment.js
```diff
@@ -0,0 +1,42 @@
...
+  payload: {
+    error,
+  },
+});
+
+
+const fetchIndividualAssignment = options => (
+  (dispatch) => {
+    dispatch(fetchIndividualAssignmentRequest(options));
+    return EcommerceApiService.fetchIndividualAssignment(options)
```
adamstankiewicz: Right now, the `EcommerceApiService` has some placeholder functions for the other 2 API calls currently being used for the code management page. You could create an additional placeholder in there for this function.

As is, once the `CodeAssignmentModal` is wired up to these actions/reducers, this would likely throw a JS error because you're trying to invoke a function that doesn't exist yet.

Related to the overall PR comment, this function should also be named generic enough to be applicable for both individual and bulk assign.

src/data/actions/individualAssignment.js
```diff
@@ -0,0 +1,42 @@
...
+  type: INDIVIDUAL_ASSIGNMENT_FAILURE,
+  payload: {
+    error,
+  },
+});
+
+
+const fetchIndividualAssignment = options => (
+  (dispatch) => {
+    dispatch(fetchIndividualAssignmentRequest(options));
```
adamstankiewicz: It's not necessary to pass the `options` into this action as it doesn't use any of the form data (which is what the `options` variable will contain).

src/data/reducers/individualAssignment.js
```diff
@@ -0,0 +1,36 @@
...
+      return {
+        loading: false,
+        error: action.payload.error,
+      };
+    default:
+      return state;
+  }
+};
+
+export default individualAssignmentReducer;
```
adamstankiewicz: You'll also need to import this reducer into the root reducer in `src/data/reducers/index.js`.

src/data/actions/individualAssignment.js
```diff
@@ -0,0 +1,42 @@
...
+});
+
+const fetchIndividualAssignmentFailure = error => ({
+  type: INDIVIDUAL_ASSIGNMENT_FAILURE,
+  payload: {
+    error,
+  },
+});
+
+
```
adamstankiewicz: nit: Extra line

src/data/reducers/index.js
```diff
@@ -9,6 +9,7 @@ import csv from './csv';
 import userProfile from './userProfile';
 import coupons from './coupons';
 import sidebar from './sidebar';
+import codeAssignmentReducer from './codeAssignment';
```
adamstankiewicz: This should be named just `codeAssignment`. It uses this name as a key in the Redux store object.

src/data/constants/codeAssignment.js
```diff
@@ -0,0 +1,10 @@
+const CODE_ASSIGNMENT_REQUEST = 'COUPONS_REQUEST';
+const CODE_ASSIGNMENT_SUCCESS = 'COUPONS_SUCCESS';
+const CODE_ASSIGNMENT_FAILURE = 'COUPONS_FAILURE';
+
+
```
adamstankiewicz: nit: Extra line

PR #144, opened 2018-12-19
ENT-1173 Gate access to code management based on enterprise configuration

brittneyexline:


src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -65,12 +68,34 @@ describe('<Sidebar />', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('renders correctly when code management is hidden', () => {
+    const store = mockStore({
+      sidebar: {
+        ...initialState.sidebar,
+      },
+      portalConfiguration: {
+        ...initialState.portalConfiguration,
```
adamstankiewicz: Because `enableCodeManagementScreen` is the only key within `portalConfiguration` in the store, we probably don't need to spread on the `initialState.portalConfiguration` here (i.e., could remove this line).

src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -65,12 +68,34 @@ describe('<Sidebar />', () => {
...
+    expect(tree).toMatchSnapshot();
+  });
+
   it('renders correctly when expanded', () => {
     const store = mockStore({
       sidebar: {
         ...initialState.sidebar,
         isExpanded: true,
       },
+      portalConfiguration: {
```
adamstankiewicz: I'm wondering if Instead of including `portalConfiguration: { ...initialState.portalConfiguration }` in the subsequent tests, we could just do something like this instead:

```jsx
const store = mockStore({
  ...initialState,
  sidebar: {
    ...initialState.sidebar,
    isExpanded: true,
  },
});
```

This would make setting up the stores in the following tests a little more concise.

PR #149, opened 2019-01-08
Individual & Bulk Code Assignment

adamstankiewicz:
This PR introduces the ability to individual and bulk assign codes. When selecting the bulk action, the bulk version of the `CodeAssignmentModal` is shown. When selecting an individual assign button, the individual version of the `CodeAssignmentModal` is shown.

A form is in both modals, and validates on submit with both a global error status alert and inline validation messages.

The submit handler formats the data needed to pass to the assignment API endpoint, however does not wire up the frontend with the backend API call.

I've also included basic selection logic for selecting codes that will be used for bulk assign (keeps track of the selected codes in the `CouponDetails` component. Additionally, there is now the ability to change the selected toggle, which will cause a refetch of the table data.

The following is not included in this PR and will be introduced in subsequent PRs:
* Tests, tests, tests (these have been started, but need to be finished up). Due to the size of the PR and that these features are behind a flag, I suggest we can review this code and its tests separately.
* Logic around selecting all codes.


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -0,0 +1,338 @@
...
+      error,
+    } = this.props;
+
+    const errorMessageRef = this.errorMessageRef && this.errorMessageRef.current;
+
+    if (submitSucceeded && submitSucceeded !== prevProps.submitSucceeded) {
+      onClose();
+    }
+
+    if (submitFailed && error && error.length > 0 && error !== prevProps.error && errorMessageRef) {
```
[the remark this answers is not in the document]
adamstankiewicz: Hmm, good point. The `error` array, set during form validation, will always be a non-empty array whenever `submitFailed` is `true`. 

I suppose we could even get rid of both the `error && error.length > 0` conditions!

src/components/CodeAssignmentModal/index.jsx
```diff
@@ -0,0 +1,338 @@
...
+      // When there is an error, focus on the error message status alert
+      errorMessageRef.focus();
+    }
+  }
+
+  validateBulkAssign(formData) {
+    const { data: { unassignedCodes } } = this.props;
```
brittneyexline: i'm not quite able to tell where unassignedCodes comes from, but does it mean codes or code usages? there could be just one unassigned code, but it could have enough usages left for assignment, and i can't tell if the logic below would reflect that.

src/components/TextAreaAutoSize/index.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import classNames from 'classnames';
+import TextArea from 'react-textarea-autosize';
+import { ValidationMessage, Variant } from '@edx/paragon';
+
+import './TextAreaAutoSize.scss';
+
+const TextAreaAutoSize = ({
```
brittneyexline: is there a reason why this wouldn't be in paragon? it doesn't seem specific to code assignment

src/components/CodeAssignmentModal/index.jsx
adamstankiewicz: `unassignedCodes` is passed to the `CodeAssignmentModal` as a [key within the `data` object prop](https://github.com/edx/edx-portal/pull/149/files#diff-6a1729462bebbd0c4ca0ec8b6fe2d0daR204).

The actual source of the `unassignedCodes` value is from the `num_unassigned` field from the coupon overview API endpoint ([see example response here](https://ecommerce-business.sandbox.edx.org/api/v2/enterprise/coupons/1b542b06d35345e698679579d9e12745/overview/))., e.g.:

```json
"results": [
    {
      "end_date": "2019-01-26T00:00:00Z",
      "has_error": false,
      "id": 418,
      "max_uses": null,
      "num_codes": 1,
      "num_unassigned": 0,
      "num_uses": 0,
      "start_date": "2019-01-05T00:00:00Z",
      "title": "123456",
      "usage_limitation": "Single use"
    },
],
```

.eslintrc
```diff
@@ -15,6 +15,12 @@
     "jsx-a11y/anchor-is-valid": [ "error", {
       "components": [ "Link" ],
       "specialLink": [ "to" ]
+    }],
+    "jsx-a11y/label-has-for": [ "error", {
```
georgebabey: Should this go in https://github.com/edx/eslint-config-edx?

adamstankiewicz:
@brittneyexline Yeah, I am working in a separate branch for the tests and the select all logic because of the size of the PR and potential for merge conflicts.

src/index.scss
```diff
@@ -8,7 +8,25 @@ $fa-font-path: "~font-awesome/fonts";
...
+            margin-left: 4px;
+        }
+    }
+
+    textarea {
+        resize: none;
+    }
+}
+
+.font-weight-600 {
```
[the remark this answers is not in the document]
adamstankiewicz: I'm removing this class, and instead just putting `font-weight: 600;` on the `p` element which the SCSS is already targeting.

PR #151, opened 2019-01-10
Demo data for code management

mushtaqak:
This PR adds demo data for code management screen.

[ENT-1335](https://openedx.atlassian.net/browse/ENT-1335)

### DIFF
--- a/src/components/CodeManagement/index.jsx
+++ b/src/components/CodeManagement/index.jsx
@@ -9,6 +9,7 @@ import Hero from '../Hero';
 import Coupon from '../Coupon';
 import LoadingMessage from '../LoadingMessage';
 import StatusAlert from '../StatusAlert';
+import { updateUrl } from '../../utils';
 
 class CodeManagement extends React.Component {
   constructor(props) {
@@ -17,6 +18,7 @@ class CodeManagement extends React.Component {
     this.couponRefs = [];
     this.state = {
       hasRequestedCodes: false,
+      currentPage: 1,
     };
 
     this.handleCouponExpand = this.handleCouponExpand.bind(this);
@@ -54,8 +56,27 @@ class CodeManagement extends React.Component {
     this.props.clearCouponOrders();
   }
 
+  paginateCouponOrders(pageNumber) {
+    const page = parseInt(pageNumber, 10);
+    if (page !== this.state.currentPage) {
+      this.props.fetchCouponOrders({ page });
+      this.setState({
+        currentPage: page,
+      });
+      updateUrl({ page });
+    }
+  }
+
   handleCouponExpand(selectedIndex) {
+    const { location, history } = this.props;
     const coupons = this.couponRefs;
+
+    // reset location state for paginated coupon orders.
+    history.replace({
+      ...location.pathname,
+      state: {},
+    });
+
     if (coupons.length) {
       coupons.forEach((ref, index) => {
         // close all coupons but the coupon that was selected
@@ -115,8 +136,9 @@ class CodeManagement extends React.Component {
         ))}
         <div className="d-flex mt-4 justify-content-center">
           <Pagination
-            onPageSelect={() => {}} // TODO replace with actual callback
+            onPageSelect={page => this.paginateCouponOrders(page)}
             pageCount={coupons.num_pages}
+            currentPage={coupons.current_page}
             paginationLabel="coupons pagination"
           />
         </div>
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -459,7 +459,7 @@ class CouponDetails extends React.Component {
                 key={`table-${selectedToggle}-${refreshIndex}`}
                 id="coupon-details"
                 className="coupon-details-table"
-                fetchMethod={() => EcommerceApiService.fetchCouponDetails(id)}
+                fetchMethod={options => EcommerceApiService.fetchCouponDetails(id, options)}
                 columns={tableColumns}
                 formatData={this.formatCouponData}
               />
--- a/src/demo/coupons.js
+++ b/src/demo/coupons.js
@@ -12,25 +12,55 @@ if (configuration.NODE_ENV === 'development') {
   firstCouponHasError = true;
 }
 
-const coupons = {
-  count: couponsCount,
-  num_pages: Math.ceil(couponsCount / 10),
-  current_page: 1,
-  results: [...Array(couponsCount)].map((_, index) => {
-    const validFromDate = faker.date.past();
-    const totalEnrollments = faker.random.number({ min: 1, max: 50 });
-    return {
-      id: index,
-      title: faker.random.words(),
-      start_date: validFromDate.toISOString(),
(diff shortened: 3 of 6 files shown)


config/webpack.dev.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = Merge.smart(commonConfig, {
     path.resolve(__dirname, '../src/index.jsx'),
 
     // Uncomment this entrypoint to return static demo data instead of calling the data-api
-    // path.resolve(__dirname, '../src/demo/index.js'),
+    path.resolve(__dirname, '../src/demo/index.js'),
```
mushtaqak: Note t reviewers: I use vagrant so had to change these files often - will remove this file before merging.

mushtaqak:
@adamstankiewicz Can you review this?

There is one pagination bug + ambiguity for inner pagination (coupon codes detail table), it's not getting correctly paginated. I discussed with @georgebabey - we can handle it in a separate story. 

Also, for that to work (I tried deep-linking codes in https://github.com/edx/edx-portal/pull/152 - so that ambiguitity between two tables is lessened and user knows which table is being paginated but there are more issues that need to handled, may be, you can have a pre-look at it)

config/webpack.dev.config.js
adamstankiewicz: In the future, can you `git stash` these types of changes instead of including them in the PR? Having these changes in the PR is also causing one of the tests to fail, specifically overriding the `enableCodeManagementScreen` configuration.

src/demo/rewireEcommerceApiService.js
```diff
@@ -1,21 +1,40 @@
...
 
 const findCouponIndexById = couponId =>
-  coupons.results.findIndex(coupon => coupon.id === couponId);
+  coupons.findIndex(coupon => coupon.id === couponId);
 
 const rewire = () => {
-  EcommerceApiService.fetchCouponOrders = () => Promise.resolve({
-    data: coupons,
-  });
+  const fetchData = (options = { page: 1 }, couponData) => {
```
adamstankiewicz: Can we move the `fetchData` function outside of the `rewire` function? We should try to keep the `rewire` function to contain only the actual rewirings of the service functions, and not any helper functions.

src/components/CodeManagement/index.jsx
```diff
@@ -65,6 +80,11 @@ class CodeManagement extends React.Component {
         }
       });
     }
+    // reset location state for paginated coupon orders.
+    history.replace({
```
adamstankiewicz: I think I understand why this is here, but do you mind clarifying in more detail? Does this make sense to do on expand or on collapse of coupons?

src/demo/coupons.js
```diff
@@ -12,25 +12,20 @@ if (configuration.NODE_ENV === 'development') {
...
-      has_error: false,
-    };
-  }),
-};
+const coupons = [...Array(couponsCount)].map((_, index) => {
+  const validFromDate = faker.date.past();
+  const totalEnrollments = faker.random.number({ min: 1, max: 50 });
+  return {
+    id: index,
+    title: faker.lorem.words(faker.random.number({ min: 1, max: 3 })).toUpperCase(),
```
adamstankiewicz: Instead of generating random words from faker, which still may not be realistic, can we create an array of realistic coupon names and the `title` pulling in a random one for the title?

It seems like some of the coupon names we have are titled like `10OFF`, or `100OFF`. Using titles that reflect actual names may be a good idea here. Thoughts?

src/demo/coupons.js
```diff
@@ -49,34 +44,32 @@ const getAssignedTo = (isAssigned = false) => {
...
+          max: codeHasError ? redemptionsAvailablePerCode - 1 : redemptionsAvailablePerCode,
+        }) : 0,
+      },
+      error: codeHasError ? `Unable to deliver email to ${assignedTo.name} (${assignedTo.email})` : null,
+    };
+  });
 };
 
-coupons.results[0].has_error = firstCouponHasError;
+const getCodesCsv = () => 'assigned_to,code,redeem_url,redemptions.available,redemptions.used\r\n,ZBXF24QK6TJQMUSU,https://testserver.fake/coupons/offer/?code=ZBXF24QK6TJQMUSU,3,3\r\n,ZBXF24QK6TJQMUSU,https://testserver.fake/coupons/offer/?code=ZBXF24QK6TJQMUSU,3,3\r\n,ZBXF24QK6TJQMUSU,https://testserver.fake/coupons/offer/?code=ZBXF24QK6TJQMUSU,3,3\r\n,VF7TV7HMMQF63GS2,https://testserver.fake/coupons/offer/?code=VF7TV7HMMQF63GS2,3,3\r\n,VF7TV7HMMQF63GS2,https://testserver.fake/coupons/offer/?code=VF7TV7HMMQF63GS2,3,3\r\n,VF7TV7HMMQF63GS2,https://testserver.fake/coupons/offer/?code=VF7TV7HMMQF63GS2,3,3\r\n';
```
adamstankiewicz: Is there a way we can generate the CSV data string programmatically using the `getCodes` function instead of hardcoding it?

src/components/CodeManagement/index.jsx
mushtaqak: This is when we collapse a coupon details.

Idea is when a coupon is expanded, the pagination or any other query param in url should be removed ... (this is done to minimise ambiguity between two tables) ... but I am open to ideas.

src/components/CouponDetails/index.jsx
```diff
@@ -32,7 +32,7 @@ class CouponDetails extends React.Component {
         },
         {
           label: 'Code',
-          key: 'title',
+          key: 'code',
```
mushtaqak: This is a fix. API service actually returns `code` column for code details endpoint.

mushtaqak: **edit**: This will be covered in https://github.com/edx/edx-portal/pull/153 along with further fixes needed.

mushtaqak:
@adamstankiewicz Addressed all the concerns.

mushtaqak:
Since filters are not yet implemented/completed, I guess we can merge this one and handle them in a separate PR, thoughts?

adamstankiewicz:
@mushtaqak From the frontend side of things, the toggles are mostly implemented. The only thing that's not there right now is passing the currently selected toggle/filter to the `EcommerceServiceApi` function(s).

I don't think it'd be too much extra work in this PR to get that set up and working with the demo data. But, if you'd rather do it in a separate PR, I'm also OK with that. Just let me know what you prefer, and if you do decide to go the separate PR route, I'll approve this PR so you can merge it.

adamstankiewicz:
@mushtaqak For sending the currently selected filter to the API, you could refer to #129 (specifically, https://github.com/edx/edx-portal/pull/129/files#diff-7cb662368037c7e1e05fb5a0c95bdfe5R61).

mushtaqak:
Frontend work for filters + demo data for filters will be handled in [sub task](https://openedx.atlassian.net/browse/ENT-1471)

PR #154, opened 2019-01-17
send handled/unhandled errors to newrelic

muhammad-ammar:
[ENT-1290](https://openedx.atlassian.net/browse/ENT-1290)

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "clean-webpack-plugin": "^0.1.19",
     "codecov": "^3.0.0",
     "css-loader": "^0.28.9",
-    "enzyme": "^3.3.0",
+    "enzyme": "^3.6.0",
     "enzyme-adapter-react-16": "^1.6.0",
     "es-check": "^5.0.0",
     "eslint-config-edx": "^4.0.3",
--- a/src/components/ErrorBoundary/ErrorBoundary.test.jsx
+++ b/src/components/ErrorBoundary/ErrorBoundary.test.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { shallow } from 'enzyme';
+
+import NewRelicService from '../../data/services/NewRelicService';
+import ErrorBoundary from './index';
+
+global.newrelic = {
+  addPageAction: jest.fn(),
+  noticeError: jest.fn(),
+};
+
+function Welcome() {
+  return <h1>Hi!</h1>;
+}
+
+describe('<ErrorBoundary />', () => {
+  it('renders childern correctly when no error', () => {
+    const tree = renderer
+      .create((
+        <ErrorBoundary>
+          <Welcome name="MA" />
+        </ErrorBoundary>
+      ))
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+  it('logs the error', () => {
+    const welcomeError = new Error('Catch me if you can!');
+    NewRelicService.logError = jest.fn();
+
+    const wrapper = shallow((
+      <ErrorBoundary>
+        <Welcome />
+      </ErrorBoundary>
+    ));
+    wrapper.find(Welcome).simulateError(welcomeError);
+    expect(NewRelicService.logError).toHaveBeenCalledWith(new Error('Catch me if you can!'));
+  });
+});
--- a/src/components/ErrorBoundary/index.jsx
+++ b/src/components/ErrorBoundary/index.jsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import NewRelicService from '../../data/services/NewRelicService';
+
+class ErrorBoundary extends React.Component {
+  componentDidCatch(error) {
+    NewRelicService.logError(error);
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+export const withErrorBoundary = WrappedComponent => props => (
+  <ErrorBoundary>
+    <WrappedComponent {...props} />
+  </ErrorBoundary>
+);
+
+ErrorBoundary.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default ErrorBoundary;
--- a/src/components/RequestCodesPage/index.jsx
+++ b/src/components/RequestCodesPage/index.jsx
@@ -9,6 +9,8 @@ import LoadingMessage from '../LoadingMessage';
 
 import LmsApiService from '../../data/services/LmsApiService';
 
+import NewRelicService from '../../data/services/NewRelicService';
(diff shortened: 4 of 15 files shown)


config/webpack.dev.config.js
```diff
@@ -4,6 +4,7 @@ const Merge = require('webpack-merge');
 const path = require('path');
 const webpack = require('webpack');
```
muhammad-ammar: Note to Reviewers: Changes in this file are for local testing only. Will be removed before merge.

package.json
```diff
@@ -65,7 +65,7 @@
     "clean-webpack-plugin": "^0.1.19",
     "codecov": "^3.0.0",
     "css-loader": "^0.28.9",
-    "enzyme": "^3.3.0",
+    "enzyme": "^3.6.0",
```
muhammad-ammar: Note to Reviewers: This is needed to test the ErrorBoundary. Please see https://github.com/airbnb/enzyme/issues/1255

src/Logging/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+function logError(error) {
+  if (typeof newrelic !== 'undefined') {
+    newrelic.noticeError(error); // eslint-disable-line no-undef
+  }
+}
+
+function logAPIErrorResponse(error) {
+  let { message } = error;
+  if (error.response) {
```
adamstankiewicz: I'm curious where you're getting these variations of errors. Do you have a link to some documentation? It seems mostly inline with axios's error handling documentation: https://github.com/axios/axios#handling-errors.

src/data/actions/dashboardAnalytics.js
```diff
@@ -5,6 +5,7 @@ import {
   CLEAR_DASHBOARD_ANALYTICS,
 } from '../constants/dashboardAnalytics';
 import EnterpriseDataApiService from '../services/EnterpriseDataApiService';
+import * as logging from '../../Logging';
```
adamstankiewicz: I would recommend only importing the functions we need in this file, instead of using the wildcard asterisk.

src/data/actions/enterpriseList.js
```diff
@@ -5,6 +5,7 @@ import {
   PAGINATION_SUCCESS,
   PAGINATION_FAILURE,
 } from '../constants/table';
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/data/actions/portalConfiguration.js
```diff
@@ -5,6 +5,7 @@ import {
   CLEAR_PORTAL_CONFIGURATION,
 } from '../constants/portalConfiguration';
 import LmsApiService from '../services/LmsApiService';
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/data/actions/table.js
```diff
@@ -9,6 +9,7 @@ import {
   SORT_FAILURE,
   CLEAR_TABLE,
 } from '../constants/table';
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/components/ErrorBoundary/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+        <ErrorBoundary>
+          <Welcome name="MA" />
+        </ErrorBoundary>
+      ))
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+  it('logs the error', () => {
+    const logErrorMock = jest.fn();
+    logging.logError = logErrorMock;
```
adamstankiewicz: Unnecessary to create the `logErrorMock` variable since your assertion below does not assert no it; it asserts that logging.logError is called with the specific error, but not `logErrorMock`. You could just do:
```js
logging.logError = jest.fn();
```

src/components/ErrorBoundary/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { shallow } from 'enzyme';
+
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above; we should only import the functions we need for the file.

src/components/ErrorBoundary/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+import ErrorBoundary from './index';
+
+global.newrelic = {
+  addPageAction: jest.fn(),
+  noticeError: jest.fn(),
+};
+
+const welcomeError = new Error('Catch me if you can!');
```
adamstankiewicz: This variable is only used in 1 place right now. Should we move its declaration into the test that uses it?

src/components/ErrorBoundary/index.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import * as logging from '../../Logging';
+
+class ErrorBoundary extends React.Component {
+  componentDidCatch(error, info) { // eslint-disable-line no-unused-vars
```
adamstankiewicz: If you don't need the `info` variable, you don't need to provide it as an argument.

src/components/RequestCodesPage/index.jsx
```diff
@@ -9,6 +9,8 @@ import LoadingMessage from '../LoadingMessage';
 
 import LmsApiService from '../../data/services/LmsApiService';
 
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/data/actions/coupons.js
```diff
@@ -4,6 +4,7 @@ import {
   COUPONS_FAILURE,
   CLEAR_COUPONS,
 } from '../constants/coupons';
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/data/actions/csv.js
```diff
@@ -7,6 +7,7 @@ import {
   CLEAR_CSV,
 } from '../constants/csv';
 import store from '../store';
+import * as logging from '../../Logging';
```
adamstankiewicz: Same as above.

src/Logging/index.js
```diff
@@ -0,0 +1,29 @@
+function logInfo(message) {
```
adamstankiewicz: I'm wondering if `src/Logging/index.js` is the right place for this file. It might make more sense to have as another file in our `services` directory instead, e.g. `src/data/services/NewRelicService.js`. Thoughts?

PR #157, opened 2019-01-18
Adding tests for code management related components

adamstankiewicz:
This PR adds and/or updates tests for the following components:
* `Coupon`
* `CodeManagementPage`
* `CouponDetails`
* `CodeAssignmentModal`
  * The tests for this modal are still minimal; will be completed in a subsequent PR to keep this PR size down.
* `EnterpriseApp`
  * Fixes a test to avoid a prop type warning.

### DIFF
--- a/src/components/Coupon/Coupon.test.jsx
+++ b/src/components/Coupon/Coupon.test.jsx
@@ -3,7 +3,7 @@ import { Provider } from 'react-redux';
 import renderer from 'react-test-renderer';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
-import { mount } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { MemoryRouter } from 'react-router-dom';
 
 import Coupon from './index';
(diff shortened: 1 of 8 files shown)


src/components/Coupon/Coupon.test.jsx
```diff
@@ -119,5 +123,71 @@ describe('<Coupon />', () => {
...
+      simulateCouponKeyDown('Enter');
+      expect(wrapper.find(Coupon).find(CouponDetails).prop('isExpanded')).toBeTruthy();
+
+      simulateCouponKeyDown('Enter');
+      expect(mockOnCollapse).toBeCalledTimes(1);
+      expect(wrapper.find(Coupon).find(CouponDetails).prop('isExpanded')).toBeFalsy();
+    });
+
+    it('does not handle unknown keydown event', () => {
+      const mockOnExpandOrOnCollapse = jest.fn();
```
christopappas: question: should we be asserting things about `mockOnExpandOrOnCollapse` here?

adamstankiewicz: Good catch, @christopappas! I meant to remove `mockOnCollapse` and assert that `mockOnExpandOrOnCollapse` is not called instead of `mockOnCollapse`!

adamstankiewicz: Fixed!

PR #159, opened 2019-01-24
ENT-1455: Remove Unneeded Error Handling

BessieSteinberg:
- The ```/enterprise/api/v0/enterprise/<enterprise_uuid>/enrollments```
  api used to return 404 if there were no enrollments for that
  enterprise.  So, the front-end code had to catch that error and
  display an empty list.
- That error handling is no longer necessary because the api returns an
  empty list now. (https://github.com/edx/edx-enterprise-data/pull/106)

### DIFF
--- a/src/data/actions/table.js
+++ b/src/data/actions/table.js
@@ -70,12 +70,6 @@ const paginateTable = (tableId, fetchMethod, pageNumber) => (
       dispatch(paginationSuccess(tableId, response.data, options.ordering));
     }).catch((error) => {
       NewRelicService.logAPIErrorResponse(error);
-      // This endpoint returns a 404 if no data exists,
-      // so we convert it to an empty response here.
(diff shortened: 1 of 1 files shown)


BessieSteinberg:
@adamstankiewicz and @georgebabey the API changes are merged (https://github.com/edx/edx-enterprise-data/pull/106) but I wanted to check, are there any manual steps to deploying the edx-enterprise-data repo?

adamstankiewicz:
@BessieSteinberg To deploy `edx-enterprise-data`, you'll need to version bump that repo (see https://github.com/edx/edx-enterprise-data/pull/108/files for an example), and then create/push a new tag for that version in order for the pypi deploy to kick off.

Then, in edx-analytics-data-api, you'll need to update the requirement file(s) to reflect the new edx-enterprise-data version, and deploy that.

PR #161, opened 2019-01-29
Fix selected toggles demo data 

mushtaqak:
This PR fixes a miner bug where we get `used > available`  redemptions(screenshot below)

demo data - used available

This also modifies filters data such that data resembles data returned by backend api.

Below is short description for modified filters in demo data:

**Unassigned**: Codes with assignedTo ket as `undefined`.
**Unredeemed**: Assigned codes with assignedTo key set.
**Partially redeemed**: Codes with at-least one redemption but not completely redeemed (there is no 
 way in frontend to determine from the `used` key if a code was redeemed or assigned.)
**Redeemed**: Codes with `used === available`.

[ENT-1471](https://openedx.atlassian.net/browse/ENT-1471)

### DIFF
--- a/src/demo/coupons.js
+++ b/src/demo/coupons.js
@@ -5,7 +5,7 @@ import { configuration } from '../config';
 const couponsCount = 15;
 const codesCount = 65;
 
-const minRedemptionsCount = 0;
+const minRedemptionsCount = 1;
 const maxRedemptionsCount = 5;
 
 let firstCouponHasError = false;
@@ -15,7 +15,7 @@ if (configuration.NODE_ENV === 'development') {
   firstCouponHasError = true;
 }
 
-// Coupon details filter.
+// Coupon details filters.
 const unassignedCodesFilter = 'unassigned';
 const unredeemedCodesFilter = 'unredeemed';
 const partiallyRedeemedCodesFilter = 'partially-redeemed';
@@ -99,9 +99,9 @@ const getAllCodes = (couponHasError = false) => [...Array(codesCount)].map((_, i
 
   const maxUsed = maxRedemptionsCount - redemptionsAvailablePerCode;
   const redemptionsUsedPerCode = isAssigned ? faker.random.number({
-    min: minRedemptionsCount + 1,
-    max: codeHasError ? redemptionsAvailablePerCode : maxUsed,
-  }) : minRedemptionsCount;
+    min: minRedemptionsCount,
+    max: codeHasError ? maxUsed : redemptionsAvailablePerCode,
+  }) : faker.random.number({ min: minRedemptionsCount - 1, max: minRedemptionsCount });
 
   const errorMessage = `Unable to deliver email to ${assignedTo.name} (${assignedTo.email})`;
   return {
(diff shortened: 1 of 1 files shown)


src/demo/coupons.js
```diff
@@ -119,18 +119,17 @@ const getAllCodes = (couponHasError = false) => [...Array(codesCount)].map((_, i
 const allCodes = getAllCodes();
 
 const unassignedCodes = allCodes.filter(code => (
-  code.redemptions.available > minRedemptionsCount && code.redemptions.used < maxRedemptionsCount
+  code.assigned_to === undefined && code.redemptions.used < code.redemptions.available
 ));
 const unredeemedCodes = allCodes.filter(code => (
-  (code.redemptions.available === minRedemptionsCount &&
-    code.redemptions.used > minRedemptionsCount && code.redemptions.used < maxRedemptionsCount)
+  code.assigned_to !== undefined && code.redemptions.used < code.redemptions.available
```
adamstankiewicz: This filter is for unredeemed codes. The logic here should be `code.redemptions.used === 0`, otherwise you'll get codes that have been partially redeemed.

src/demo/coupons.js
```diff
@@ -119,18 +119,17 @@ const getAllCodes = (couponHasError = false) => [...Array(codesCount)].map((_, i
 const allCodes = getAllCodes();
 
 const unassignedCodes = allCodes.filter(code => (
-  code.redemptions.available > minRedemptionsCount && code.redemptions.used < maxRedemptionsCount
+  code.assigned_to === undefined && code.redemptions.used === minRedemptionsCount - 1
```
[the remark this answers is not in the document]
adamstankiewicz: @mushtaqak The unassigned codes filter should return codes that have not been assigned and do not have any redemptions yet.

```
code.assigned_to === undefined && code.redemptions.used === 0
```

src/demo/coupons.js
```diff
@@ -99,9 +99,9 @@ const getAllCodes = (couponHasError = false) => [...Array(codesCount)].map((_, i
 
   const maxUsed = maxRedemptionsCount - redemptionsAvailablePerCode;
   const redemptionsUsedPerCode = isAssigned ? faker.random.number({
-    min: minRedemptionsCount + 1,
-    max: codeHasError ? redemptionsAvailablePerCode : maxUsed,
-  }) : minRedemptionsCount;
+    min: minRedemptionsCount,
```
adamstankiewicz: If we're always setting the `min` to `1`, we won't have any unassigned codes with 0 redemptions used.

adamstankiewicz: Nevermind, I see that this is taken into account on [line 104](https://github.com/edx/edx-portal/pull/161/files#diff-f6af3eb2f15b8537b5eba2432574fe84R104).

src/demo/coupons.js
mushtaqak: Actually `used` key from backend actually calculates this based on a voucher's assignments + it's redemptions ... and `available` is `max_use` ... so if a voucher has 10 `max_use`, it has 10 potential seats for `used` column, when 1 code is assigned ( it means 1 assignment used / 10 available), we still get to get the voucher in UNASSIGNED filter because it still has 9 potential seats left ... 

We will get the voucher in unassigned filter unless it has no slot available left for assignment ...

mushtaqak: The unredeemed filter means it gets the coupon vouchers which have been assigned but have not yet been redeemed ...

However, that said, this should have a had assigned_to value set ... I will need to fix this.

src/demo/coupons.js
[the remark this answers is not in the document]
mushtaqak: Previously, I had done that, but when we think of backend, we see that any voucher having 4/5 **redemptions** has one potential slot left which can be assigned ...

Here **redemption** in UI is a bit confusing as well ... should have been something like *voucher usage* or something more better.

mushtaqak:
@adamstankiewicz I have replied to your concerns above, let me know if they don't make sense and we need to update them.

I will still need to fix one bug in unredeemed filter with no assigned_to value.

src/demo/coupons.js
mushtaqak: Fixed `assigned_to` value set if a voucher is assigned otherwise it is `undefined`

src/demo/coupons.js
```diff
@@ -119,18 +119,17 @@ const getAllCodes = (couponHasError = false) => [...Array(codesCount)].map((_, i
...
+  code.assigned_to === undefined && code.redemptions.used < code.redemptions.available
 ));
 const unredeemedCodes = allCodes.filter(code => (
-  (code.redemptions.available === minRedemptionsCount &&
-    code.redemptions.used > minRedemptionsCount && code.redemptions.used < maxRedemptionsCount)
+  code.assigned_to !== undefined && code.redemptions.used < code.redemptions.available
 ));
 const partiallyRedeemedCodes = allCodes.filter(code => (
-  code.redemptions.available > minRedemptionsCount && code.redemptions.used > minRedemptionsCount
+  (code.redemptions.used > minRedemptionsCount &&
```
mushtaqak: **Note to reviewers:** Ideally this should be `code.redemptions.used => minRedemptionsCount` (at-least 1 redemption) but for the sake of demo data, `partiallyRedeemedCodes` need to be different than `unRedeemedCodes`, we had to do this (since there is no distinction in frontend demo data for differentiating unredeemed/redeemed assignments and redemptions used).

PR #163, opened 2019-02-05
Updating CouponDetails selection logic to handle select all

adamstankiewicz:
This PR updates the code selection logic to better handle the selection of individual codes as well as selecting all codes through the "select all" checkbox in the table header.

If there are more codes than shown on the page, a status alert is shown, prompting the user to select all codes across all pages.

The `CodeAssignmentModal` component is updated to handle when codes across all pages are selected.

If codes are selected on a page, but not all of them, a partial "select all" checkbox is displayed, as seen below.

**UI Example of Selection Checkboxes**
image

image 2

pasted_image_at_2019-02-05__4_46_pm

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -3,11 +3,10 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { Button, CheckBox, Icon, InputSelect } from '@edx/paragon';
 
-import H3 from '../H3';
-
 import TableContainer from '../../containers/TableContainer';
 import DownloadCsvButton from '../../containers/DownloadCsvButton';
 import CodeAssignmentModal from '../../containers/CodeAssignmentModal';
+import H3 from '../H3';
 import StatusAlert from '../StatusAlert';
 
 import EcommerceApiService from '../../data/services/EcommerceApiService';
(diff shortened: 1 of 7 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -281,23 +313,83 @@ class CouponDetails extends React.Component {
...
-      this.setState({
-        selectedCodes: [...selectedCodes, code.code],
-      });
+      selectedCodes = [...selectedCodes, code];
     } else {
       // Remove code from selected codes array
-      this.setState({
-        selectedCodes: selectedCodes.filter(codeTitle => codeTitle !== code.code),
-      });
+      selectedCodes = selectedCodes.filter(selectedCode => selectedCode.code !== code.code);
```
georgebabey: the `code` versus `code.code` is pretty confusing in this method - especially since selectedCodes stores the enveloped `code`. Maybe just a comment to explain whats in selectedCodes array versus the actual code in `code.code`

adamstankiewicz: @georgebabey This could be changed to:
```js
selectedCodes = selectedCodes.filter(selectedCode => selectedCode !== code)
```
Would that be clearer?

georgebabey: Yes, I think that is easier to read 👍

PR #172, opened 2019-02-07
Only send code name instead of full code object

adamstankiewicz:


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -242,7 +242,7 @@ class CodeAssignmentModal extends React.Component {
 
       // Only includes `codes` in `options` if not all codes are selected.
       if (!hasAllCodesSelected) {
-        options.codes = selectedCodes;
+        options.codes = selectedCodes.map(selectedCode => selectedCode.code);
       }
     } else {
       options.emails = [formData['email-address']];
```
brittneyexline: we also need it in the line below to be code.code

adamstankiewicz: Added!

PR #175, opened 2019-02-08
Upgrade to frontend-auth 3.0.1.

douglashall:


package.json
```diff
@@ -19,7 +19,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^1.0.0",
-    "@edx/frontend-auth": "^1.3.0",
+    "@edx/frontend-auth": "^2.0.0",
```
davidjoy: As of this writing, we're up to 3.0.1. 😄

davidjoy:
I've pushed a new commit to this PR that bumps the frontend-auth version to 3.0.1, which is the version that fixes the issue with ES6 code finding its way into an ES5 bundle, causing us to fail es-check.

PR #177, opened 2019-02-13
Remind all coupon assignments

mushtaqak:
This PR adds the capability to send reminders to all assignments associated with a coupon.

Related backend PR: https://github.com/edx/ecommerce/pull/2132

[ENT-1540](https://openedx.atlassian.net/browse/ENT-1540)

### DIFF
--- a/src/components/CodeReminderModal/index.jsx
+++ b/src/components/CodeReminderModal/index.jsx
@@ -20,6 +20,7 @@ class CodeReminderModal extends React.Component {
 
     this.validateFormData = this.validateFormData.bind(this);
     this.handleModalSubmit = this.handleModalSubmit.bind(this);
+    this.getNumberOfSelectedCodes = this.getNumberOfSelectedCodes.bind(this);
   }
 
   componentDidMount() {
@@ -50,6 +51,21 @@ class CodeReminderModal extends React.Component {
     }
   }
 
+  getNumberOfSelectedCodes() {
+    const {
+      data: { selectedCodes },
+      couponDetailsTable: { data: tableData },
+    } = this.props;
+
+    let numberOfSelectedCodes = 0;
+    if (selectedCodes && selectedCodes.length) {
+      numberOfSelectedCodes = selectedCodes.length;
+    } else if (tableData && tableData.count) {
+      numberOfSelectedCodes = tableData.count;
+    }
+    return numberOfSelectedCodes;
+  }
+
   validateFormData(formData) {
     const emailTemplateKey = 'email-template';
     const errors = {
@@ -78,14 +94,10 @@ class CodeReminderModal extends React.Component {
     const {
       couponId,
       isBulkRemind,
+      selectedToggle,
       data,
       sendCodeReminder,
     } = this.props;
-
-    /* eslint-disable no-underscore-dangle */
-    const errors = {
-      _error: [],
-    };
 
     // Validate form data
     this.validateFormData(formData);
@@ -95,12 +107,9 @@ class CodeReminderModal extends React.Component {
       template: formData['email-template'],
     };
 
-    if (isBulkRemind) {
-      if (!data.selectedCodes.length) {
-        errors._error.push('At least one code must be selected.');
-        throw new SubmissionError(errors);
-      }
-
+    if (isBulkRemind && !data.selectedCodes.length) {
+      options.code_filter = selectedToggle;
+    } else if (isBulkRemind && data.selectedCodes.length) {
       options.assignments = data.selectedCodes.map(code => ({
         email: code.assigned_to,
         code: code.code,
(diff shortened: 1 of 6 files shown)


src/components/CodeReminderModal/index.jsx
```diff
@@ -137,9 +131,12 @@ class CodeReminderModal extends React.Component {
           )}
           {isBulkRemind && data.selectedCodes && (
             <React.Fragment>
-              {data.selectedCodes.length > 0 && <p className="bulk-selected-codes">Selected Codes: {data.selectedCodes.length}</p>}
+              <p className="bulk-selected-codes">
+                Selected Codes: {data.selectedCodes.length > 0 ? data.selectedCodes.length : 'All'}
```
adamstankiewicz: Good call on saying "All" here when all codes are selected. This was something I considered adding as well. I think we should do this for the other modals (i.e., assign/revoke) in a separate PR. I'll file a ticket for it.

src/components/CodeReminderModal/index.jsx
```diff
@@ -137,9 +131,12 @@ class CodeReminderModal extends React.Component {
...
+              </p>
             </React.Fragment>
           )}
+
```
adamstankiewicz: nit: Unnecessary new line.

src/components/CodeReminderModal/index.jsx
```diff
@@ -187,12 +184,13 @@ class CodeReminderModal extends React.Component {
...
   renderTitle() {
-    const { title } = this.props;
-
+    const { title, data } = this.props;
     return (
       <React.Fragment>
         <span className="d-block">{title}</span>
-        <small>Assignment Reminder</small>
+        <small>
+          {data.selectedCodes && data.selectedCodes.length === 0 ? 'Remind All' : 'Assignment Reminder'}
```
adamstankiewicz: Similar to the comment above, I think this is something we should do for the other modals (i.e., assign, revoke) as well.

src/components/CouponDetails/index.jsx
```diff
@@ -106,7 +110,7 @@ class CouponDetails extends React.Component {
     }, {
       label: 'Remind',
       value: 'remind',
-      disabled: isAssignView || isRedeemedView || selectedCodes.length === 0,
+      disabled: isAssignView || isRedeemedView || (tableData && tableData.count === 0),
```
adamstankiewicz: I think it would make sense if all options are disabled when there's no table data. I would recommend creating a new variable:

```js
const hasTableData = tableData && tableData.count;
```
Then, use it in the `disabled` conditional for each of the options.

src/containers/CouponDetails/CouponDetails.test.jsx
```diff
@@ -195,7 +195,7 @@ describe('CouponDetailsWrapper', () => {
       ));
     });
 
-    it('sets remind modal state on Revoke button click', () => {
+    it('sets remind modal state on Remind button click', () => {
```
adamstankiewicz: 👍

src/components/CodeReminderModal/index.jsx
adamstankiewicz: Alternatively, I wonder if it makes more sense to state the actual number of codes instead of just saying "All". Like we do when the user manually selects codes across all pages, the `CodeAssignmentModal` will say the total number of codes across all pages.

adamstankiewicz: I chatted with Jennifer about showing "All" versus the actual number. We decided that showing the actual number of codes is the approach we would want to take to reduce ambiguity around how many "All" is.

To get the total number, you can map the coupon details table data to the `CodeReminderModal` container similar to the [`CodeAssignmentModal` container](https://github.com/edx/edx-portal/blob/master/src/containers/CodeAssignmentModal/index.jsx#L7).

mushtaqak: Ok 👍 

I just looked at the code for codeAssignment selection handling - PR https://github.com/edx/edx-portal/pull/163 , I guess, we should add it to other filter views too at-least Unredeemed (may be in a separate ticket). That's a nice UX addition (making things more explicit for assign-all/remind-all/revoke-all). My just concern is bulk "revoke", which would need key value pair ...

mushtaqak:
@adamstankiewicz addressed your feedback :)

src/components/CodeReminderModal/index.jsx
adamstankiewicz: @mushtaqak If we're supporting bulk revoke in the API, I would assume we would need to do similar work as you did for the remind all work, where we make the `assignments` field option and revoke all if `assignments` is not provided?

mushtaqak: makes sense, once revoke-all is implemented we can use (select-all) UI for unredeemed and partially-redeemed toggles as well. I will document this in a separate JIRA story.

mushtaqak:
@mushtaqak This is ready for your final review. This also includes supporting remind all for partial-redeemed toggle view as requested by @brittneyexline on backend PR.

PR #179, opened 2019-02-14
 Fetch overview data for specific coupon after successful assign/revoke

adamstankiewicz:


src/data/actions/coupons.js
```diff
@@ -51,6 +59,18 @@ const fetchCouponOrders = options => (
   }
 );
 
+const fetchCouponOrder = couponId => (
+  dispatch => (
+    EcommerceApiService.fetchCouponOrders({ coupon_id: couponId })
+      .then((response) => {
+        dispatch(fetchCouponOrderSuccess(response.data));
+      })
+      .catch((error) => {
```
brittneyexline: do we need a corresponding coupon failure action?

adamstankiewicz: @brittneyexline I didn't include a failure action here because I was unsure what should happen if the request does fail. I suppose we could throw an error status alert message and display that above the coupon details table, with the message including a link to retry fetching the data.

I think I would prefer to do that in a separate PR, though. Thoughts?

PR #182, opened 2019-02-15
 Hide partial redeem toggle for single use and once per customer coupons

adamstankiewicz:
[ENT-1580](https://openedx.atlassian.net/browse/ENT-1580)

This PR removes the "Partially Redeemed" toggle option for "Single use" and "Once per customer" coupons.

I also did some refactoring to pass through all the data associated with a coupon to the `CouponDetails` component instead of as individual props, e.g.:

```jsx
<CouponDetails
  isExpanded={isExpanded}
  couponData={data}
/>
```

instead of:

```jsx
<CouponDetails
  isExpanded={isExpanded}
  id={data.id}
  hasError={data.has_error}
  unassignedCodes={data.num_unassigned}
  couponTitle={data.title}
  // etc...
/>
```

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -11,6 +11,7 @@ import CodeRevokeModal from '../../containers/CodeRevokeModal';
 import H3 from '../H3';
 import StatusAlert from '../StatusAlert';
 
+import { SINGLE_USE, ONCE_PER_CUSTOMER } from '../../data/constants/coupons';
 import EcommerceApiService from '../../data/services/EcommerceApiService';
 
 import './CouponDetails.scss';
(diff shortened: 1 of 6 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -90,9 +90,34 @@ class CouponDetails extends React.Component {
     }
   }
 
+  getTableFilterSelectOptions() {
+    const { couponData: { usage_limitation: usageLimitation } } = this.props;
+    const shouldHidePartialRedeem = ['Single use', 'Once per customer'].includes(usageLimitation);
```
christopappas: Does JS like global constants at all? I know in Python I'd probably make 'Single Use' a variable that could be imported/referenced in case we change 'Single use' to 'Single Use' or something like that. Is that a thing that 1) people do? 2) is worth doing? Just a thought anyway!

src/containers/CouponDetails/CouponDetails.test.jsx
```diff
@@ -150,6 +162,63 @@ describe('CouponDetailsWrapper', () => {
...
+          usage_limitation: 'Single use',
+        }}
+        isExpanded
+      />
+    ));
+    options = wrapper.find('select').first().prop('children').map(option => option.props.value);
+    expect(options).toHaveLength(3);
+    expect(options.includes('partially-redeemed')).toBeFalsy();
+
+    wrapper = mount((
```
christopappas: I think this might be the same bit as in the following test?

adamstankiewicz:
@christopappas Addressed your feedback and will merge this now, thanks! 😄

PR #183, opened 2019-02-21
Render error status alert if individual coupon overview request fails

adamstankiewicz:
[ENT-1561](https://openedx.atlassian.net/browse/ENT-1561)

Currently, after a successful assign/revoke, we fetch the overview data for the coupon such that the number of `Unassigned Codes` is updated accordingly. However, in the case where the API request to fetch this data fails, we should display an error status alert informing the user of this. We should also offer a way to retry fetching that data.

This PR addresses that failure state.


src/components/CouponDetails/index.jsx
```diff
@@ -569,12 +573,12 @@ class CouponDetails extends React.Component {
   }
 
   renderSuccessMessage({ title, message }) {
-    const { hasError } = this.props;
+    const { hasError, couponOverviewError } = this.props;
```
christopappas: I think you said you had an update in mind for this, but wanted to drop a note here that the variable names are potentially a tad confusing here 👍

adamstankiewicz: Yeah, I realized the names are potentially confusing but am opting to address renaming them to be clearer in a subsequent PR.

PR #184, opened 2019-02-22
ENT-1522: Update coupon code remind/revoke backend and frontend with proper error checking

saleem-latif:
**JIRA TIcket:** [ENT-1522](https://openedx.atlassian.net/browse/ENT-1522)

**Description:**
Currently, remind/revoke endpoints return HTTP 200 even if there is an error. We do send the error message in the returned response but frontend ignores that and show a generic successful message which is misleading because, in reality, it happens that remind/revoke failed for every assignment and we showed the successful message on the frontend. It would be nice to consider the error messages coming from the server and either highlight the rows to show that remind/revoke failed or some better way to convey the error message.

**Acceptance Criteria:**
Correctly show the error message to the user if there is an error with remind/revoke.

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -56,6 +56,7 @@ class CouponDetails extends React.Component {
       isCodeAssignmentSuccessful: undefined,
       isCodeReminderSuccessful: undefined,
       isCodeRevokeSuccessful: undefined,
+      doesCodeActionHaveErrors: undefined,
       selectedCodes: [],
       hasAllCodesSelected: false,
       /**
@@ -351,6 +352,7 @@ class CouponDetails extends React.Component {
       isCodeAssignmentSuccessful,
       isCodeReminderSuccessful,
       isCodeRevokeSuccessful,
+      doesCodeActionHaveErrors,
     } = this.state;
 
     const hasStatusAlert = [
@@ -359,6 +361,7 @@ class CouponDetails extends React.Component {
       isCodeAssignmentSuccessful,
       isCodeReminderSuccessful,
       isCodeRevokeSuccessful,
+      doesCodeActionHaveErrors,
       this.shouldShowSelectAllStatusAlert(),
     ].some(item => item);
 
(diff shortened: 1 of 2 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -479,6 +486,12 @@ class CouponDetails extends React.Component {
 
     this.resetCodeActionStatus();
 
+    if (doesCodeActionHaveErrors) {
+      this.setState({
```
muhammad-ammar: It would be great if we can combine this state update with the one below.

saleem-latif: Done

src/components/CouponDetails/index.jsx
```diff
@@ -56,6 +56,7 @@ class CouponDetails extends React.Component {
       isCodeAssignmentSuccessful: undefined,
       isCodeReminderSuccessful: undefined,
       isCodeRevokeSuccessful: undefined,
+      doesCodeActionHaveErrors: undefined,
```
mushtaqak: `hasCodeActionErrors` sounds more better?

src/components/CouponDetails/index.jsx
```diff
@@ -483,11 +490,15 @@ class CouponDetails extends React.Component {
 
     this.resetCodeActionStatus();
 
-    if (stateKey) {
+    if (stateKey || doesCodeActionHaveErrors) {
```
mushtaqak: We can only check one variable, if `stateKey` is `null` so would be `doesCodeActionHaveErrors`.

src/components/CouponDetails/index.jsx
```diff
@@ -483,11 +490,15 @@ class CouponDetails extends React.Component {
...
       this.setState({
-        [stateKey]: true,
-        refreshIndex: this.state.refreshIndex + 1, // force new table instance
-        selectedCodes: [],
+        ...(stateKey ? {
```
mushtaqak: Can we not just add `doesCodeActionHaveErrors` in the state? It would be simpler?

saleem-latif:
@mushtaqak done, please take another look.

src/components/CouponDetails/index.jsx
saleem-latif: I think `doesCodeActionHaveErrors` is better as it conforms to the style of boolean variable names defined above.

PR #188, opened 2019-02-25
Support assigning more than num selected codes for "once per customer" coupons

adamstankiewicz:
[ENT-1573](https://openedx.atlassian.net/browse/ENT-1573)

For "Once per customer" coupons, we should allow the user to assign more emails than the number of selected codes since an individual code can be assigned to multiple users.

This PR removes the validation check for "Once per customer" coupons that looks to see if the number of emails provided is <= the number of selected codes. This will allow more emails to be assigned than the number of selected codes per the ticket description.

If a particular cannot be assigned, the modal will show an error and display the associated error message. In some cases, the response will contain an array of `non_field_errors` which should be included in the error status alert within the modal. If these `non_field_errors` exist in the response, we now include them in the status alert.

### DIFF
--- a/src/components/CodeAssignmentModal/index.jsx
+++ b/src/components/CodeAssignmentModal/index.jsx
@@ -11,6 +11,7 @@ import BulkAssignFields from './BulkAssignFields';
 import IndividualAssignFields from './IndividualAssignFields';
 
 import emailTemplate from './emailTemplate';
+import { ONCE_PER_CUSTOMER } from '../../data/constants/coupons';
 
 import './CodeAssignmentModal.scss';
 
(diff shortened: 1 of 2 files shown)


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -145,7 +146,10 @@ class CodeAssignmentModal extends React.Component {
 
       errors[csvFileKey] = message;
       errors._error.push(message);
-    } else if (csvEmails && numberOfSelectedCodes && csvEmails.length > numberOfSelectedCodes) {
+    } else if (
+      csvEmails && numberOfSelectedCodes && couponType !== ONCE_PER_CUSTOMER &&
```
brittneyexline: i think we should include MULTI_USE codes as well, because they can be assigned to multiple users too.

adamstankiewicz: @brittneyexline This now includes MULTI_USE codes as well!

PR #191, opened 2019-02-27
Fixed request codes success banner not properly showing bug.

Tj-Tracy:
It was an issue with remounting the code management page after redirecting back from the request form. It was solved by checking props on mount instead of on receiving now props. Also changed a test to reflect the new structure.

### DIFF
--- a/src/components/CodeManagement/index.jsx
+++ b/src/components/CodeManagement/index.jsx
@@ -25,32 +25,24 @@ class CodeManagement extends React.Component {
   }
 
   componentDidMount() {
-    const { enterpriseId } = this.props;
-
+    const { enterpriseId, location } = this.props;
     if (enterpriseId) {
       this.props.fetchCouponOrders();
+    }
+    if (location.state && location.state.hasRequestedCodes) {
+      this.setState({ // eslint-disable-line react/no-did-mount-set-state
+        hasRequestedCodes: location.state.hasRequestedCodes,
+      });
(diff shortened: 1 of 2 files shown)


src/components/CodeManagement/index.jsx
```diff
@@ -25,33 +25,25 @@ class CodeManagement extends React.Component {
...
       this.props.fetchCouponOrders();
     }
-
-    if (location.state && location.state.hasRequestedCodes && location !== prevProps.location) {
-      this.setState({ // eslint-disable-line react/no-did-update-set-state
-        hasRequestedCodes: location.state.hasRequestedCodes,
-      });
-
-      // reset location state so the "requested more codes" status alert disappears
-      history.replace({
```
adamstankiewicz: I think this `history.replace({ ... })` part also should be moved up to `componentDidMount`.

If you request more codes, the success status alert appears. If you then go to the request more codes page, and then immediately back to the code management page, the status alert still appears. I would expect it to disappear when you switch pages.

src/components/CodeManagement/index.jsx
```diff
@@ -25,33 +25,25 @@ class CodeManagement extends React.Component {
...
-        ...location.pathname,
-        state: {},
-      });
-    }
   }
 
+
```
adamstankiewicz: nit: Unnecessary new line

PR #192, opened 2019-03-05
Fixes pagination / history for Code Management

adamstankiewicz:
[ENT-1491](https://openedx.atlassian.net/browse/ENT-1491)

For coupon pagination, we modify the `overview_page` query parameter. For code pagination (coupon details), we modify the `page` query parameter.

When a coupon is expanded, a `coupon_id` query parameter is added. This gets removed when the coupon is collapsed. On initial page load, if the `coupon_id` query parameter is set, the component will try to expand the coupon with that id. If the id is not in the results from the API, the `coupon_id` query parameter is removed.

For example, if you expand a coupon on page 2 of the overview page, and are on page 3 of the coupon details, the URL would look something like: `/coupons?coupon_id=1&overview_page=2&page=3`.

This now supports pagination for both coupons and codes, as well as deep linking to specific coupons.

I've also tested how the component handles history (i.e., browser back/forward buttons).

### DIFF
--- a/src/data/services/EcommerceApiService.js
+++ b/src/data/services/EcommerceApiService.js
@@ -11,7 +11,7 @@ class EcommerceApiService {
     const { enterpriseId } = store.getState().portalConfiguration;
     const queryParams = {
       page: 1,
-      page_size: 50,
+      page_size: 2,
       ...options,
     };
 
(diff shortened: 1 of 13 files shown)


src/data/services/EcommerceApiService.js
```diff
@@ -11,7 +11,7 @@ class EcommerceApiService {
     const { enterpriseId } = store.getState().portalConfiguration;
     const queryParams = {
       page: 1,
-      page_size: 50,
+      page_size: 2,
```
christopappas: based on our convo, take a look here!

mushtaqak:
@adamstankiewicz One suggestion can be to show pagination widget even though a page is not found (with no page selected), letting user know page does not exist. But this can be done in a separate ticket and PR.

PR #194, opened 2019-03-08
Restrict bounced codes on redeemed toggle

mushtaqak:
Restrict bounced codes on redeemed toggle because it does not make sense to show error messages for codes that do not appear on unassigned, redeeemed or partial-redeemed toggles.

[ENT-1635](https://openedx.atlassian.net/browse/ENT-1635)

### DIFF
--- a/src/containers/CouponDetails/CouponDetails.test.jsx
+++ b/src/containers/CouponDetails/CouponDetails.test.jsx
@@ -10,6 +10,7 @@ import { mount } from 'enzyme';
 import { SINGLE_USE, MULTI_USE, ONCE_PER_CUSTOMER } from '../../data/constants/coupons';
 import EcommerceaApiService from '../../data/services/EcommerceApiService';
 
+import CouponDetailsComponent from '../../components/CouponDetails';
 import CouponDetails from './index';
 
 const mockStore = configureMockStore([thunk]);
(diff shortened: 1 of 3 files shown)


src/containers/CouponDetails/CouponDetails.test.jsx
```diff
@@ -120,6 +120,7 @@ describe('CouponDetailsWrapper', () => {
     });
 
     it('with error', () => {
+      // TODO: set selectedToggle: unredeemed state on component mount
```
mushtaqak: @adamstankiewicz @muhammad-ammar Any thoughts how to set component only state here?

adamstankiewicz: @mushtaqak You could try something like this:
```jsx
import CouponDetailsComponent from '../../components/CouponDetails';
// ...

it('with error', () => {
  const tree = renderer.create((
    <CouponDetailsWrapper
      couponData={{
        ...initialCouponData,
        errors: [{ code: 'test-code', user_email: 'test@example.com' }],
      }}
      isExpanded
    />
  ));
  tree.root.findByType(CouponDetailsComponent).instance.setState({ selectedToggle: 'unredeemed' });
  expect(tree.toJSON()).toMatchSnapshot();
});
```

mushtaqak: This works 👍 

even though I had tried this approach before but this didn't work ... there is only one exception I was finding the container `CouponDetails` whereas it had to be the underlying CouponDetails component `import CouponDetailsComponent from '../../components/CouponDetails';` ... ughhh miner mistakes  🤕

adamstankiewicz: Glad it works! Yeah, if you just look for `CouponDetails`, it's actually finding the container, not the component itself. The rest of the PR looks good, though; I'll give this a 👍once the test is updated :)

PR #205, opened 2019-04-30
Add a 403 component

mushtaqak:
In this PR, instead of logging user out of the portal, we will show a 403 component whenever a 403 response is returned by the API.

- [x]  Merge https://github.com/edx/frontend-auth/pull/25

[ENT-1753](https://openedx.atlassian.net/browse/ENT-1753)

### DIFF
--- a/src/data/apiClient.js
+++ b/src/data/apiClient.js
@@ -1,4 +1,5 @@
 import { getAuthenticatedAPIClient } from '@edx/frontend-auth';
+import { NewRelicLoggingService } from '@edx/frontend-logging';
 
 const apiClient = getAuthenticatedAPIClient({
   appBaseUrl: process.env.BASE_URL,
@@ -9,6 +10,8 @@ const apiClient = getAuthenticatedAPIClient({
   refreshAccessTokenEndpoint: process.env.REFRESH_ACCESS_TOKEN_ENDPOINT,
   accessTokenCookieName: process.env.ACCESS_TOKEN_COOKIE_NAME,
   userInfoCookieName: process.env.USER_INFO_COOKIE_NAME,
+  loggingService: NewRelicLoggingService,
+  handleRefreshAccessTokenFailure: () => {},
 });
 
 export default apiClient;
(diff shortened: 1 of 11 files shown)


mushtaqak:
**Error message**
819634a2a04e2a40c2f137fb27914d0ad05ca56b implements the approach where we just change the message of the `ErrorPage`:

403-message-approach-1


**ForbiddenPage**
d1a0fc76c5f33cd68cf59ef6333c6bfe5c414a87 adds a new component `ForbiddenPage` which is shown whenever a `403` response is given:

403-component-approach-1

adamstankiewicz:
Following up on the Slack conversation, let's go with the `ForbiddenPage` component. And, per Roya's feedback, we can change the `.lead` text to "You do not have access to this page."

mushtaqak:
**Reminder before merging**: Only merge after https://github.com/edx/frontend-auth/pull/25 PR is merged.

mushtaqak:
@adamstankiewicz This ready for review :)

src/data/apiClient.js
```diff
@@ -1,5 +1,7 @@
 import { getAuthenticatedAPIClient } from '@edx/frontend-auth';
 
+import NewRelicService from './services/NewRelicService';
+
```
muhammad-ammar: We need to use the `NewRelicLoggingService` from `@edx/frontend-logging` like we are doing in https://github.com/edx/frontend-app-learner-portal/pull/13/files#diff-75f76d91899c96b633e6282624e91fbd

src/data/apiClient.js
```diff
@@ -1,5 +1,7 @@
...
 const apiClient = getAuthenticatedAPIClient({
```
adamstankiewicz: Now that https://github.com/edx/frontend-auth/pull/29 has merged, we should version bump `@edx/frontend-auth` and update our API client config to include the newly added callback to prevent the 403 logout.

```jsx
const apiClient = getAuthenticatedAPIClient({
  // ...

  /* I think this will be a noop function, as we don't want to actually handle the error 
   * here; rather, the component(s) making the API requests themselves should handle the error.
   */
  handleRefreshAccessTokenFailure={() => {}}
});
```

PR #206, opened 2019-05-08
Fix enterprise list loading screen

mushtaqak:
This PR fixes a loading screen stuck issue on enterprise list page introduced due to https://github.com/edx/edx-portal/pull/205


mushtaqak:
@adamstankiewicz This is table overlay loading message that comes for tiny bit of second before the table is loaded.

Screenshot 2019-05-08 at 6 40 20 AM

This appeared after we added the logic to only show the Enterprise List view when loading is off and `enterpriseList` is populated. Any thoughts on how to fix this better way ?

I have merged this PR since this was making enterprise admin users see "Loading ..." screen all the time.

PR #208, opened 2019-07-29
Ekii/paragonedits

erinakii:
PR is for fixing sidebar padding as well header and footer padding

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -28,3 +28,9 @@ form {
         resize: none;
     }
 }
+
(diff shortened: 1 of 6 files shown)


src/index.scss
```diff
@@ -28,3 +28,9 @@ form {
         resize: none;
     }
 }
+
+.container-fluid {
+    max-width: 100%;
```
adamstankiewicz: width: 100%;
max-width: initial;

PR #211, opened 2019-08-15
Replace 404 with loading component

Tj-Tracy:
Replace the 404 component flash with a loading component instead. This does create a situation where 2 loading screens are show in a row if the root url of the portal is hit and the user is only authorized for a single enterprise.

### DIFF
--- a/src/components/Admin/index.jsx
+++ b/src/components/Admin/index.jsx
@@ -16,7 +16,6 @@ import EnrolledLearnersForInactiveCoursesTable from '../EnrolledLearnersForInact
 import CompletedLearnersTable from '../CompletedLearnersTable';
 import PastWeekPassedLearnersTable from '../PastWeekPassedLearnersTable';
 import LearnerActivityTable from '../LearnerActivityTable';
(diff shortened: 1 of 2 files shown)


src/components/Admin/index.jsx
```diff
@@ -232,7 +231,7 @@ class Admin extends React.Component {
 
     return (
       <React.Fragment>
-        {!loading && !error && !this.hasAnalyticsData() ? <NotFoundPage /> : (
+        {!loading && !error && !this.hasAnalyticsData() ? this.renderLoadingMessage() : (
```
tuchfarber: This probably fixes it, but I'm confused why `loading` would be false if we're currently loading (and showing the loading message)

Tj-Tracy: it's confusing because one is the loading state for the page and the other is the "loading" state just for analytics data. I'd have to look more into the code to consolidate them, but idea is that since that function is waiting on data, it will return false for that time -- thus triggering the 404. Since its _really_ not a missing page (its just awaiting data), I figured loading here a second time was a workable solution.

PR #212, opened 2019-08-29
Add code search UI functionality

adamstankiewicz:
This PR adds the code search UI functionality as seen in the gif below:

code_search

Note: I am currently working on getting the action buttons (remind/revoke) in the search results table, so that a user can take action on the relevant results right from the search results table. I am trying to reuse the same logic as those action buttons from the `CouponDetails` component and will be abstracting those action buttons and their associated logic into their own reusable component. Tests are still to come.

Supporting pagination will be some extra work as our `TableContainer` component does not currently support pagination when there is more than one table on the page, as is the case here when a coupon batch is expanded. We may want to punt on supporting pagination initially, and follow-up with a new ticket to support pagination in the `TableContainer` when there are multiple tables on the page.

### DIFF
--- a/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
+++ b/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
@@ -37,14 +37,15 @@ const tableColumns = [
 ];
 
 /**
- * The search API response contains a record for each coupon batch a
- * user is associated with (i.e., has unused assignments, or actual
- * redemptions). Under each coupon batch is an array of assignments and
- * redemptiions, that contains all the metadata associated with a redemption,
- * if applicable. Because we want the search results UI to show a row
- * for each unused assignment and all redeemed courses associated with a
- * particular user, this function flattens the search results by creating
- * a new table row for each value in `redemptions_and_assignments`.
+ * The search API response contains records for all vouchers a user is associated
+ * with (i.e., has unused assignments, or actual redemptions). To get a complete
+ * grouping of assignments or redemptions for a particular coupon batch for the
+ * user, we need to combine the results in `assignments_and_redemptions` (which
+ * contains all the metadata associated with a redemption, if applicable) for
+ * all vouchers that share the same `coupon_id` field. Because we want the search
+ * results UI to show a row for each unused assignment and all redeemed courses
+ * associated with a particular user, this function flattens the search results by
+ * created a new table row for each value in the `redemptions_and_assignments` fields.
  */
 const transformSearchResults = (items) => {
   const itemsSortedByCouponId = items.sort((itemA, itemB) => itemA.coupon_id - itemB.coupon_id);
--- a/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
+++ b/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
@@ -1,0 +1,162 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import { Icon } from '@edx/paragon';
+
+import TableContainer from '../../containers/TableContainer';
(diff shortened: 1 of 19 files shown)


adamstankiewicz:
Updated gif showing the code search functionality:

code_search_ux

adamstankiewicz:
Note to reviewers: This PR is large, but mainly due to the addition of snapshot tests 😃

src/components/CodeSearchResults/CodeSearchResultsTable.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * The search API response contains a record for each coupon batch a
+ * user is associated with (i.e., has unused assignments, or actual
+ * redemptions). Under each coupon batch is an array of assignments and
+ * redemptiions, that contains all the metadata associated with a redemption,
+ * if applicable. Because we want the search results UI to show a row
+ * for each unused assignment and all redeemed courses associated with a
+ * particular user, this function flattens the search results by creating
+ * a new table row for each value in `redemptions_and_assignments`.
```
brittneyexline: might be good for @christopappas to review this part just to make sure the api contract lines up as expected

src/components/CodeSearchResults/CodeSearchResultsTable.jsx
```diff
@@ -0,0 +1,162 @@
...
+      couponName,
+      code,
+      isRedeemed: isRedeemed ? (
+        <Icon className="fa fa-check text-primary" screenReaderText="has been redeemed" />
+      ) : defaultEmptyValue,
+      courseTitle: courseTitle || defaultEmptyValue,
+      redemptionDate: redemptionDate || defaultEmptyValue,
+      actions: !isRedeemed ? (
+        <React.Fragment>
+          <RemindButton
```
brittneyexline: i like that we get to reuse these!

src/components/RevokeButton/index.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@edx/paragon';
+
+import CodeRevokeModal from '../../containers/CodeRevokeModal';
+
+class RevokeButton extends React.Component {
+  state = {
```
brittneyexline: seems like there is some shared boilerplate with the RemindButton that could potentially be consolidated?

adamstankiewicz: Agreed. Working on consolidating these with a new component (something like `ActionButtonWithModal` perhaps) using a HOC or render props :)

src/components/CodeSearchResults/CodeSearchResultsTable.jsx
```diff
@@ -0,0 +1,162 @@
...
+    key: 'courseTitle',
+  },
+  {
+    label: 'Actions',
+    key: 'actions',
+  },
+];
+
+/**
+ * The search API response contains a record for each coupon batch a
```
christopappas: The API response returns records around "vouchers" that exist, and it is possible (and likely) that a particular coupon batch can have multiple vouchers. That said, we'll know if vouchers are part of the same coupon batch based on the `coupon_id` returned in the response. Suggestion for changing the language:

`The search API response contains records for all vouchers a user/user_email is associated with (i.e., has unused assignments, or actual redemptions). To get a complete grouping of assignments or redemptions for a particular coupon batch for the user, we need to combine the results in assignments_and_redemptions (which contains all the metadata associated with a redemption, if applicable) for all vouchers that share the same coupon_id field.` Because we want the search results UI to show a row for each unused assignment and all redeemed courses associated with a particular user, this function flattens the search results by created a new table row for each value in each redemptions_and_assignments field.`

PR #213, opened 2019-09-05
Pass thru page parameter to search API request, fix table loading height

adamstankiewicz:


src/components/CodeSearchResults/CodeSearchResults.scss
```diff
@@ -3,7 +3,6 @@
 .code-search-results {
   .loading {
     @extend .mb-3;
-    height: 220px;
```
adamstankiewicz: Note: Setting `height` here was causing the table loading overlay box to take up too much space. The height on the initial load of the table data is already defined in the table component styles anyways :)

PR #214, opened 2019-09-05
Update code search to match new API response format

adamstankiewicz:
Note: ready for review, but will hold off merging until API response change is deployed on ecommerce stage 😃


christopappas:
Actually, one note: adding in `"voucher_id": 18,` to the response too so we know which thing comes from which voucher

PR #217, opened 2019-09-06
Ttracy/ent 2192 reporting config ui

Tj-Tracy:
This PR adds an MVP UI for enterprise admins to be able to manage their reporting configurations.

list:
image

first half of the form:
image

second half (email):
image

second half(sftp):
image


src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+  const [submitState, setSubmitState] = useState('default');
+  return (
+    <form
+      onSubmit={async (e) => {
+        e.preventDefault();
+        setSubmitState('pending');
+        const formData = new FormData(e.target);
+        const requiredFields = REQUIRED_FIELDS;
+        if (deliveryMethod === 'email') {
+          requiredFields.push(...REQUIRED_EMAIL_FIELDS);
```
staubina: Could you construct these as Constants above and then just set the required fields below?

```
const REQUIRED_FIELDS = [
  'hourOfDay',
];
// for the email delivery method
const REQUIRED_EMAIL_FIELDS = [
  ...REQUIRED_FIELDS,
  'email',
];
// for the sftp delivery mothod
const REQUIRED_SFTP_FIELDS = [
  ...REQUIRED_FIELDS,
  'sftpPort',
  'sftpHostname',
  'sftpUsername',
  'sftpPassword',
  'sftpFilePath',
];
```

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+        if (valid === true) {
+          // we need to transform the formdata back to snake_case before sending it
+          const transformedData = transformFormData(formData);
+          // if this is not a new config, add the uuids
+          if (config) {
+            transformedData.append('uuid', config.uuid);
+            transformedData.append('enterprise_customer_id', config.enterpriseCustomer.uuid);
+            const err = await updateConfig(transformedData, config.uuid);
+            if (err) {
+              setSubmitState('error');
```
staubina: nit: Consider making states like this that end up being used more than once a Constant.  It makes the code maintenance easier.

i.e. 'default', 'error', 'complete', etc.

src/utils.js
```diff
@@ -78,6 +80,38 @@ const isRequired = (value = '') => (isEmpty(value) ? 'This field is required.' :
 const isValidEmail = (value = '') => (!isEmail(value) ? 'Must be a valid email address.' : undefined);
 const isValidNumber = (value = '') => (!isEmpty(value) && !isNumeric(value, { no_symbols: true }) ? 'Must be a valid number.' : undefined);
 
+
+// camelCase <--> snake_case functions
```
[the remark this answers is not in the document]
staubina: It might be worth it to add a comment explaining why this is here or how it can be used.

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+  updateConfig,
+}) => {
+  const [frequency, setFrequency] = useState(config ? config.frequency : 'monthly');
+  const [deliveryMethod, setDeliveryMethod] = useState(config ? config.deliveryMethod : 'email');
+  const [invalidFields, setInvalidField] = useState({});
+  const [active, setActive] = useState(config ? config.active : false);
+  const [submitState, setSubmitState] = useState('default');
+  return (
+    <form
+      onSubmit={async (e) => {
```
staubina: Please split this method out into its own function if possible.  It will make the code a little easier to follow.

src/data/services/LmsApiService.js
```diff
@@ -50,6 +50,22 @@ class LmsApiService {
...
     return apiClient.post(requestCodesUrl, postParams, 'json');
   }
+
+  static fetchReportingConfigs() {
+    const reportingConfigUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise_customer_reporting/`;
+    return apiClient.get(reportingConfigUrl);
+  }
+
+  static postNewReportingConfig(formData) {
+    const reportingConfigUrl = `${LmsApiService.baseUrl}/enterprise/api/v1/enterprise_customer_reporting/`;
```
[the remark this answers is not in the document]
Tj-Tracy: Sounds good, but I'm not sure where to put that constant. Just in this services file? or somewhere else?

src/utils.js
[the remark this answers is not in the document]
Tj-Tracy: Added in the refactor

src/components/ReportingConfig/index.jsx
```diff
@@ -0,0 +1,119 @@
...
+class ReportingConfig extends React.Component {
+  state = {
+    loading: true,
+    reportingConfigs: [],
+  };
+
+  // eslint-disable-next-line react/sort-comp
+  newConfigFormRef = React.createRef();
+
+  async componentDidMount() {
```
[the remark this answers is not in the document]
Tj-Tracy: Nope! Removed in the refactor.

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+              return;
+            }
+          }
+          setSubmitState('complete');
+        } else {
+          // the form was not valid, highlight the invalid fields
+          setInvalidField({ ...valid });
+          setSubmitState('default');
+        }
+      }}
```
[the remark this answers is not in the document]
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+      <div className="row">
+        <div className="col">
+          <ValidationFormGroup
+            for="dayOfMonth"
+            helpText="The day of the month to send the report. This field is required and only valid when the frequency is monthly"
+          >
+            <label htmlFor="dayOfMonth">Day of Month</label>
+            <Input
+              type="number"
+              max={31}
```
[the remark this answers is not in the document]
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,436 @@
...
+              invalid={invalidFields.email}
+            >
+              <label htmlFor="email">Email(s)</label>
+              <Input
+                type="textarea"
+                id="email"
+                name="email"
+                defaultValue={config ? config.email.join('\n') : undefined}
+                onBlur={(e) => {
+                  const emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/igm;
```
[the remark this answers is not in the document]
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: handled in the refactor

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+                <Input
+                  type="textarea"
+                  id="email"
+                  name="emailRaw"
+                  defaultValue={config ? config.email.join('\n') : undefined}
+                  onBlur={(e) => {
+                    this.setState({
+                      invalidFields: {
+                        ...invalidFields,
+                        email: !EMAIL_REGEX.test(e.target.value),
```
adamstankiewicz: There is a `isValidEmail` function in `utils.js` that you could potentially use in place of instead of the custom `EMAIL_REGEX`. The utility function makes use of the `isEmail` function from the `validator` library (i.e., `import isEmail from 'validator/lib/isEmail';`). Up to you if you want to use it :)

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+        >
+          <label htmlFor="pgpEncryptionKey">PGP Ecryption Key</label>
+          <Input
+            type="textarea"
+            id="pgpEncryptionKey"
+            name="pgpEncryptionKey"
+            defaultValue={config ? config.pgpEncryptionKey : undefined}
+          />
+        </ValidationFormGroup>
+        {deliveryMethod === 'email' && (
```
adamstankiewicz: This component could probably be split into a few subcomponents. One good opportunity could be the conditional rendering based on `deliveryMethod`. For example:
```
{deliveryMethod === 'email' && <EmailDeliveryMethodFields />}
{deliveryMethod === 'sftp' && <SftpDeliveryMethodFields />}
```
There are other opportunities for subcomponents as well. This could help reduce the complexity and improve the readability of the `render` function.

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+                    defaultValue={config ? config.sftpUsername : undefined}
+                    onBlur={e => this.setState({
+                      invalidFields: {
+                        ...invalidFields,
+                        sftpUsername: !e.target.value,
+                      },
+                    })}
+                  />
+                </ValidationFormGroup>
+                {/* TODO: check if the password can be blank */}
```
adamstankiewicz: Is this still meant as a TODO? Or, should this validation be incorporated into this PR as well?

src/components/ReportingConfig/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+            className="col justify-content-center align-items-center"
+            ref={this.newConfigFormRef}
+          >
+            <div>
+              <ReportingConfigForm
+                createConfig={this.createConfig}
+              />
+            </div>
+          </Collapsible>
+
```
adamstankiewicz: nit: extra line

adamstankiewicz:
Also, a couple design improvements based on the screenshot:
1. Is there a way to expand the width of the columns in the Collapsible toggle so that "Delivery Method" doesn't wrap on 2 lines? The boxes seem a bit taller than they need to be.
2. For design consistency with other pages:
  a) the icons used for the collapsible component would ideally be chevrons (up/down) to match the convention established in the expandable coupon batches on "Code Management" page.
  b) Might want to include the hero at the top with the title of the page similar to other pages, e.g.:
image

src/components/ReportingConfig/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+  render() {
+    const { reportingConfigs, loading, error } = this.state;
+
+    if (loading) {
+      return <LoadingMessage className="overview" />;
+    }
+
+    if (error) {
+      return <ErrorPage />;
```
tuchfarber: I think we'll want to add some status / message props to the ErrorPage

src/components/ReportingConfig/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+                <Collapsible
+                  styling="card"
+                  className="shadow"
+                  title={
+                    <div className="row justify-content-around">
+                      <Icon
+                        className={`col-1 ${config.active ? ' fa fa-check text-success-300' : ' fa fa-times text-danger-300'}`}
+                      />
+                      <div className="col">
+                        <h6>Report Type:</h6>
```
tuchfarber: Are we actually at the 6th header level, or was this done for style? Make sure we aren't skipping levels and causing a11y issues. It looks like we don't get below `h3` anywhere in the admin portal.

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+
+// for the submit button
+const SUBMIT_STATES = {
+  ERROR: 'error',
+  DEFAULT: 'default',
+  COMPLETE: 'complete',
+  PENDING: 'pending',
+};
+
+const MONTHLY_MAX = 31;
```
tuchfarber: This is probably outside the scope of this ticket, but do we actually support someone putting 31? Would they only get reports 7 times a year?

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+const MONTHLY_MIN = 1;
+
+const EMAIL_REGEX = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/igm;
```
tuchfarber: super nit: If we're doing case insensitive (`i`), we can drop the lower case letters from the beginning (`a-z`).

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+    frequency: this.props.config ? this.props.config.frequency : 'monthly',
+    deliveryMethod: this.props.config ? this.props.config.deliveryMethod : 'email',
+    invalidFields: {},
+    active: this.props.config ? this.props.config.active : false,
+    submitState: SUBMIT_STATES.DEFAULT,
+  }
+
+  /**
+   * Validates this form. If the form is valid, it will return true.
+   * Otherwise, it returns the invalid fields.
```
tuchfarber: This seems a bit confusing once you get into the code and have lines like: `if (invalidFields !== true)`. I think it'd make more sense if it always returned the invalid fields, and then the lack of invalid fields is your check that it's confirmed. 

As a general rule, having inconsistent return types can be confusing.

src/components/ReportingConfig/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+    try {
+      const response = await LMSApiService.postNewReportingConfig(transformedData);
+      this.setState(prevState => ({
+        reportingConfigs: [
+          ...prevState.reportingConfigs,
+          response.data,
+        ],
+      }));
+      this.newConfigFormRef.current.close();
+      return undefined;
```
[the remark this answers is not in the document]
tuchfarber: Not saying it's incorrect, I just don't understand the use of it, and it think it's been talked about recently (regarding `void`)?

src/components/ReportingConfig/ReportingConfigForm.jsx
tuchfarber: This doesn't have to be researched/answered for this PR. Just something we'll want to check in the future.

src/components/ReportingConfig/ReportingConfigForm.jsx
tuchfarber: If we already have a utility function (per Adam), that's probably a better solution.

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: Good point, I'm not sure if anyone would want that. No easy solution here (that I can see), but I can put a note in the code if you'd like.

src/components/ReportingConfig/index.jsx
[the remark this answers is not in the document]
Tj-Tracy: This was done because eslint wants a consistent return since it's returning an error. I don't know if this is the right choice, I'm open to suggestions (eslint doesn't like void either).

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: I've switched to the better utility function (Thanks Adam!)

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: I've changed this. Now it will just return an empty object. In turn the if statement makes more sense than before. `if (!isEmpty(invalidFields))` feels better.

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: I've split out these two separate components. The fields are going to change slightly soon, so I haven't yet split it further, but if you think there's more to split out I'm open to suggestions.

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: This is now using that function

src/components/ReportingConfig/ReportingConfigForm.jsx
Tj-Tracy: This is being changed now (We got feedback from ECS), so the fields will be changing soon to add a password to both email and sftp methods. Once that is finished, I will remove the TODO

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -0,0 +1,512 @@
...
+                  />
+                </ValidationFormGroup>
+                {/* TODO: check if the password can be blank */}
+                <ValidationFormGroup
+                  for="sftpPassword"
+                  helpText="The password to use to securely access the host. The password will be encrypted when stored in the database"
+                  invalid={invalidFields.sftpPassword}
+                  invalidMessage="Required. Password must not be blank"
+                >
+                  <label htmlFor="sftpPassword">Decrpyted SFTP Password</label>
```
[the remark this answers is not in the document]
Tj-Tracy: I changed this by removing that word entirely. Not necessary.

src/components/ReportingConfig/index.jsx
Tj-Tracy: done.

src/components/ReportingConfig/index.jsx
Tj-Tracy: I've changed these to h3 with the h6 class. If there's a better, more a11y way to do this, I'd be happy to change it.

PR #224, opened 2019-10-08
Added a notes field so that the requestor can write notes for the codes.

zamanafzal:
This ticket is related to provide a box so that the requestor can write notes & specifications for the codes they're ordering, to streamline the communication. 

**Page URL** = ` /{enterprise-slug}/admin/coupons/request`

**Pre PR: **
When the user will visit this page there will be three fields.
1.Email Address
2.Company
3.Number of codes
**Post PR: **
When the user will visit this page there will be an extra field named notes.
1.Email Address
2.Company
3.Number of codes
4.Notes
Max Length For this field is assumed 512 characters.
**Dependency:**
Backend PR for this ticket is
https://github.com/edx/edx-enterprise/pull/594

[ENT-2213](https://openedx.atlassian.net/browse/ENT-2213)

### DIFF
--- a/src/components/RequestCodesPage/RequestCodesForm.jsx
+++ b/src/components/RequestCodesPage/RequestCodesForm.jsx
@@ -7,7 +7,7 @@ import { Button, Icon } from '@edx/paragon';
 import RenderField from '../RenderField';
 import StatusAlert from '../StatusAlert';
 
-import { isRequired, isValidEmail, isValidNumber } from '../../utils';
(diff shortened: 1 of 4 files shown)


src/components/RequestCodesPage/RequestCodesForm.jsx
```diff
@@ -98,6 +98,13 @@ class RequestCodesForm extends React.Component {
                 label="Number of Codes"
                 validate={[isValidNumber]}
               />
+              <Field
+                name="notes"
```
irfanuddinahmad: messageNotes may be more descriptive ...

PR #226, opened 2019-10-24
Add old progress data type support for reporting configurations

Tj-Tracy:
Some Enterprise reporting configs have an old reporting data type that we don't want to create more of. We still need to maintain these reports, so they will still show up in the list of configs. You will not be able to change the type of the report, so if a different type of report is needed, you must create a new one.

### DIFF
--- a/src/components/ReportingConfig/ReportingConfigForm.test.jsx
+++ b/src/components/ReportingConfig/ReportingConfigForm.test.jsx
@@ -21,8 +21,8 @@ const config = {
   uuid: 'test-config-uuid',
 };
 
-const createConfig = () => {};
-const updateConfig = () => {};
+const createConfig = () => { };
+const updateConfig = () => { };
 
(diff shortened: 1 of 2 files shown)


src/components/ReportingConfig/ReportingConfigForm.test.jsx
```diff
@@ -90,4 +90,19 @@ describe('<ReportingConfigForm />', () => {
     expect(wrapper.find('input#sftpFilePath').hasClass('is-invalid')).toBeTruthy();
     expect(wrapper.find('input#encryptedSftpPassword').hasClass('is-invalid')).toBeTruthy();
   });
+  it('Does not let you select a new value for data type if it uses the old progress_v1', () => {
```
crice100: Minor: should we test that the new progress type can have its value changed?

Tj-Tracy: test added!

PR #230, opened 2019-11-06
ENT-2373 | Improve enterprise code email template messages

justinhynes:
[ENT-2373]
- Adjust wording within the code assignment, reminder, and revocation email templates

### DIFF
--- a/src/components/CodeAssignmentModal/emailTemplate.js
+++ b/src/components/CodeAssignmentModal/emailTemplate.js
@@ -1,16 +1,16 @@
 const emailTemplate = {
-  greeting: `Your learning manager has provided you with a new access code to take a course at edX.
-  `,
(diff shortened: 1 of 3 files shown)


src/components/CodeAssignmentModal/emailTemplate.js
```diff
@@ -1,16 +1,16 @@
 const emailTemplate = {
-  greeting: `Your learning manager has provided you with a new access code to take a course at edX.
```
crice100: Yay for consistency in capitalization! But should _Learning Manager_ be capitalized?

justinhynes: A good question! I had pondered this as well. In the end, I read it as a formal title so I thought it should (would?) be capitalized. I'm not set on this if you think otherwise. I just wanted to try and make it consistent.

PR #234, opened 2019-11-19
For an unavailable coupon disable `Assign` button for single and bulk code assignment

muhammad-ammar:
**Description:** Do not allow to assign codes from an unavailable coupon. To achieve this `Assign` button is disabled for single and bulk assignments. 

**Available Coupon:** `coupon start datetime < current datetime < coupon end datetime`

**JIRA:** [ENT-2467](https://openedx.atlassian.net/browse/ENT-2467)

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -135,7 +135,7 @@ class CouponDetails extends React.Component {
   getBulkActionSelectOptions() {
     const { selectedToggle, selectedCodes } = this.state;
     const {
-      couponData: { num_unassigned: unassignedCodes },
+      couponData: { num_unassigned: unassignedCodes, available: couponAvailable },
       couponDetailsTable: { data: tableData },
     } = this.props;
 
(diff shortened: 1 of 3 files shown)


muhammad-ammar:
@adamstankiewicz Can you please review the changes?

src/components/CouponDetails/index.jsx
```diff
@@ -221,7 +228,8 @@ class CouponDetails extends React.Component {
       remainingUses -= redemptions.num_assignments;
     }
 
-    return (
+    // Don't show `Assign` button for an unavailable coupon
+    return couponAvailable ? (
```
adamstankiewicz: The conditional check on `couponAvailable` can probably occur higher up in this function, e.g.:

```
getActionButton(code) {
  const {
    couponData: {
      id,
      errors,
      available: couponAvailable,
      title: couponTitle,
    },
  } = this.props;

  // ...

  let remainingUses = redemptions.total - redemptions.used;

  if (!couponAvailable) {
    return null;
  }

  // ...

  return (
    <Button /> // assign button
  );
}
```

muhammad-ammar: @adamstankiewicz As per ticket we only want to disable/hide the `Assign` button for coupons which are not available.

**Available Coupon:** `coupon start datetime < current datetime < coupon end datetime`

PR #238, opened 2019-12-04
ENT-2263 Added Search By Code Functionality

sameenfatima78:
**ENT 2263**: customers need to be able to search for codes in batches.

This pull request is an enhancement of search functionality in **Code Management Portal**. As requested in the ticket (**ENT-2263**), the enterprise admins can now make a search by code as well. The search results would help them figure out which voucher code belongs to which coupon batch and will also give them useful information about the assignment of the _searched_ voucher.

**STEPS TO TEST IT:**
- Open Code Management Portal
- Pick out any voucher from any of the coupons shown below
- Copy/Paste the code in the search bar at the top
- See results

### DIFF
--- a/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
+++ b/src/components/CodeSearchResults/CodeSearchResultsTable.jsx
@@ -68,12 +68,12 @@ const transformSearchResults = results => results.map(({
 }));
 
 const handleTableColumns = (searchQuery) => {
-  const getColumnIndexForKey = key => tableColumns.findIndex(column => column.key === key);
+  const assignedToColumnIndex = tableColumns.findIndex(column => column.key === 'assignedTo');
   // If search is made by email, no need to show "Assigned To" field
-  if (isValidEmail(searchQuery) === undefined && getColumnIndexForKey('assignedTo') > -1) {
+  if (isValidEmail(searchQuery) === undefined && assignedToColumnIndex > -1) {
     // Remove "Assigned To" column if it already exists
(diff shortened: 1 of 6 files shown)


sameenfatima78:
Hi @adamstankiewicz, I have incorporated the feedback given by you. If you're satisfied with the new changes, can you please approve this PR? Thanks.

src/components/CodeSearchResults/CodeSearchResultsTable.jsx
```diff
@@ -51,17 +52,37 @@ const transformSearchResults = results => results.map(({
...
   courseTitle,
+  assignedTo,
   redemptionDate: getFormattedDate(redemptionDate),
   isRedeemed: !!redemptionDate,
+  isAssigned: !!isAssigned,
   ...rest,
 }));
 
+const handleTableColumns = (searchQuery) => {
+  const getColumnIndexForKey = key => tableColumns.findIndex(column => column.key === key);
```
adamstankiewicz: It looks like we're only trying to find the column index of `assignedTo` here. This is a bit of nitpick but, maybe instead of a function where you can specify a `key`, we just get the index of the `assignedTo` column directly and assign it to a variable, e.g.:

```
const assignedToColumnIndex = tableColumns.findIndex(column => column.key === 'assignedTo');
```

That way, we won't need to find the column index independently in each conditional below.

sameenfatima78: Agreed. Thanks for approving the PR! I'll incorporate this new suggestion as well.

PR #246, opened 2020-01-16
ENT-2547 | Refactor services method in Frontend App Admin Portal

HammadAhmadWaqas:
refactored enterpriseDataApiServices to make it independent of the `store`.

### DIFF
--- a/src/components/Admin/index.jsx
+++ b/src/components/Admin/index.jsx
@@ -46,11 +46,12 @@ class Admin extends React.Component {
   }
 
   getMetadataForAction(actionSlug) {
+    const { enterpriseId } = this.props;
     const defaultData = {
       title: 'Full Report',
       component: <EnrollmentsTable />,
       csvFetchMethod: () => (
-        EnterpriseDataApiService.fetchCourseEnrollments({}, { csv: true })
+        EnterpriseDataApiService.fetchCourseEnrollments(enterpriseId, {}, { csv: true })
       ),
       csvButtonId: 'enrollments',
     };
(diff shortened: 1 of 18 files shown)


HammadAhmadWaqas:
Here are some [reasons](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md) listed why do we need to commit package-lock.json.

adamstankiewicz:
> @HammadAhmadWaqas overall looks good. However, I'm wondering why we need to push `package-lock.json`?

Yeah, like the link @HammadAhmadWaqas shared above mentions, `package-lock.json` should always be committed as it helps ensure everyone is guaranteed to be installing the same dependency tree.

src/components/Admin/index.jsx
```diff
@@ -45,12 +45,12 @@ class Admin extends React.Component {
     this.props.clearDashboardAnalytics();
   }
 
-  getMetadataForAction(actionSlug) {
+  getMetadataForAction(enterpriseId, actionSlug) {
```
adamstankiewicz: `enterpriseId` is already a prop on the `Admin` component. Instead of passing `enterpriseId` in as an function argument, you could retrieve it from props within this function itself:

```
const { enterpriseId } = this.props;
```

src/components/Admin/index.jsx
```diff
@@ -211,7 +235,7 @@ class Admin extends React.Component {
     return (
       <Link to={path} className="reset btn btn-sm btn-outline-primary ml-3">
         <Icon className="fa fa-undo mr-2" />
-        Reset to {this.getMetadataForAction().title}
+        Reset to {this.getMetadataForAction(enterpriseId).title}
```
adamstankiewicz: No longer need to pass `enterpriseId` as a function argument (see other comment).

HammadAhmadWaqas:
@adamstankiewicz Thank you for your review and suggestions.
I have made all the mentioned changes. Could you please review it again.

adamstankiewicz:
> @adamstankiewicz Thank you for your review and suggestions.
> I have made all the mentioned changes. Could you please review it again.

Looks good to me 👍

PR #247, opened 2020-01-23
ENT-1983 | added search bar to filter out enrollments on main learner…

HammadAhmadWaqas:
… management page.

### DIFF
--- a/src/components/Admin/index.jsx
+++ b/src/components/Admin/index.jsx
@@ -1,3 +1,4 @@
+import qs from 'query-string';
 import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
@@ -17,15 +18,24 @@ import CompletedLearnersTable from '../CompletedLearnersTable';
 import PastWeekPassedLearnersTable from '../PastWeekPassedLearnersTable';
 import LearnerActivityTable from '../LearnerActivityTable';
 
+import SearchBar from '../SearchBar';
 import AdminCards from '../../containers/AdminCards';
 import DownloadCsvButton from '../../containers/DownloadCsvButton';
 import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
 
-import { formatTimestamp } from '../../utils';
+import { formatTimestamp, updateUrl } from '../../utils';
 
 import './Admin.scss';
 
 class Admin extends React.Component {
+  constructor(props) {
+    super(props);
+    const { location } = props;
+    const queryParams = qs.parse(location.search);
+    this.state = {
+      searchQuery: queryParams.search,
+    };
+  }
   componentDidMount() {
     const { enterpriseId } = this.props;
     if (enterpriseId) {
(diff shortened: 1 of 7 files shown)


src/data/actions/enrollmentsList.js
```diff
@@ -0,0 +1,56 @@
...
+import {
+  PAGINATION_REQUEST,
+  PAGINATION_SUCCESS,
+  PAGINATION_FAILURE,
+} from '../constants/table';
+import NewRelicService from '../services/NewRelicService';
+
+const tableId = 'enrollments';
+
+const searchEnrollmentsListRequest = options => ({
```
ziafazal: Why we need this if enrollments were already show on admin portal?

HammadAhmadWaqas: we are [dispatching](https://github.com/edx/frontend-app-admin-portal/blob/master/src/data/actions/enterpriseList.js#L43) this action before the fetching search results. we can see this in [reducers](https://github.com/edx/frontend-app-admin-portal/blob/master/src/data/reducers/table.js#L24) that this is setting `loading` state to `true`.

src/components/Admin/index.jsx
```diff
@@ -199,6 +226,42 @@ class Admin extends React.Component {
     return [courseCompletions, enrolledLearners, numberOfUsers].every(item => item === 0);
   }
 
+  renderButtonsRow() {
+    if (this.displaySearchBar()) {
+      return (
+        <div className="col-12 col-md-6 text-md-right">
+          <div className="row">
+            <div className="col-sm-12 col-md-7 pr-0">
```
adamstankiewicz: Could we use the `.no-gutters` on the `.row` instead of the padding utility classes?

src/components/Admin/index.jsx
```diff
@@ -199,6 +226,42 @@ class Admin extends React.Component {
     return [courseCompletions, enrolledLearners, numberOfUsers].every(item => item === 0);
   }
 
+  renderButtonsRow() {
```
adamstankiewicz: nit: A more descriptive name may be `renderSearchBarDownloadBtnRow` or similar. Up to you if you want to change this, but search bar isn't just a button, so having it rendered in a `renderButtonsRow` function seems a bit out of place. Personal preference, though.

src/data/actions/enrollmentsList.js
```diff
@@ -0,0 +1,56 @@
...
+  payload: {
+    tableId,
+    error,
+  },
+});
+
+// This is doing nearly the same thing the table actions do however this
+// is necessary so we can pass in the `search` parameter. We leverage the same
+// events as the table actions, so the same `table` reducers will be called.
+const searchEnrollmentsList = searchOptions => (
```
adamstankiewicz: Do we need to create a new action creator function here? Could we use the already existing [`paginateTable`](https://github.com/edx/frontend-app-admin-portal/blob/master/src/data/actions/table.js#L62) action instead? Seems like we could, which would eliminate the need for the new enrollmentsList.js file.

Also, the `getPageOptionsFromUrl` util function updates in this PR below is also used in the `paginateTable` function. The update made in this PR to that util function should be enough to get the search query parameter  from the URL without needing to pass in `searchOptions` here.

HammadAhmadWaqas:
Screen Shot 2020-02-04 at 5 13 19 PM
Screen Shot 2020-02-04 at 5 15 03 PM

HammadAhmadWaqas:
@adamstankiewicz
Can you please review it again. I have incorporated all the mentioned changes.

PR #249, opened 2020-02-21
ENT-2520 : fetch custom email templates

sameenfatima78:
This PR adds the ability to fetch the custom email templates for **Assign**, **Remind**, and **Revoke** and show the active email template instead of the static email template that was shown previously. 

Ticket: https://openedx.atlassian.net/browse/ENT-2520


src/components/CodeManagement/index.jsx
```diff
@@ -49,6 +49,11 @@ class CodeManagement extends React.Component {
         state: {},
       });
     }
+
+    // To fetch email templates for assign, remind, and revoke
+    this.props.fetchEmailTemplates({
```
zamanafzal: What is the purpose of active here? Can you please add the comment in code for it.

src/data/reducers/emailTemplate.js
```diff
@@ -0,0 +1,70 @@
...
+    'email-template-closing': remindEmailTemplate.closing,
+  },
+  revoke: {
+    'email-template-greeting': revokeEmailTemplate.greeting,
+    'email-template-body': revokeEmailTemplate.body,
+    'email-template-closing': revokeEmailTemplate.closing,
+  },
+};
+
+const transformTemplates = ({ results }) => {
```
zamanafzal: I think this function should be somewhere else except reducers.

PR #251, opened 2020-03-10
ENT-2640 Add new Subscription Management tab to admin dashboard

brittneyexline:
This PR adds a new tab for Subscription Management to the admin dashboard, based on a new feature flag and the enterprise customer configuration from https://github.com/edx/edx-enterprise/pull/724. For now, the route just loads the code management screen component, which will be replaced later.

### DIFF
--- a/src/data/reducers/portalConfiguration.js
+++ b/src/data/reducers/portalConfiguration.js
@@ -13,6 +13,8 @@ const initialState = {
   enterpriseSlug: null,
   enterpriseLogo: null,
   enableCodeManagementScreen: false,
+  enableReportingConfigScreen: false,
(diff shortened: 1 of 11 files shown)


brittneyexline:
still need to write/fix tests here.

src/data/reducers/portalConfiguration.js
```diff
@@ -13,6 +13,8 @@ const initialState = {
   enterpriseSlug: null,
   enterpriseLogo: null,
   enableCodeManagementScreen: false,
+  enableReportingConfigScreen: false,
```
adamstankiewicz: 💯

PR #254, opened 2020-03-25
add subscription management UI

adamstankiewicz:
This PR adds an initial prototype for the subscription management UI.

It mock outs 3 potential API calls that may be made to the subscription/license backend service:
1. **Fetch subscription details.** This includes the start/end dates and the total number of licenses available to the plan.
2. **Fetch users associated with the subscription.** This includes a single list of users with a field for license status (e.g., "active", "assigned", "deactivated"). This call also supports supply a search query.

The search behavior is also mocked out. A search query will result in updating the user count nav links in the sidebar.

This PR does **not** account for pagination, or unit tests.

### Screenshots

image

#### With search query
image

#### Empty state
image

#### "Add users" dropdown menu
image


src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export default function LicenseStatus({ user }) {
+  const licenseStatus = useMemo(
+    () => {
+      switch (user.licenseStatus) {
+        case 'active':
```
bbaker6225: Status stuff still seems up in the air, but what you have so far around statuses seems like a good start/placeholder

src/components/subscriptions/SubscriptionData.jsx
```diff
@@ -0,0 +1,88 @@
...
+    ...getPendingUsers(),
+    ...getDeactivatedUsers(),
+  ];
+
+  return {
+    all: getAllUsers(),
+    licensed: getLicensedUsers(),
+    pending: getPendingUsers(),
+    deactivated: getDeactivatedUsers(),
+  };
```
bbaker6225: I thought we had planned for the API to share the number of users in each of these status buckets?

src/components/subscriptions/SubscriptionDetails.jsx
```diff
@@ -0,0 +1,40 @@
...
+            {moment(details.purchaseDate).format('MMMM D, YYYY')}
+          </div>
+        </div>
+        <div className="mr-5">
+          <div className="text-uppercase text-muted">
+            <small>Start Date</small>
+          </div>
+          <div className="lead">
+            {moment(details.startDate).format('MMMM D, YYYY')}
+          </div>
```
bbaker6225: Start date might no longer be a subscription level field from our Thursday discussion

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -0,0 +1,115 @@
...
+  );
+
+  const tableData = useMemo(
+    () => activeTabData.users.map(user => ({
+      emailAddress: user.emailAddress,
+      status: <LicenseStatus user={user} />,
+      actions: <LicenseActions user={user} />,
+    })),
+    [activeTabData],
+  );
```
bbaker6225: From the previous comment, might want to include start date here

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -0,0 +1,115 @@
...
+              data={tableData}
+              columns={columns}
+              className="table-striped"
+            />
+          </div>
+          <div className="mt-3 d-flex justify-content-center">
+            <Pagination
+              // eslint-disable-next-line no-console
+              onPageSelect={page => console.log(page)}
+              pageCount={Math.ceil(activeTabData.users.length / 10)}
```
bbaker6225: Not a you issue, but 10 users per page seems pretty low

adamstankiewicz: Yeah, 10 users per page was just a placeholder for now. We can decide the appropriate number of results later on as the number of results will likely be determined by the API endpoint.

src/components/subscriptions/SubscriptionData.jsx
adamstankiewicz: Are you referring to the `overview` endpoint we had discussed in the API doc?

bbaker6225: I thought part of the users endpoint was to include total counts of each status, but it could be part of the general subscription metadata endpoint. The main thing id that I thought we were planning to rely on the backend for that info

adamstankiewicz: Yes, you are correct. I think the original intent of this PR was to be an interactive prototype that would inform the next steps of how it would need to interact with the API endpoints. The fetching the total counts of each status could have been done as follow-up work.

That said, I just pushed up a commit that makes a mock API request to fetch the total counts based on the `overview` endpoint we discussed in the doc.

I made sure it gets called on initial page load and on any search or clearing of a search. The total counts of each status now comes from a mock API endpoint.

PR #257, opened 2020-04-06
increase limit on email send and email template save for assign/remind/revoke

muhammad-ammar:
**Description:** Increase limit on offer assignment email send and save template.

**JIRA:** [ENT-2708](https://openedx.atlassian.net/browse/ENT-2708)


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -19,6 +19,7 @@ class CodeAssignmentModal extends React.Component {
   constructor(props) {
     super(props);
 
+    this.maxTemplateFieldLimit = 50000;
```
zamanafzal, line 22: I think we should keep following in the utils and then import them in the components. `maxTemplateFieldLimit`, `templateErrorMessages` and 
```
Object.entries(templateErrorMessages).forEach(([key, message]) => {
      if (formData[key] && formData[key].length > this.maxTemplateFieldLimit) {
        errors[key] = message;
        errors._error.push(message);
      }
    });

```

muhammad-ammar: I thought about this but didn't found a location where I can place the util. Also I am not sure how we will handle the above mentioned `Object.entries` in the util function? Should we pass the `errors`  object to util function?

zamanafzal: Yeah, I think we should pass errors to this util function. 
Should we create a general constants file for this kind(maxTemplateFieldLimit) of entries?

muhammad-ammar: Can you suggest where we can create a utils/constants file?

zamanafzal: I think about this path
https://github.com/edx/frontend-app-admin-portal/tree/master/src/data/constants/

muhammad-ammar: But this path is for only redux action based constants.

zamanafzal: Then can we create it here.
https://github.com/edx/frontend-app-admin-portal/tree/master/src

muhammad-ammar: @zamanafzal @sameenfatima78 I have refactored the code and moved validation logic to utils. Please check.

PR #258, opened 2020-04-08
Migrate Babel to latest packages and use core-js@3 to fix IE11/Edge bugs

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-2754

1. Migrates legacy Babel packages to the latest Babel packages
2. Makes use of core-js v3 over v2 (the root cause of the IE/Edge bugs).
3. Fixes an issue with the upgrade in terms of how CSS minification is done for prod, resulting in reduced CSS file size.
4. Bumps Paragon version to hide noisy deprecation warnings during build.

These changes have been tested in IE11 and Edge via BrowserStack.


src/components/CodeReminderModal/index.jsx
```diff
@@ -389,7 +379,11 @@ CodeReminderModal.propTypes = {
   initialValues: PropTypes.shape({}).isRequired,
   isBulkRemind: PropTypes.bool,
   selectedToggle: PropTypes.string,
-  data: PropTypes.shape({}),
+  data: PropTypes.shape({
+    selectedCodes: PropTypes.arrayOf(PropTypes.shape({})),
+    code: PropTypes.string,
+    email: PropTypes.string,
+  }),
```
bbaker6225, line 386: Nit: Could make this prop shape reusable for the other modal, but I don't think it's a huge deal here

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -308,12 +308,12 @@ class ReportingConfigForm extends React.Component {
           for="pgpEncyptionKey"
           helpText="The key for encyption, if PGP encrypted file is required"
         >
-          <label htmlFor="pgpEncryptionKey">PGP Ecryption Key</label>
+          <label htmlFor="pgpEcryptionKey">PGP Ecryption Key</label>
```
bbaker6225, line 311: Is it intentionally `Ecryption` vs. `Encryption`??

src/components/EnterpriseApp/index.jsx
```diff
@@ -190,6 +191,7 @@ EnterpriseApp.propTypes = {
       enterpriseSlug: PropTypes.string.isRequired,
     }).isRequired,
   }).isRequired,
+  enterpriseId: PropTypes.string,
```
bbaker6225, line 194: Should the enable code management & subscription management screens also have default props of false (to match the sidebar)?

src/components/ReportingConfig/ReportingConfigForm.jsx
adamstankiewicz: Oops, good catch. I copy/pasted the incorrect one throughout :) Updating now!

src/components/CodeReminderModal/index.jsx
adamstankiewicz: Good suggestion, but I think I'll hold off on that for now since we're likely going to be doing some refactoring of the modal components to make them more shareable between "Code Management" and "Subscription Management".

PR #260, opened 2020-04-29
save templates with name

muhammad-ammar:
**Description:** Update single/bulk assign/remind/revoke modals with new UI. Add validations. And send the template name along with greeting and closing to backend API for save.

**JIRA:** [ENT-2807](https://openedx.atlassian.net/browse/ENT-2807)


irfanuddinahmad:
The Save Template button does not become active unless I change the Greeting or Closing. It should become active as soon as I put a name for the template.
image

muhammad-ammar:
> The Save Template button does not become active unless I change the Greeting or Closing. It should become active as soon as I put a name for the template.
> image

@irfanuddinahmad I am thinking about removing this enable/disable of Save button. New behaviour will be to enable save button at all times. Any concerns?

irfanuddinahmad:
@muhammad-ammar so if it is already saved ... it behaves like an update ... right?

muhammad-ammar:
> @muhammad-ammar so if it is already saved ... it behaves like an update ... right?

no...currently it always create a new record.

muhammad-ammar:
@irfanuddinahmad As per our discussion, I will do the below
* Remove the enable/disable of `Save Template` button
* Update `redux store` whenever we toggled between `New Email` and `From Template` buttons. This way other components can decide what to do based on the current state.

muhammad-ammar:
@irfanuddinahmad I have updated the code as discussed. You can pull the latest changes.

muhammad-ammar:
> Upon saving the template without a name, not only the modal displays an error message for template but at the top "Unable to assign codes" message is also displayed. This error message should only be linked to "Assign" button functionality. Will send a screenshot on slack for better understanding

Thanks @sameen. Great catch. I have fixed the now.

irfanuddinahmad:
If I click "New Email", then "Save Template" and save another template(with a different name) and then try to use "From Template" to use the first saved template ... Would the templates be available in the dropdown?
Also, if I save a template and then save it again with the same name ... does it update or is that covered by ENT-2826?

muhammad-ammar:
> If I click "New Email", then "Save Template" and save another template(with a different name) and then try to use "From Template" to use the first saved template ... Would the templates be available in the dropdown?
> Also, if I save a template and then save it again with the same name ... does it update or is that covered by ENT-2826?

1. Template names are not unique, so every time a new template will be created.
2. Things related to dropdown and template update will be covered in `ENT-2805` and `ENT-2826`

PR #265, opened 2020-05-13
Ent-2805 load saved email templates

zamanafzal:
Whenever any of of individual/bulk assign/remind/revoke modal is opened and user selects an existing template from the dropdown the template greeting and closing should be filled with correct text.
Also templates in the dropdown can be updated after this PR.


src/components/SaveTemplateButton/index.jsx
```diff
@@ -62,17 +63,29 @@ class SaveTemplateButton extends React.Component {
       setMode,
       saveTemplate,
       templateType,
+      emailTemplateSource,
+      emailTemplates,
```
irfanuddinahmad, line 67: Why do we need to have this here ... we should use the state instead ... i.e. we can keep a state of number of emailTemplates.

src/components/SaveTemplateButton/index.jsx
```diff
@@ -62,17 +63,29 @@ class SaveTemplateButton extends React.Component {
...
+    }
 
     const options = {
       email_type: templateType,
       email_greeting: formData['email-template-greeting'],
       email_closing: formData['email-template-closing'],
       name: formData['template-name'],
     };
+    if (!newTemplateSource) {
+      /* eslint-disable react/prop-types */
```
irfanuddinahmad, line 86: Why are we disabling?

zamanafzal:
@irfanuddinahmad In the SaveTemplateButton file I am only getting the data `emailTemplate` from the store which is needed in case of PUT Api to update the existing template when the SaveTemplateButton is pressed or raise the validation error if there is any when SaveTemplateButton is pressed.

src/components/TemplateSourceFields/index.jsx
```diff
@@ -14,7 +14,33 @@ import {
...
+  }
+
+  componentDidUpdate(prevProps) {
+    const { allEmailTemplates, emailTemplateSource } = this.props;
+    if (allEmailTemplates !== prevProps.allEmailTemplates &&
+        emailTemplateSource === EMAIL_TEMPLATE_SOURCE_FROM_TEMPLATE) {
+      this.makeOptions(this.props);
+      const emailTemplate = this.props.allEmailTemplates.filter(item =>
+        item.email_type === prevProps.emailTemplateType);
+      this.dispatchUpdatedTemplate(emailTemplate);
```
muhammad-ammar, line 42: I don't think we need to dispatch any action here.

src/components/TemplateSourceFields/index.jsx
```diff
@@ -23,27 +49,58 @@ class TemplateSourceFields extends React.Component {
...
 
   updateState(emailTemplateSource) {
-    const { setEmailTemplateSource } = this.props;
+    const {
+      setEmailTemplateSource, allEmailTemplates,
+    } = this.props;
     setEmailTemplateSource(emailTemplateSource);
+    if (emailTemplateSource === EMAIL_TEMPLATE_SOURCE_FROM_TEMPLATE) {
+      this.makeOptions(this.props);
+      this.dispatchUpdatedTemplate(allEmailTemplates);
```
muhammad-ammar, line 74: I don't think we need to dispatch any action here.

src/components/TemplateSourceFields/index.jsx
```diff
@@ -23,27 +49,58 @@ class TemplateSourceFields extends React.Component {
...
+  }
+
+  dispatchUpdatedTemplate(emailTemplate) {
+    const { saveTemplateSuccess, emailTemplateType } = this.props;
+    if (emailTemplate.length > 0) {
+      saveTemplateSuccess(emailTemplateType, emailTemplate[0]);
+    }
+  }
+
+  changeFromEmailTemplate(e) {
```
muhammad-ammar, line 97: I don't think we need this and the above method.

src/data/reducers/emailTemplate.js
```diff
@@ -68,6 +78,33 @@ const emailTemplate = (state = initialState, action) => {
...
+    }
+    case ADD_NEW_TO_ALL_TEMPLATES_SUCCESS:
+      return {
+        ...state,
+        loading: false,
+        allTemplates: [
+          ...state.allTemplates,
+          action.payload.data,
+        ],
+      };
```
muhammad-ammar, line 107: As mentioned earlier we don't need these the above 2 action/reducers.

src/data/reducers/emailTemplate.js
```diff
@@ -68,6 +78,33 @@ const emailTemplate = (state = initialState, action) => {
         error: null,
         ...transformTemplates(action.payload.data, initialState),
       };
+    case ALL_EMAIL_TEMPLATE_SUCCESS:
+      return {
+        ...state,
+        loading: false,
+        error: null,
+        allTemplates: action.payload.data,
```
muhammad-ammar, line 86: I think it would be great if we can use the existing `assign`, `remind` and `revoke` in store and make them arrays of related templates instead of introducing a new field(`allTemplates`) store

src/data/actions/emailTemplate.js
```diff
@@ -49,6 +69,9 @@ const fetchEmailTemplates = options => (
     return EcommerceApiService.fetchEmailTemplate(options)
       .then((response) => {
         dispatch(emailTemplateSuccess(response.data));
+        if (!options.active) {
+          dispatch(allEmailTemplateSuccess(response.data.results));
```
muhammad-ammar, line 73: I don't think we need to dispatch this action. Just fetch templates from backend and in reducer store them under `assign`, `remind` and `revoke` categories like I mentioned there

src/data/actions/emailTemplate.js
```diff
@@ -33,11 +43,21 @@ const saveTemplateRequest = emailType => ({
...
 
+const updateAllTemplatesSuccess = data => ({
+  type: UPDATE_ALL_TEMPLATES_SUCCESS,
+  payload: { data },
+});
+
+const AddNewToAllTemplatesSuccess = data => ({
+  type: ADD_NEW_TO_ALL_TEMPLATES_SUCCESS,
+  payload: { data },
+});
```
muhammad-ammar, line 59: I don't think we need above two methods.

src/data/reducers/emailTemplate.js
```diff
@@ -37,20 +40,27 @@ export const initialState = {
...
     'email-template-body': remindEmailTemplate.body,
     'email-template-closing': remindEmailTemplate.closing,
   },
   revoke: {
+    'template-id': 0,
+    'template-name-select': '',
     'email-template-greeting': revokeEmailTemplate.greeting,
     'email-template-body': revokeEmailTemplate.body,
     'email-template-closing': revokeEmailTemplate.closing,
   },
```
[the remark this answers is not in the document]
zamanafzal: I have implemented it with a mindset that there should be a direct objective of each object in store. Also if any change required in future, we can handle it at one place except of three.
In the case of `assign, remind and revoke` we have the values that we need to show once the template is saved and keys here matches to the redux form to initialize redux form.
In that case allTemplates will be provided to form number of times it initializes even there is no need of them.

src/data/actions/emailTemplate.js
```diff
@@ -65,6 +88,11 @@ export const saveTemplate = ({
     dispatch(saveTemplateRequest(options.email_type));
     return EcommerceApiService.saveTemplate(options)
       .then((response) => {
+        if (options.id) {
+          dispatch(updateAllTemplatesSuccess(response.data));
+        } else {
+          dispatch(AddNewToAllTemplatesSuccess(response.data));
+        }
```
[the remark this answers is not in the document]
zamanafzal: Yes that makes sense, But with this approach we will be making API hits even when we don't need it. Whenever user will update or save new template then we will be making an api call to fetch new results although we can handle it without hitting unnecessary api calls.
What we are now doing is whenever a new template is created or updated then we are adding or updating it in already fetched allTemplated array in store which is saving us API calls.

src/components/TemplateSourceFields/index.jsx
muhammad-ammar: @zamanafzal Sorry, I was mistaken. We do need this.

src/data/actions/emailTemplate.js
zamanafzal: I have mentioned the reason of the field in upper comment. Please let me know if you think otherwise.

zamanafzal:
@muhammad-ammar I have incorporated feedback. Please take a look when you find sometime.

src/data/actions/emailTemplate.js
muhammad-ammar: I still think we don't need this action. Please see my above comment.

src/components/TemplateSourceFields/index.jsx
```diff
@@ -23,27 +49,58 @@ class TemplateSourceFields extends React.Component {
...
+        }
+      });
+      this.setState({ templateOptions });
+    }
+  }
+
+  dispatchUpdatedTemplate(emailTemplate) {
+    const { saveTemplateSuccess, emailTemplateType } = this.props;
+    if (emailTemplate.length > 0) {
+      saveTemplateSuccess(emailTemplateType, emailTemplate[0]);
```
muhammad-ammar, line 93: I think it is better to have separate action fired when an option is changed in dropdown something like `CURRENT_TEMPLATE_VISIBLE` and store the related state. `saveTemplateSuccess` is an entirely different action which is not related to dropdown option change.

PR #267, opened 2020-05-18
Add Fullstory integration to Enterprise Dashboard

moconnell1453:
Add Fullstory SDK to project and use the SDK to initialize
FullStory integration for the Enterprise admin dashboard

ENT-2620


src/index.jsx
```diff
@@ -25,6 +26,8 @@ import store from './data/store';
 import history from './data/history';
 import './index.scss';
 
+FullStory.init({ orgId: 'CWDGA' });
```
adamstankiewicz, line 29: Is the `orgId` secret? Should it be in a configuration variable?

moconnell1453: I don't think it is intended to be secret. If we were using the simple Javascript snippet method of integration, the Org ID would be visible. Still, I think it's good practice to conceal it if we can. I agree we should convert this to a config variable.

PR #276, opened 2020-06-09
ENT-2837: Add customer support email link on subscription management page.

saleem-latif:
**Jira Ticket:** [ENT-2837](https://openedx.atlassian.net/browse/ENT-2837)

**Description:**
As an administrator, I want to easily reach out to ECS if I have issues performing my duties so that I know where to go to get them quickly resolved.

### DIFF
--- a/src/components/subscriptions/SubscriptionManagementPage.jsx
+++ b/src/components/subscriptions/SubscriptionManagementPage.jsx
@@ -1,5 +1,6 @@
 import React, { useState, createContext } from 'react';
 import { Helmet } from 'react-helmet';
+import { MailtoLink } from '@edx/paragon';
 
 import Hero from '../Hero';
 import SearchBar from '../SearchBar';
@@ -13,8 +14,10 @@ import StatusAlert from '../StatusAlert';
 import './styles/SubscriptionManagementPage.scss';
 import { TAB_PENDING_USERS } from './constants';
 import AddUsersButton from './AddUsersButton';
+import { configuration } from '../../config';
 
 const PAGE_TITLE = 'Subscription Management';
+const { CUSTOMER_SUPPORT_EMAIL } = configuration;
 
 export const StatusContext = createContext();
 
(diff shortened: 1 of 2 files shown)


src/components/subscriptions/SubscriptionManagementPage.jsx
```diff
@@ -54,9 +56,9 @@ export default function SubscriptionManagementPage() {
               <SubscriptionDetails />
             </div>
             <div className="col-12 col-lg-4 text-md-right">
-              <button className="btn btn-outline-primary">
-                Request assistance
-              </button>
+              <a href={`mailto: ${CUSTOMER_SUPPORT_EMAIL}`} className="btn btn-outline-primary">
```
zamanafzal, line 59: In react we use `link` at the place of `a` and `to` at the place of `href`. So this would be
<Link to={`mailto: ${CUSTOMER_SUPPORT_EMAIL}`} className="btn btn-outline-primary">
Contact Customer Support
</Link>

src/config/index.js
```diff
@@ -21,6 +21,7 @@ const configuration = {
   ACCESS_TOKEN_COOKIE_NAME: process.env.ACCESS_TOKEN_COOKIE_NAME,
   USER_INFO_COOKIE_NAME: process.env.USER_INFO_COOKIE_NAME,
   NODE_ENV: process.env.NODE_ENV,
+  CUSTOMER_SUPPORT_EMAIL: 'customersuccess@edx.org',
```
zamanafzal, line 24: Do we need to define it like other `env variables` or is it good like this?

src/components/subscriptions/SubscriptionManagementPage.jsx
saleem-latif: Thanks for the suggestion, Link is not the best solution though. I found `MailtoLink` from `@edx/paragon` that better suits this scenario. I am using that.

src/config/index.js
saleem-latif: I think @moconnell1453 can better answer that.
My understanding is that we are hardcoding the email address, so no need to expect it coming from the environment variables.

moconnell1453: I think this is acceptable, since the customer success email address is unlikely to change over time. It's hardcoded in a number of other places throughout this same repo.

PR #278, opened 2020-06-22
ENT-2967: Added assignment and reminder dates columns

irfanuddinahmad:
JIRA: https://openedx.atlassian.net/browse/ENT-2967

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -342,19 +342,29 @@ class CouponDetails extends React.Component {
 
     const getColumnIndexForKey = key => tableColumns.findIndex(column => column.key === key);
 
-    // `assigned_to` column
+    // `assigned_to, assignment_date, last_reminder_date` columns
     if (value !== 'unassigned' && getColumnIndexForKey('assigned_to') === -1) {
-      // Add `assigned_to` column if it doesn't already exist
+      // Add columns if they donot already exist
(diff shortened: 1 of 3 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -342,19 +342,29 @@ class CouponDetails extends React.Component {
...
       });
+      tableColumns.splice(tableColumns.length - 1, 0, {
+        label: 'Last Reminder Date',
+        key: 'last_reminder_date',
+      });
+      tableColumns.splice(tableColumns.length - 2, 0, {
+        label: 'Assignment Date',
+        key: 'assignment_date',
+      });
     } else if (value !== 'unassigned' && getColumnIndexForKey('assigned_to') > -1) {
```
sameenfatima78, line 360: Nit: `getColumnIndexForKey()` is being called with the same value `assigned_to` at 4 different places, can we get the index of the `assigned_to` column in a variable and use that instead of calling the method?

muhammad-ammar: good catch

irfanuddinahmad: Good observation. However, there are only about 5-6 columns with minuscule impact on performance, if any. I would keep the current code in the interest of code readability.

irfanuddinahmad:
> CodeSearchResultsTable

I dont think there is space for two more columns in that table ....
image

PR #281, opened 2020-06-24
Added Support Page

anaidu0:
## Description

Before, the support links throughout the portal were mailto links to edxcustomersuccess@gmail.com. Now, the sidebar contains a dedicated support page with a form. The submit button implemented here simply updates the page to show the success banner. The backend functionality will be implemented in a separate story.

## Support Page

Screen Shot 2020-06-24 at 11 32 50 AM

## After Submitting

Screen Shot 2020-06-24 at 11 33 55 AM


src/components/SupportPage/index.jsx
```diff
@@ -1,23 +1,68 @@
...
+        <Helmet>
+          <title>Contact Support</title>
+        </Helmet>
+        <Hero title="Contact Support" />
+        <div className="container-fluid">
+          <div className="row my-3">
+            <div className="col">
+              {features.SUPPORT && this.hasEmailAndEnterpriseName() ? (
+                <SupportForm
+                  initialValues={{
```
adamstankiewicz, line 36: Looks like you're passing in these `initialValues` (which makes sense) but not using them within `SupportForm`. I think you'll want to use the `initialValues` prop to initialize the form fields with these default values. That way, the `emailAddress` and `enterpriseName` fields will already be filled out.

src/components/SupportPage/SupportForm.jsx
```diff
@@ -0,0 +1,137 @@
...
+  invalid: PropTypes.bool.isRequired,
+  submitSucceeded: PropTypes.bool.isRequired,
+  submitFailed: PropTypes.bool.isRequired,
+  error: PropTypes.instanceOf(Error),
+
+  // custom props
+  match: PropTypes.shape({
+    url: PropTypes.string.isRequired,
+  }).isRequired,
+  initialValues: PropTypes.shape({ // eslint-disable-line react/no-unused-prop-types
```
adamstankiewicz, line 131: nit: see my other comment about the `initialValues` prop. Since we're passing this in, we should probably use the default values and then remove this `eslint-disable-line` comment.

PR #282, opened 2020-06-25
ENT-2838-update Pending Since timestamp when user is reminded for lic…

zamanafzal:
Update Time Stamp for last interaction with a particular user remind

How To Test:
When you will click on the remind button then-pending since timestamp for that particular user will be updated.
If Remind All is clicked then Pending since timestamp for all the users will be updated.

### DIFF
--- a/src/components/LicenseRemindModal/index.jsx
+++ b/src/components/LicenseRemindModal/index.jsx
@@ -9,6 +9,7 @@ import StatusAlert from '../StatusAlert';
 
 import { validateEmailTemplateFields } from '../../utils';
 import emailTemplate from './emailTemplate';
+import NewRelicService from '../../data/services/NewRelicService';
 
 
 class LicenseRemindModal extends React.Component {
@@ -78,6 +79,9 @@ class LicenseRemindModal extends React.Component {
       isBulkRemind,
       user,
       sendLicenseReminder,
+      fetchSubscriptionDetails,
+      fetchSubscriptionUsers,
+      searchQuery,
     } = this.props;
     // Validate form data
     this.validateFormData(formData);
@@ -92,11 +96,18 @@ class LicenseRemindModal extends React.Component {
       options.bulkRemind = true;
     } else if (!isBulkRemind && user) {
       options.email = user.emailAddress;
+      options.userId = user.userId;
     }
 
     return sendLicenseReminder(options)
-      .then((response) => {
-        this.props.onSuccess(response);
+      .then(async (response) => {
+        try {
+          await fetchSubscriptionDetails();
+          await fetchSubscriptionUsers({ searchQuery });
+          this.props.onSuccess(response);
+        } catch (error) {
+          NewRelicService.logAPIErrorResponse(error);
+        }
       })
       .catch((error) => {
         throw new SubmissionError({
@@ -229,6 +240,7 @@ LicenseRemindModal.defaultProps = {
   user: {},
   pendingUsersCount: 0,
   subtitle: null,
+  searchQuery: null,
 };
 LicenseRemindModal.propTypes = {
   // props From redux-form
@@ -247,8 +259,12 @@ LicenseRemindModal.propTypes = {
   isBulkRemind: PropTypes.bool,
   pendingUsersCount: PropTypes.number,
   user: PropTypes.shape({
+    userId: PropTypes.string,
     emailAddress: PropTypes.string,
   }),
+  fetchSubscriptionDetails: PropTypes.func.isRequired,
+  fetchSubscriptionUsers: PropTypes.func.isRequired,
+  searchQuery: PropTypes.string,
 };
 export default reduxForm({
   form: 'license-reminder-modal-form',
--- a/src/components/subscriptions/LicenseStatus.jsx
+++ b/src/components/subscriptions/LicenseStatus.jsx
@@ -14,7 +14,12 @@ export default function LicenseStatus({ user }) {
         case 'assigned':
           return {
             iconClassName: 'fa-hourglass-half text-muted',
-            text: 'Pending',
+            text: (
+              <React.Fragment>
+                <span>Pending</span>
+                <span className="d-block text-muted">Since {user.pendingSince.format('MMMM DD, YYYY')}</span>
+              </React.Fragment>
+            ),
           };
         case 'deactivated':
           return {
@@ -43,5 +48,6 @@ export default function LicenseStatus({ user }) {
 LicenseStatus.propTypes = {
   user: PropTypes.shape({
     licenseStatus: PropTypes.string.isRequired,
+    pendingSince: PropTypes.object,
   }).isRequired,
 };
--- a/src/components/subscriptions/data/service.js
+++ b/src/components/subscriptions/data/service.js
@@ -7,11 +7,12 @@ export function createSampleUser(licenseStatus) {
   return {
     userId: faker.random.uuid(),
     emailAddress: faker.internet.email(),
+    pendingSince: moment(faker.date.past(10)),
     licenseStatus,
   };
 }
 
-const users = [
+let users = [
   [...Array(6)].map(() => createSampleUser(ACTIVE)),
   [...Array(3)].map(() => createSampleUser(ASSIGNED)),
   [...Array(1)].map(() => createSampleUser(DEACTIVATED)),
@@ -41,6 +42,25 @@ function updateUserLicenseStatus({ userId, status }) {
     };
 
     return users[index];
+  }
+
+  return null;
+}
+
+function updateUserRemindTimeStamp({ userId, bulkRemind, pendingSince }) {
+  if (!bulkRemind) {
+    const index = users.findIndex(item => item.userId === userId);
+    if (index !== -1) {
+      users[index] = {
+        ...users[index],
+        pendingSince,
+      };
+      return users;
+    }
+  } else {
+    users = users.map(user =>
+      (user.licenseStatus === ASSIGNED ? { ...user, pendingSince } : user));
+    return users;
   }
 
   return null;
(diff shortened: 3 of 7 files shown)


src/components/LicenseRemindModal/index.jsx
```diff
@@ -91,12 +94,19 @@ class LicenseRemindModal extends React.Component {
...
-      .then((response) => {
-        this.props.onSuccess(response);
+      .then(async (response) => {
+        try {
+          await fetchSubscriptionDetails();
+          await fetchSubscriptionUsers({ searchQuery });
+          this.props.onSuccess(response);
+        } catch (error) {
+          // TODO: log error in NewRelic
+          console.log(error); // eslint-disable-line no-console
```
adamstankiewicz, line 108: nit: can replace with `NewRelicService.logAPIErrorResponse()`.

src/components/LicenseRemindModal/index.jsx
```diff
@@ -91,12 +94,19 @@ class LicenseRemindModal extends React.Component {
     if (isBulkRemind) {
       options.bulkRemind = true;
     } else if (!isBulkRemind && user) {
-      options.email = user.emailAddress;
+      options.userId = user.userId;
```
adamstankiewicz, line 97: Looks like the [`/remind` endpoint in license-manager](https://github.com/edx/license-manager/blob/142abe8de909825644bbfe613014dde7487a78c7/license_manager/apps/api/serializers.py#L88) will take a `user_email` parameter in the POST body so I think we should keep `options.email` here.

I believe the `/revoke` endpoint (currently in code review) also uses `user_email` so we should probably also pass `options.email` in the `LicenseRevokeModal`. (could be done when integrating with the APIs if you'd like)

src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -43,5 +44,6 @@ export default function LicenseStatus({ user }) {
 LicenseStatus.propTypes = {
   user: PropTypes.shape({
     licenseStatus: PropTypes.string.isRequired,
+    pendingSince: PropTypes.string.isRequired,
```
adamstankiewicz, line 47: I might recommend making this a non-required prop and render the "pending since" date conditionally, just in case the API response includes `null` on the `last_reminder_date`.

src/components/subscriptions/styles/LicenseStatus.scss
```diff
@@ -0,0 +1,9 @@
+@import "~@edx/paragon/scss/edx/utilities-only";
+
+.table-striped {
+  td {
+    p {
+      color:#ec90f5
```
adamstankiewicz, line 6: This is magenta; it should be a light gray (try $gray-300 or $gray-500, which ever is more legible). Related, would it better to target a class name, e.g. `.pending-since-date`, instead of the table elements?

nit: space between `color:` and the property value. Also, a semicolon is needed :)

src/components/subscriptions/data/service.js
```diff
@@ -130,16 +150,11 @@ export function fetchSubscriptionUsers(options = {}) {
...
+  const { userId, bulkRemind } = options;
+  const pendingSince = moment().format('MMM DD, YYYY');
+  const response = updateUserRemindTimeStamp({ userId, bulkRemind, pendingSince });
 
-/**
- * This function mocks out the response from a non-existant API endpoint. Once the endpoint
- * exists, the contents of this function will use the `apiClient` to make an actual API
- * call to get this data.
- */
-export function addLicensesForUsers(payload) {
```
adamstankiewicz, line 141: Are you intentionally removing this `addLicensesForUsers` function? Getting an error about it when running `npm run start`:

image

src/components/subscriptions/data/service.js
```diff
@@ -46,6 +47,25 @@ function updateUserLicenseStatus({ userId, status }) {
   return null;
 }
 
+function updateUserRemindTimeStamp({ userId, bulkRemind, pendingSince }) {
```
adamstankiewicz, line 50: this may need to change slightly if we no longer pass in the `userId`, but rather the user's `emailAddress`, like what the remind/revoke endpoints will be expecting.

src/components/LicenseRemindModal/index.jsx
```diff
@@ -78,6 +78,9 @@ class LicenseRemindModal extends React.Component {
       isBulkRemind,
       user,
       sendLicenseReminder,
+      fetchSubscriptionDetails,
+      fetchSubscriptionUsers,
+      searchQuery,
```
adamstankiewicz, line 83: These values are already in the `SubscriptionContext` that can be read from within any of the subscriptions component tree (`const value = useContext(SubscriptionContext)`, as opposed to needing to pass them around as props into various components 😄

src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import './styles/LicenseStatus.scss';
```
adamstankiewicz, line 4: nit: new line to separate out 3rd party packages with this styles import.

src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -14,7 +15,7 @@ export default function LicenseStatus({ user }) {
         case 'assigned':
           return {
             iconClassName: 'fa-hourglass-half text-muted',
-            text: 'Pending',
+            text: <>Pending<br /><p>Since {user.pendingSince}</p></>,  /* eslint-disable-line */
```
adamstankiewicz, line 18: Perhaps this might be a bit cleaner, and would remove the need to any custom CSS:

```
case 'assigned':
  return {
    iconClassName: 'fa-hourglass-half text-muted',
    text: (
      <React.Fragment>
        <span>Pending</span>
        <span className="d-block text-muted">Since {user.pendingSince}</span>
      </React.Fragment>
    ),
  };
```

src/components/subscriptions/data/service.js
```diff
@@ -7,11 +7,12 @@ export function createSampleUser(licenseStatus) {
   return {
     userId: faker.random.uuid(),
     emailAddress: faker.internet.email(),
+    pendingSince: moment(faker.date.past(10)).format('MMM DD, YYYY'),
```
adamstankiewicz, line 10: I'd recommend not formatting the date here, and rather where it gets read (in the table) so that the mock-API here is just returning the raw timestamp.

In terms of the format, I think it should be something like: "June 25, 2020" (full month, day with no leading zero, full year).

src/components/subscriptions/data/service.js
```diff
@@ -130,16 +150,11 @@ export function fetchSubscriptionUsers(options = {}) {
  * call to get this data.
  */
 export function sendLicenseReminder(options = {}) {
-  return Promise.resolve(options);
-}
+  const { userId, bulkRemind } = options;
+  const pendingSince = moment().format('MMM DD, YYYY');
```
adamstankiewicz, line 154: same comment here about not formatting here, but instead when reading this value within the component.

src/components/LicenseRemindModal/index.jsx
zamanafzal: Yes, I think we need to do minor changes like this once we will be integrating FE code with API.

src/components/subscriptions/styles/LicenseStatus.scss
zamanafzal: @adamstankiewicz I have followed the mockups. In the mockups it is magenta. I am attaching the screenshot for clarification.
Screenshot 2020-06-25 at 7 17 23 PM

src/components/subscriptions/data/service.js
zamanafzal: I think I have mistakenly removed it while resolving conflicts.

src/components/subscriptions/styles/LicenseStatus.scss
adamstankiewicz: Magenta text is Brandon C's way of calling out specific things in a mockup, but it is not the color the text should actually be... A magenta does not align with the color scheme on this page; it should be a gray, like the `.text-muted` class will do.

src/components/subscriptions/data/service.js
zamanafzal: The date format used here is the same as you mentioned: "June 25, 2020" (full month, day with no leading zero, full year).

src/components/subscriptions/styles/LicenseStatus.scss
zamanafzal: What is your suggestion? Should I stick with magenta or gray?

adamstankiewicz: Gray, using the `.text-muted` class.

src/components/subscriptions/data/service.js
adamstankiewicz: When I ran your branch locally, I'm seeing dates as "Oct 10, 2015". The month seems to be abbreviated.

src/components/subscriptions/data/service.js
zamanafzal: Yes, this can be changed once we have a real endpoint.

src/components/LicenseRemindModal/index.jsx
zamanafzal: While ding it this way I am getting an error
Invalid hook call. Hooks can only be called inside of the body of a function component.
I think it is because we are trying to use it in class base component.

zamanafzal:
@adamstankiewicz I have incorporated the feedback. Can you please have another look?

src/components/LicenseRemindModal/index.jsx
adamstankiewicz: Yeah, hooks can only be used in functional component. To use context in a class-based component: https://reactjs.org/docs/context.html#classcontexttype

src/components/LicenseRemindModal/index.jsx
```diff
@@ -92,11 +96,19 @@ class LicenseRemindModal extends React.Component {
...
-      .then((response) => {
-        this.props.onSuccess(response);
+      .then(async (response) => {
+        try {
+          await fetchSubscriptionDetails();
+          await fetchSubscriptionUsers({ searchQuery });
+          this.props.onSuccess(response);
+        } catch (error) {
+          NewRelicService.logAPIErrorResponse(error);
+          console.log(error); // eslint-disable-line no-console
```
adamstankiewicz, line 110: nit: can remove this console.log

PR #285, opened 2020-06-29
Admin Dash: Connect read endpoint for subscriptions frontend

saleem-latif:
**Jira Ticket:** [ENT-2952](https://openedx.atlassian.net/browse/ENT-2952)

__Description:__
Since the frontend and backend have been in development independently, we need to make sure that the endpoint loading subscriptions/license data is hooked up to the frontend.

Acceptance Criteria:

1. When loading license data in the admin dash subscriptions page, data is populated by the license manager service via API call.
2. Verify that we are passing the user’s searchQuery to the license list and license status overview endpoints.
3. Verify that we are passing the currently selected tab’s statusFilter (e.g., “active”) to the license list endpoint.
4. Verify that when calling the read endpoints, an appropriate loading state(s) are shown. We don’t have mocks for this, so use best judgement for what you think works well for the user.
5. Verify that when a request to a read endpoint fails (e.g., network request error, API throws a 500), we handle this gracefully in the UI by showing an error message.


src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -47,7 +47,6 @@ export default function LicenseStatus({ user }) {
 
 LicenseStatus.propTypes = {
   user: PropTypes.shape({
-    licenseStatus: PropTypes.string.isRequired,
-    pendingSince: PropTypes.object,
```
saleem-latif, line 51: @adamstankiewicz What license attribute will replace this?

src/components/subscriptions/SubscriptionData.jsx
```diff
@@ -55,38 +130,18 @@ export default function SubscriptionData({ children }) {
...
-          [TAB_LICENSED_USERS]: ACTIVE,
+          [TAB_LICENSED_USERS]: ACTIVATED,
           [TAB_PENDING_USERS]: ASSIGNED,
           [TAB_DEACTIVATED_USERS]: DEACTIVATED,
         };
 
         setSearchQuery(options.searchQuery);
-
-        return (
-          Promise.all([
```
adamstankiewicz, line 76: Similar question as above for the rationale for moving away from `Promise.all`.

src/components/subscriptions/LicenseAllocationNavigation.jsx
```diff
@@ -36,22 +48,37 @@ export default function LicenseAllocationNavigation() {
...
-              {tab.text}
-            </button>
-          </li>
-        ))}
-      </ul>
-    </nav>
+    <React.Fragment>
+      {isLoading && <LoadingMessage className="loading mt-3" />}
+      {
+        error && <StatusAlert
```
adamstankiewicz, line 54: As is, if both the overview API call and the list API call error, through this PR, we'd be showing 2 error status alerts? One in the sidebar and one in the table area? 

I'm wondering if it makes sense to show a single error status alert if either of those calls fail rather than showing potentially 2 alerts right next to each other, especially with one being in a small sidebar that would result in a narrow width for the status alert.

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,168 @@
...
+    count: 0,
+    next: null,
+    previous: null,
+  });
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState(null);
+  const [refreshCount, setRefreshCount] = useState(0);
+  const [pageUrl, setPageUrl] = useState(null);
+
+  const refreshSubscriptions = () => setRefreshCount(refreshCount + 1);
```
adamstankiewicz, line 20: I get the rationale for `refreshCount` (to force a re-render I'm guessing?) but it feels a bit odd to be necessary within a React hook.

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (!subscriptionUUIDValue) {
+      return;
+    }
+
+    LicenseManagerApiService.fetchSubscriptionUsersOverview(subscriptionUUIDValue)
+      .then((response) => {
+        setIsLoading(false);
+        const subscriptionUsersOverviewData = response.data.reduce((accumulator, currentValue) => ({
+          [currentValue.status]: currentValue.count, ...accumulator,
```
adamstankiewicz, line 83: Does this properly handle the case where a status may be omitted from the `/overview` response? Trying to think/understand why `undefined` is showing in the tab navigation counts. Does it need the `+` preceding `currentValue.count` like on line 86?

adamstankiewicz:
When running locally, I see the "undefined" counts and the table has a bunch of empty data. I believe these are known and documented items in a follow-up ticket, though?
 
image

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,168 @@
...
+
+export const useSubscriptions = (includeInActive = false) => {
+  const [subscriptions, setSubscriptions] = useState({
+    results: [],
+    count: 0,
+    next: null,
+    previous: null,
+  });
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState(null);
```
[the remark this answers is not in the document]
saleem-latif: I added the initialState argument because PyCharm gives a warning that `useState` expects an argument and none is provided. Also, I think it is a better idea to provide initial state.

src/components/subscriptions/hooks/licenseManagerHooks.js
saleem-latif: Yes, you are right. I have removed `refreshCount` and have also removed a lot of unnecessary code from the hooks. they look a lot cleaner now.

src/data/services/LicenseManagerApiService.js
```diff
@@ -1,3 +1,5 @@
+import qs from 'query-string';
```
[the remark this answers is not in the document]
saleem-latif: Okay, I will move the service functionality inside `src/components/subscriptions/data/service.js`

src/components/subscriptions/LicenseAllocationNavigation.jsx
saleem-latif: I have moved error messages to a single place, i.e. above the grid.

src/components/subscriptions/SubscriptionData.jsx
```diff
@@ -28,25 +36,92 @@ export default function SubscriptionData({ children }) {
...
 
+  // Perform fetch operation on the data.
+  useEffect(() => fetchSubscriptions(), []);
+  useEffect(
+    () => details?.uuid && fetchSubscriptionUsersOverview(details.uuid),
+    [details],
+  );
   useEffect(
-    () => {
-      Promise.all([
```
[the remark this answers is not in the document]
saleem-latif: I agree that `Promise.all` will give simplicity to the code. However, IMO hooks give a lot more functionality with a lot of cleaner code. With hooks monitoring request status and errors becomes very easy and it also gives a cleaner way for implementing pagination, searching, and page refreshing. 
I agree that the code I previously wrote added a lot of unneeded complexity, I have refactored the hooks and code looks a lot cleaner now. 
In the hook, I have also added the same functionality `Promise.all` would have. i.e. showing loading status until all requests are done and showing error status if any of the requests fail.

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,208 @@
...
+    () => fetchSubscriptionUsers(),
+    [subscriptions, status, search, page],
+  );
+
+  useEffect(() => {
+    if (
+      isLoadingSubscriptions || isLoadingSubscriptionUsersOverview || isLoadingSubscriptionUsers
+    ) {
+      // Keep the status as isLoading until all requests complete.
+      // setIsLoading(true);
```
adamstankiewicz, line 178: Is this unintentionally commented out?

src/data/services/LicenseManagerApiService.js
```diff
@@ -8,6 +10,33 @@ class LicenseManagerApiService {
...
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/?${qs.stringify(queryParams)}`;
+    return apiClient.get(url);
+  }
+
+  static fetchSubscriptionUsers(subscriptionUUID, options) {
+    const queryParams = {
+      ...options,
+    };
+
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/${subscriptionUUID}/licenses?${qs.stringify(queryParams)}`;
```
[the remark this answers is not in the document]
adamstankiewicz: Bumping this :) Without the `/`, the request to license-manager will first do a 302 redirect to the URL with a `/`. By adding the `/` here, the 302 redirect wouldn't happen.

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,208 @@
...
+    } else if (subscriptionsError || subscriptionUsersOverviewError || subscriptionUsersError) {
+      setIsLoading(false);
+      setErrors(prevState => ({
+        ...(prevState || {}),
+        ...(subscriptionsError ? { [SUBSCRIPTIONS]: NETWORK_ERROR_MESSAGE } : {}),
```
adamstankiewicz, line 183: Would it make sense to construct the object that this function returns outside of the function, e.g.:

```
const updatedErrors = {};
updatedErrors[SUBSCRIPTIONS] = subscriptionsError && NETWORK_ERROR_MESSAGE;
updatedErrors[SUBSCRIPTION_USERS_OVERVIEW] = subscriptionUsersOverviewError && NETWORK_ERROR_MESSAGE;

setErrors(prevState => ({
  ...(prevState || {}),
  ...updatedErrors,
});
```

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,208 @@
...
+      setIsLoading(false);
+      setData(prevState => ({
+        ...prevState,
+        subscriptions,
+        overview: subscriptionUsersOverview,
+        subscriptionUsers,
+      }));
+    }
+  }, [
+    subscriptions, isLoadingSubscriptions, subscriptionsError,
```
adamstankiewicz, line 200: Quite a long dependency list 😄 Curious if you think it makes sense to move some logic out into new variables to reduce the number of dependencies here, e.g.:

```
const isLoading = useMemo(
  () => isLoadingSubscriptions || isLoadingSubscriptionUsersOverview || isLoadingSubscriptionUsers,
 [isLoadingSubscriptions, isLoadingSubscriptionUsersOverview, isLoadingSubscriptionUsers],
);

useEffect(() => {
  if (isLoading) {
    // ...
  }
  return { //... };
}, [isLoading];
```

src/data/services/LicenseManagerApiService.js
```diff
@@ -8,6 +10,33 @@ class LicenseManagerApiService {
...
+    const queryParams = {
+      ...options,
+    };
+
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/?${qs.stringify(queryParams)}`;
+    return apiClient.get(url);
+  }
+
+  static fetchSubscriptionUsers(subscriptionUUID, options) {
+    const queryParams = {
```
adamstankiewicz, line 24: Are these `queryParams` necessary if all they're doing is copying `options`? Why not just pass `options` to `qs.stringify(options)`?

src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -47,7 +53,7 @@ export default function LicenseStatus({ user }) {
...
+    status: PropTypes.string.isRequired,
+    lastRemindDate: PropTypes.object,
```
bbaker6225, line 57: I think `lastRemindDate` will still be a string (if a datetime string)

src/components/subscriptions/SubscriptionDetails.jsx
```diff
@@ -15,23 +15,26 @@ export default function SubscriptionDetails() {
             <small>Purchase Date</small>
           </div>
           <div className="lead">
-            {moment(details.purchaseDate).format('MMMM D, YYYY')}
+            {moment(details.purchase_date)
+              .format('MMMM D, YYYY')}
```
bbaker6225, line 19: nit: Might be nice to make this string format a constant as it seems to be shared for all our dates

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -106,36 +114,53 @@ export default function TabContentTable() {
...
+            className="mt-3"
+            alertType="danger"
+            iconClassName="fa fa-times-circle"
+            title={`Unable to load data for ${title}`}
+            message={`Try refreshing your screen (${message})`}
+            key={title}
           />
+         ))
+      }
+      { !errors &&
```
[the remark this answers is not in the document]
bbaker6225: +1 on hiding the table

src/components/subscriptions/constants.js
```diff
@@ -5,6 +5,13 @@ export const TAB_DEACTIVATED_USERS = 'TAB_DEACTIVATED_USERS';
 
 export const PAGE_SIZE = 10;
 
-export const ACTIVE = 'active';
+export const ACTIVATED = 'activated';
```
bbaker6225, line 8: Nice

src/components/subscriptions/data/service.js
```diff
@@ -195,3 +112,41 @@ export function addUsers(options = {}) {
...
+  static fetchSubscriptions(options) {
+    const queryParams = {
+      ...options,
+    };
+
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/?${qs.stringify(queryParams)}/`;
+    return apiClient.get(url);
+  }
+
+  static fetchSubscriptionUsers(subscriptionUUID, options) {
```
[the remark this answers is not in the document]
saleem-latif: Strange, I am not seeing this behavior. I am seeing correct behavior when selecting different items from the left navigation.
Screenshot 2020-07-10 at 6 37 23 PM

As for the 
> Not sure why the "Status" column isn't set for my test user either?
It is being handled in a separate ticket.

src/components/subscriptions/TabContentTable.jsx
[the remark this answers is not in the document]
saleem-latif: Yes, I wasn't sure about which behavior would be most appropriate. Previously, I was conditionally rendering table on `isLoading` flag, but in this case, there was flakiness for date re-fetching (as `isLoading` is true while re-fetching subscription data). I thought it would be a better experience for the user to keep the table as is instead of hiding it momentarily and show it again while re-fetching the data.

src/components/subscriptions/hooks/licenseManagerHooks.js
saleem-latif: `updatedErrors[SUBSCRIPTIONS] = subscriptionsError && NETWORK_ERROR_MESSAGE;` causes an issue as it adds an attribute with `null` as value. However, I have moved error object construction outside of the function and tried to clean it up a bit.

src/components/subscriptions/LicenseStatus.jsx
saleem-latif: You are correct, but I convert it to an object before passing it as a prop [here](https://github.com/edx/frontend-app-admin-portal/pull/285/files#diff-406bdaef383e54ef2c022202985dfc45R92)

src/components/subscriptions/data/service.js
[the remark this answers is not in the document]
adamstankiewicz: Maybe it's a problem with my local setup... I will re-try to pull your branch (though I was on the latest push to the PR at the time of review) and license-manager and see if that changes anything.

src/components/subscriptions/SubscriptionDetails.jsx
saleem-latif: 👍  found a helper method for this [here](https://github.com/edx/frontend-app-admin-portal/blob/master/src/utils.js#L14-L19)

saleem-latif:
@adamstankiewicz @bbaker6225 I have incorporated your feedback changes, please take another look.

src/components/subscriptions/SubscriptionDetails.jsx
adamstankiewicz: Nice! Forgot that helper function existed haha

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,223 @@
...
+  );
+
+  useEffect(() => {
+    if (isInProgress) {
+      // Keep the status as isLoading until all requests complete.
+      setIsLoading(true);
+    } else if (hasErrors) {
+      setIsLoading(false);
+      const updatedErrors = {
+        ...(subscriptionsError && { [SUBSCRIPTIONS]: NETWORK_ERROR_MESSAGE }),
```
adamstankiewicz, line 191: > updatedErrors[SUBSCRIPTIONS] = subscriptionsError && NETWORK_ERROR_MESSAGE; causes an issue as it adds an attribute with null as value. However, I have moved error object construction outside of the function and tried to clean it up a bit.

what about `updatedErrors[SUBSCRIPTIONS] = subscriptionsError ? NETWORK_ERROR_MESSAGE : undefined;`?

Just trying to avoid doing a lot of array spreading here since it makes it a bit less readable.

src/components/subscriptions/SubscriptionData.jsx
```diff
@@ -28,65 +23,47 @@ export default function SubscriptionData({ children }) {
...
+      }
+      setOverview(subscriptionUsersOverviewData);
+      setUsers(subscriptionUsers);
+    }
+  }, [isLoading, errors, data]);
 
   const value = useMemo(
     () => ({
-      details,
+      details: camelCaseObject(details),
```
adamstankiewicz, line 49: I'd recommend doing the camelCase as soon as we get back the response. For example, here: https://github.com/edx/frontend-app-admin-portal/pull/285/files#diff-04a96483f6eb2c0ed6ca86461790cba5R30

Are all the API responses and their keys now camelCased (beyond just the subscription details response)?

src/components/subscriptions/TabContentTable.jsx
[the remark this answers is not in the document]
adamstankiewicz: By having the loading message appear above the table, it would make the entire table move around any time its loading, which could get annoying. I think a slightly more subtle approach would be to hide the table when loading.

That said, I agree it's not ideal in either case. I think hiding the table and showing the loading message (with a reasonable height, like 360px) is fine for now, though, and we can revisit this UX later.

src/components/subscriptions/data/service.js
```diff
@@ -1,31 +1,27 @@
...
     emailAddress: faker.internet.email(),
-    pendingSince: moment(faker.date.past(10)),
+    lastRemindDate: moment(faker.date.past(10)),
     licenseStatus,
   };
 }
 
-let users = [
-  [...Array(6)].map(() => createSampleUser(ACTIVE)),
+const users = [
```
adamstankiewicz, line 19: I'm a bit confused why we still need `users` here if we're no longer creating fake users?

src/data/services/LicenseManagerApiService.js
```diff
@@ -8,6 +10,21 @@ class LicenseManagerApiService {
...
     return apiClient.post(url, options, 'json');
   }
+
+  static fetchSubscriptions(options) {
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/?${qs.stringify(options)}`;
+    return apiClient.get(url);
+  }
+
+  static fetchSubscriptionUsers(subscriptionUUID, options) {
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/${subscriptionUUID}/licenses?${qs.stringify(options)}`;
```
adamstankiewicz, line 20: Reiterating from my earlier comment that we should have trailing slash `/` at the end of these API URLs, e.g., `/licenses/?...`.

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -38,9 +40,12 @@ export default function TabContentTable() {
     fetchSubscriptionUsers,
     fetchSubscriptionDetails,
     overview,
+    isLoading,
+    errors,
   } = useContext(SubscriptionContext);
   const { setSuccessStatus } = useContext(StatusContext);
 
+
```
adamstankiewicz, line 48: nit: unecessary new line

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -80,9 +85,11 @@ export default function TabContentTable() {
   );
 
   const tableData = useMemo(
-    () => users.results.map(user => ({
-      emailAddress: user.emailAddress,
-      status: <LicenseStatus user={user} />,
+    () => users?.results?.map(user => ({
+      emailAddress: user.user_email,
```
adamstankiewicz, line 89: Another case where the API response should be camelCased.

src/data/services/LicenseManagerApiService.js
adamstankiewicz, line 1: now that we moved this to service.js, we can delete this file.

src/components/subscriptions/data/service.js
```diff
@@ -195,3 +112,41 @@ export function addUsers(options = {}) {
...
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/?${qs.stringify(queryParams)}`;
+    return apiClient.get(url);
+  }
+
+  static fetchSubscriptionUsers(subscriptionUUID, options) {
+    const queryParams = {
+      ...options,
+    };
+
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/${subscriptionUUID}/licenses?${qs.stringify(queryParams)}`;
```
adamstankiewicz, line 138: Reiterating from my earlier comment that we should have trailing slash `/` at the end of these API URLs, e.g., `/licenses/?...`.

src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -0,0 +1,223 @@
...
+    isLoading: isLoadingSubscriptionUsers,
+    error: subscriptionUsersError,
+  } = useSubscriptionUsers({
+    subscriptionUUID: subscriptions?.results[0]?.uuid,
+    options: { status, search, page },
+  });
+
+  const isInProgress = useMemo(
+    () =>
+      isLoadingSubscriptions || isLoadingSubscriptionUsersOverview || isLoadingSubscriptionUsers,
```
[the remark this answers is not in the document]
saleem-latif: yes, otherwise line length exceeds 100 characters.

src/components/subscriptions/hooks/licenseManagerHooks.js
saleem-latif: The way it is implemented, code shows all the attributes as error messages even the ones having `undefined` or `null` values. 
One other way would be to check for the value via a guard condition but I think it would not a good solution.

src/data/services/LicenseManagerApiService.js
saleem-latif: Sorry, I removed them in the previous commit by mistake, I am adding them now.

src/components/subscriptions/data/service.js
saleem-latif: I have removed the functions that are no longer being used, but some are still in use and will be removed in the upcoming PRs. Some of the functions still in use require `users` variable.

src/components/subscriptions/hooks/licenseManagerHooks.js
adamstankiewicz: OK, don't want to be a blocker on that. Was just looking for potential ways to simplify. We can keep as is!

src/components/subscriptions/SubscriptionData.jsx
saleem-latif: Yes, I have converted API responses to came case in the other places that I found

src/components/subscriptions/hooks/licenseManagerHooks.js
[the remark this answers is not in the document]
adamstankiewicz: Oh, haha. I wonder if we should increase the 100 character limit to ~120, which is what we have in our Python code? 🤔

PR #286, opened 2020-07-02
ENT-2955-Admin Dash: Connect revoke endpoint for subscriptions frontend

zamanafzal:
When revoking licenses in the admin dash subscriptions page, the expected data is sent to the license manager service via API call.

Verify that the “Revoke” submit button in the modal has a spinner while the API call is pending, and the modal closes after successful revoke.

Verify that when the revoke endpoint errors, an error alert is shown in the “Revoke” modal.

### DIFF
--- a/src/components/LicenseRevokeModal/index.jsx
+++ b/src/components/LicenseRevokeModal/index.jsx
@@ -52,10 +52,11 @@ class LicenseRevokeModal extends React.Component {
       fetchSubscriptionDetails,
       fetchSubscriptionUsers,
       searchQuery,
+      subscriptionUUID,
     } = this.props;
-    const options = { userId: user.userId };
+    const options = { user_email: user.userEmail };
 
-    return sendLicenseRevoke(options)
+    return sendLicenseRevoke(subscriptionUUID, options)
       .then(async (response) => {
         try {
           await fetchSubscriptionDetails();
@@ -85,7 +86,7 @@ class LicenseRevokeModal extends React.Component {
         <div className="license-details mb-4">
           <React.Fragment>
             <p className="message" >Revoking a license will remove access to the subscription catalog
-              for {user.emailAddress}. To re-enable access, you can assign this user to another
+              for {user.userEmail}. To re-enable access, you can assign this user to another
               license.
             </p>
           </React.Fragment>
@@ -178,13 +179,12 @@ LicenseRevokeModal.propTypes = {
   onSuccess: PropTypes.func.isRequired,
   sendLicenseRevoke: PropTypes.func.isRequired,
   user: PropTypes.shape({
-    userId: PropTypes.string.isRequired,
-    emailAddress: PropTypes.string.isRequired,
-    licenseStatus: PropTypes.string.isRequired,
+    userEmail: PropTypes.string.isRequired,
   }).isRequired,
   fetchSubscriptionDetails: PropTypes.func.isRequired,
   fetchSubscriptionUsers: PropTypes.func.isRequired,
   searchQuery: PropTypes.string,
+  subscriptionUUID: PropTypes.string.isRequired,
 };
 
 export default reduxForm({
--- a/src/containers/LicenseRevokeModal/LicenseRevokeModal.test.jsx
+++ b/src/containers/LicenseRevokeModal/LicenseRevokeModal.test.jsx
@@ -6,16 +6,17 @@ import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import { mount } from 'enzyme';
 
-import * as licenseService from '../../components/subscriptions/data/service';
+import LicenseManagerApiService from '../../components/subscriptions/data/service';
 import LicenseRevokeModal from './index';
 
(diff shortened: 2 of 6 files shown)


src/components/LicenseRevokeModal/index.jsx
```diff
@@ -52,10 +52,11 @@ class LicenseRevokeModal extends React.Component {
       fetchSubscriptionDetails,
       fetchSubscriptionUsers,
       searchQuery,
+      subscriptionUUID,
     } = this.props;
-    const options = { userId: user.userId };
+    const options = { user_email: user.user_email };
 
-    return sendLicenseRevoke(options)
+    return sendLicenseRevoke(options, subscriptionUUID)
```
irfanuddinahmad, line 59: can we add subscriptionUUID to options?

src/containers/LicenseRevokeModal/LicenseRevokeModal.test.jsx
```diff
@@ -6,16 +6,17 @@ import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import { mount } from 'enzyme';
 
-import * as licenseService from '../../components/subscriptions/data/service';
+import LicenseManagerApiService from '../../data/services/LicenseManagerApiService';
```
irfanuddinahmad, line 9: has it been decided to shift the subscription service to the new location?

src/components/LicenseRevokeModal/index.jsx
zamanafzal: Yes, we can add it. But we don't need it at the backend. To avoid sending any extra information to backend and keep the functions same as [here](https://github.com/edx/frontend-app-admin-portal/pull/285/files) I am doing it.

src/containers/LicenseRevokeModal/LicenseRevokeModal.test.jsx
zamanafzal: We have other ApiServices in one place here. To keep it generic and easy to trace it has been moved here. Should we keep it here or move it back in subscriptions?

irfanuddinahmad: @adamstankiewicz Can you respond on the above?

src/components/LicenseRevokeModal/index.jsx
irfanuddinahmad: Could we take it out of options before calling the api?

zamanafzal: Yes, we can do that, but doesn't that will look ugly? I think if we need to pass it through options then pass it to the API aswell

src/components/LicenseRevokeModal/index.jsx
```diff
@@ -85,7 +86,7 @@ class LicenseRevokeModal extends React.Component {
         <div className="license-details mb-4">
           <React.Fragment>
             <p className="message" >Revoking a license will remove access to the subscription catalog
-              for {user.emailAddress}. To re-enable access, you can assign this user to another
+              for {user.user_email}. To re-enable access, you can assign this user to another
```
adamstankiewicz, line 89: Ideally, we would be camelCasing any API response so the object keys are not snake_case (see `camelCaseObject` in utils).

src/containers/LicenseRevokeModal/LicenseRevokeModal.test.jsx
adamstankiewicz: The organization of files/folders in the admin portal have followed the container/component structure since that was "best practice" at the time of creating the initial app. 

However, most MFEs are no longer using that structure in favor of keeping things organized by the app's functionality, not by how the implementation details of how the app is built (e.g., splitting up containers/components). 

That way, components become more like self-contained modules that contain everything they need all in one place, as opposed to having related files scattered throughout the codebase. See https://github.com/edx/frontend-app-course-authoring/blob/master/docs/decisions/0002-feature-based-application-organization.rst for example.

This is the pattern the Enterprise Learner Portal follows, and the "Subscriptions" page within the admin portal was the first attempt at starting to use this feature-based organization approach within the repo.

That said, I would prefer if we tried to keep anything subscriptions related under the "subscriptions" folder so it is self-contained. I know we've strayed from this already, though... 🤷‍♂️

src/data/services/LicenseManagerApiService.js
```diff
@@ -0,0 +1,13 @@
+import apiClient from '../apiClient';
+import { configuration } from '../../config';
+
+class LicenseManagerApiService {
+  static licenseManagerBaseUrl = `${configuration.LICENSE_MANAGER_BASE_URL}/api/v1/`;
```
adamstankiewicz, line 5: nit: remove the trailing `/` and add it below between `${licenseManagerBaseUrl}/subscriptions`.

src/components/LicenseRevokeModal/index.jsx
adamstankiewicz: Agree it probably makes sense to keep as a separate argument, rather than taking it out of `options` before the API.

PR #287, opened 2020-07-02
ENT-2953: Connect Assign Endpoint for subscription page

sameenfatima78:
- When assigning licenses in the admin dash subscriptions page, the expected data is sent to the license manager service via API call.
- The “Assign Learner(s)” button has a spinner while the API call is pending, and the modal closes after successful assignment 
- When the assign endpoint errors, an error alert is shown in the “Add Users” modal 

**TESTING INSTRUCTIONS:**

- Make sure you have a SubscriptionPlan
- To add a subscription plan visit: http://localhost:18170/admin/subscriptions/subscriptionplan/add/
- Make sure your plan has available (unassigned) licenses associated with it
- Go to http://localhost:1991/{enterprise_uuid}/admin/subscriptions and click on Add Users, enter an email and then assign
- Verify that all three points written above are working fine

**Note**: For testing, you will have to hardcode the subscription_uuid value here for now: https://github.com/edx/frontend-app-admin-portal/blob/master/src/components/subscriptions/data/service.js#L75.
This will be fixed once @saleem’s PR for ENT-2952 is merged.


config/webpack.dev.config.js
```diff
@@ -110,6 +110,7 @@ module.exports = Merge.smart(commonConfig, {
       REFRESH_ACCESS_TOKEN_ENDPOINT: 'http://localhost:18000/login',
       DATA_API_BASE_URL: 'http://localhost:8000',
       ECOMMERCE_BASE_URL: 'http://localhost:18130',
+      LICENSE_MANAGER_BASE_URL: 'http://localhost:18170',
```
adamstankiewicz, line 113: We'll also need to add `LICENSE_MANAGER_BASE_URL` to edx-internal so that stage/prod are aware of it.

sameenfatima78: Sure, I can take care of that. What's the prod/stage URL for license manager?

sameenfatima78: Here's the PR: https://github.com/edx/edx-internal/pull/2390

src/data/actions/userSubscription.js
```diff
@@ -26,13 +26,14 @@ const sendUserSubscriptionFailure = error => ({
...
-  payload,
+  options,
+  subscriptionUUID,
   onSuccess = () => {},
   onError = () => {},
 }) => (
   (dispatch) => {
     dispatch(sendSubscribeUsersRequest());
-    return addLicensesForUsersEndpoint(payload).then((response) => {
+    return LicenseManagerApiService.licenseAssign(options, subscriptionUUID).then((response) => {
```
irfanuddinahmad, line 36: Can we add the UUID in options and take it out of options here? e.g. delete options['subscriptionUUID']

sameenfatima78: `options` includes all the parameters that we are passing to the backend. That's the reason we are passing subscription UUID separately to the service.

PR #288, opened 2020-07-07
ENH: Move saml configuration from django admin to portal

taliaedX:
Description:
This adds front end support for SAML Configuration such that customers can configure their third party authentication without direct intervention by CMS.

Notes:
- I implemented delete functionality before realizing we'd like soft deletes and that we're punting it. Instead of deleting deletes (heh) I left that functionality in here, but with a bit of a feature flag in the form of a state variable "deleteEnabled"
- Upon saving new providerData I am seeing a ReactState memory warning. Something something attempting to update state on unmounted component something something memory leak. If anyone can figure out what is actually causing that I'd love to know. The few state updates should be happening prior to the component unmounting AFAICT so not clear what is going on here.

Testing Configuration:
- You should probably just use my edx-platform branch from this PR: https://github.com/edx/edx-platform/pull/24557 otherwise errors are bound to happen 
- Enable the portal saml configuration screen on an enterprise customer.
  - http://localhost:18000/admin/enterprise/enterprisecustomer/
  - Click on a customer name
  - Scroll down and check the box next to "Enable portal saml configuration screen"
  - Click save at the bottom.
- Make sure you have a user with admin dashboard or operator role access *associated* with that enterprise.
- Then on a refresh of the portal, you should see the saml configuration icon in the sidebar of the frontend portal.


package.json
```diff
@@ -9,6 +9,7 @@
     "deploy:gh-pages": "npm run build:gh-pages && gh-pages -d dist",
     "is-es5": "es-check es5 ./dist/*.js",
     "lint": "eslint --ext .js --ext .jsx .",
+    "lint:fix": "eslint --fix --ext .js --ext .jsx .",
```
taliaedX, line 12: I added this because doing `npm run lint -- --fix` doesn't work properly for some reason and I wanted the option to auto-fix linter errors.

taliaedX:
Ah this is the warning I get in the console after creating a providerData config object:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in SamlProviderDataForm (created by SamlConfiguration)
    in div (created by SamlConfiguration)
    in div (created by CollapsibleBody)
    in div (created by Transition)
    in Transition (created by CSSTransition)
    in CSSTransition (created by TransitionReplace)
    in div (created by TransitionGroup)
    in TransitionGroup (created by TransitionReplace)
    in TransitionReplace (created by CollapsibleBody)
    in CollapsibleBody (created by ForwardRef)
    in div (created by CollapsibleAdvanced)
    in CollapsibleAdvanced (created by ForwardRef)
    in ForwardRef (created by SamlConfiguration)
```

src/components/EnterpriseApp/index.jsx
```diff
@@ -10,6 +10,7 @@ import RequestCodesPage from '../../containers/RequestCodesPage';
 import Sidebar from '../../containers/Sidebar';
 import ReportingConfig from '../../components/ReportingConfig';
 import SupportPage from '../../containers/SupportPage';
+import SamlConfiguration from '../../containers/SamlConfiguration';
```
binodpant, line 13: Would Recommend: `SAMLConfiguration` instead

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+   * @param {FormData} formData
+   */
+  createProviderConfig = async (formData) => {
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('name', this.props.enterpriseName);
+    transformedData.append('slug', this.props.enterpriseSlug);
+    transformedData.append('enterprise_customer_uuid', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.postNewProviderConfig(transformedData);
+      this.setState({ providerConfig: response.data });
```
binodpant, line 56: does this need to be 
```
this.setState({ ...this.state, providerConfig: response.data})
```

to preserve the rest of the state?

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+          loading: false,
+        });
+      })
+      .catch(error => this.setState({
+        error,
+      }));
+    LmsApiService.getProviderData(this.props.enterpriseId)
+      .then((response) => {
+        this.setState({
+          providerData: response.data.results[0],
```
binodpant, line 35: would double check if the setStat calls throughout may need to preserve existing state

config/webpack.dev.config.js
```diff
@@ -121,6 +121,7 @@ module.exports = Merge.smart(commonConfig, {
         REPORTING_CONFIGURATIONS: true,
         SUBSCRIPTION_MANAGEMENT: true,
         SUPPORT: true,
+        SAML_CONFIGURATION: true,
```
binodpant, line 124: for locally viewing I cannot find the path http://localhost:1991/test-enterprise/admin/samlconfiguration I get a 404 after npm start from this branch, do I need to do any other steps to see this page?

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
```diff
@@ -0,0 +1,360 @@
...
+                pending: <Icon className="fa fa-spinner fa-spin" />,
+                complete: <Icon className="fa fa-check" />,
+                error: <Icon className="fa fa-times" />,
+              }}
+              disabledStates={[SUBMIT_STATES.PENDING]}
+              className="btn-primary ml-3 col"
+            />
+          </div>
+          <div className="col col-2">
+            {config && deleteEnabled &&
```
binodpant, line 311: Do we throw up a confirmation dialog before delete? perhaps we won't even have delete in the form anymore after some recent slack chats? or do we keep this?

src/components/SamlConfiguration/index.jsx
taliaedX: setState on a specific value doesn't change the state of the other items. You may be thinking of nested state items.
Ie something like:
```
this.state = {
toplevel: {subkey: subvalue, subkey2: subvalue2}
}
```
in which case changing one of those keys would look like:
`this.setState(prevState => ({...prevState.toplevel, subkey2: newvalue }))`

taliaedX: Another thing here is that I don't want to preserve any portion of the existing provider config. The response data gets me everything that's saved to that object in edx-platform, so I'd much rather get everything back than hope something got updated properly.

src/components/SamlConfiguration/index.jsx
taliaedX: See previous comment.

config/webpack.dev.config.js
taliaedX: Ah sorry, I keep forgetting testing notes. :( Will add them, but you need to check the box in LMS admin for a specific enterprise Customer (that you have a superuser associated with) here: http://localhost:18000/admin/enterprise/enterprisecustomer/ (open a customer > check box near the bottom for "enable portal saml configuration screen" or whatever it says exactly > save > refresh the frontend and you should see it in the sidebar

src/components/EnterpriseApp/index.jsx
taliaedX: That's generally inconsistent with Pascal's case which is used for JS class names (ie the ClassName capitalization), which recommends only using capitalized acronyms for 2 letter acronyms like "IO" though I see whoever worked on ReportingConfiguration disregarded it as well (I see a class "SFTPDeliveryMethodForm").
I know the backend is `SAMLProviderConfig`, but this isn't really meant to match, if it was JS would be written in snake case lol.

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
taliaedX: deleteEnabled is set to false, so this button is not visible. However, that is a good point about having a confirmation dialogue. Not sure if I should put it in or not. On the one hand, it could end up being a nontrivial amount of work, on the other it means if we get to the point of someone working on soft delete they theoretically shouldn't need to update the front end outside of setting deleteEnabled to true (or just removing that flag altogether really)

src/components/EnterpriseApp/index.jsx
taliaedX: I should clarify: it's fine to change it to match the formatting on the backend, it's just not consistent with the naming scheme we're using here (nor is the backend name really)

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
binodpant: yup let's hide for now and deal with the dialog if we bring it back yes

src/components/EnterpriseApp/index.jsx
binodpant: I see, I guess I have not dug much into how js naming wants acronyms to be handled, but based on your explanation this seems apt.  I would not chagne anything here then, I just read this interesting thread https://github.com/airbnb/javascript/issues/653 did not quite get what their conclusion was lol

binodpant: ok wait I did find this one, https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms but again don't need to change anything, this maybe worth a check in fedx at some point I am curious too now

src/components/SamlConfiguration/index.jsx
binodpant: yes you are correct, I misread the setState calls as if it was setStage(singleValue), but you are passing the object form so we are all set

config/webpack.dev.config.js
binodpant: I have not yet figured this out, after creaeting a saml provider config through the admin, I can't find it in the EnterpriseCustomer edit page. I do not want to be a blocker, but I think post this PR I may want to set this up together with you over zoom/slack if you can help me ... sth is off I guess

I may also need to get your edx-platform changes but this looks like a different issue

binodpant:
> Can't perform a React state update on an unmounted component

don['t know yet what in particular is leading to this in the PR but going over this https://www.debuggr.io/react-update-unmounted-component/ it seems that if this is coming from our codebase, this would happen when a component unmounts before some async fetch occurs, which can then try to re-render that component that is now unmounted.

binodpant:
One remote possibility I see is that the StatefulButton which uses the state that is updated https://github.com/edx/frontend-app-admin-portal/blob/41fadc1fc0ff3605c8c7bb066ec256a40d1404ab/src/components/SamlConfiguration/SamlProviderDataForm.jsx#L170

is it possible that this button is re-rendered due to a change in submitState, but at the time pData has already been set, and hence this component has been unmounted?

But this is only a hunch since I am still not fully sure why or how that would happen

taliaedX:
@binodpant This is actually good to bring up because I had thought that was the case and ended up commenting out a line that was updating the submit button's state on completion. However, thanks to your reminder, I a. realize I only did that for the config, not for Data, and 2. realize I left the state update commented out but present in the config class lol. So let me give that a shot. (for reference, I'm referring to line 76 in providerData and line 81 in providerConfig which looks like this: `this.setState({ submitState: SUBMIT_STATES.COMPLETE });`

taliaedX:
Yep that worked! I'm still not sure why it's considering the state call within the function to be after the component unmounts....but I don't really feel that state there is necessary.

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+        this.setState({
+          providerConfig: response.data.results[0],
+          loading: false,
+        });
+      })
+      .catch(error => this.setState({
+        error,
+      }));
+    LmsApiService.getProviderData(this.props.enterpriseId)
+      .then((response) => {
```
adamstankiewicz, line 33: Curious if you considered combining these calls with `Promise.all()` and have a single `setState` for `loading`, `providerConfig`, and `providerData`?

As is, I believe one call could resolve and set `loading: false` before the other has finished.

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('name', this.props.enterpriseName);
+    transformedData.append('slug', this.props.enterpriseSlug);
+    transformedData.append('enterprise_customer_uuid', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.postNewProviderConfig(transformedData);
+      this.setState({ providerConfig: response.data });
+      return undefined;
+    } catch (error) {
+      return error;
```
adamstankiewicz, line 59: Should we also call [`logError`](https://github.com/edx/frontend-app-admin-portal/blob/9fe269b344b1237d4b2a0e04201bbc24aad7091d/src/data/services/NewRelicService.js#L8) so the error is tracked in NewRelic? 

Also applies to the few other places where we catch/return an error.

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+                key={providerConfig.id}
+                className="mb-3"
+              >
+                <Collapsible
+                  styling="card"
+                  className="shadow"
+                  title={
+                    <div className="row justify-content-around flex-fill">
+                      <Icon
+                        className={`col-1 ${providerConfig.enabled ? ' fa fa-check text-success-300' : ' fa fa-times text-danger-300'}`}
```
adamstankiewicz, line 161: you could use the `classnames` package here, e.g.:

```

className={classNames(
  'col-1', 
  {
    'fa fa-check text-success-300': providerConfig.enabled,
    'fa fa-times text-danger-300': !providerConfig.enabled,
  },
)}
```

src/components/SamlConfiguration/index.jsx
```diff
@@ -0,0 +1,265 @@
...
+                      <div className="col">
+                        <h3 className="h6">Fetched At:</h3>
+                        <p>{providerData.fetched_at}</p>
+                      </div>
+                      <div className="col">
+                        <h3 className="h6">Entity ID:</h3>
+                        <p>{providerData.entity_id}</p>
+                      </div>
+                      <div className="col">
+                        <h3 className="h6">SSO Url:</h3>
```
adamstankiewicz, line 203: nit: "URL" (capitalized) is used elsewhere. Should this be consistent?

src/components/SamlConfiguration/index.jsx
taliaedX: Ah! I am not familiar with Promise.all(). I will look into that.

> As is, I believe one call could resolve and set loading: false before the other has finished.

I was unfortunately aware of this bit but wasn't sure how to handle it, so this seems like a good direction!

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
```diff
@@ -0,0 +1,359 @@
...
+  PENDING: 'pending',
+};
+
+
+class SamlProviderConfigForm extends React.Component {
+  state = {
+    invalidFields: {},
+    submitState: SUBMIT_STATES.DEFAULT,
+    enabled: this.props.config ? this.props.config.enabled : false,
+    syncLearnerProfileData: this.props.config &&
```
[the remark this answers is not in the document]
taliaedX: is that an actual format `this.props.config?.syncLearnerProfileData` or a typo? If that's an actual format that's awesome! I mostly just wanted to avoid errors caused by calling an attribute of an object (props.config) that didn't exist.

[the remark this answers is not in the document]
adamstankiewicz: It's a feature in the JS spec called optional chaining specifically for this purpose: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining. I recently learned about it from doing a PR review for the Markhors team and have started using it as well!

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
```diff
@@ -0,0 +1,359 @@
...
+          ...this.state.invalidFields,
+          ...invalidFields,
+        },
+        submitState: SUBMIT_STATES.default,
+      });
+      return;
+    }
+
+    if (config) {
+      // formData.append('enterprise_customer_id', config.enterpriseCustomer.uuid);
```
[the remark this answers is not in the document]
taliaedX: yeah that's unnecessary. Good catch

src/components/SamlConfiguration/SamlProviderDataForm.jsx
```diff
@@ -0,0 +1,228 @@
...
+                disabled={!(pData === undefined)}
+              />
+            </ValidationFormGroup>
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="ssoUrl"
+              helpText="The SSO (Single Sign On URL) of the provider. Example: https://samltest.id/idp/profile/SAML2/Redirect/SSO"
```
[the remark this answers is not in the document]
taliaedX: Oh yes. Derp.

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
```diff
@@ -0,0 +1,359 @@
...
+                onClick={() => this.props.deleteProviderConfig(config.id)}
+              >
+                <Icon className="fa fa-times danger" /> Delete
+              </Button>
+            }
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-3">
+            <br />
```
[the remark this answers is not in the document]
taliaedX: ah hah! Yes. I was unsure of what all the classes were for stylizing. I tried using various class names I found on other error alerts, but nothing seemed to position it nicely, hence it ended up with a rather unfortunate break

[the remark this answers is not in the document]
taliaedX: Is there a way (besides digging through stylesheets) to know what class stylizations are available and what they are typically used for (ie "this classname is useful for search bars or status alerts or buttons, etc."

src/components/SamlConfiguration/index.jsx
taliaedX: hmmm didn't know about this method.
Two questions:
1. It seems like I should be either handling the error format here or have the `logError` call in my child components that get the returned error. Is there a preferred place to put that? It seems like it would be best for most of the logic to live here. rather than in the child, but just want to be sure.
The reason I mention the error formats is because if the server does fail with a 500, the error data is ridiculously large. Django spits out a lot of unnecessary data (this is also why I use `error.message` in the case of 500s)

2. Do we want errors that are going to be seen by users to also be tracked? (It's absolutely fine either way, but in the past I've seen distinguishing from backend and frontend errors usually just either not tracked or tracked in different systems, so just confirming how we do things around here. :) )

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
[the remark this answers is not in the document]
adamstankiewicz: we use Bootstrap v4, so you can look at utility classes on their docs site. We also have the Paragon docs site, which includes available Bootstrap classes as well: https://edx.github.io/paragon/foundations/spacing

src/components/SamlConfiguration/index.jsx
taliaedX: Ah also in trying to use the logError wrapper I noticed that it seems we've not been using the logError method at least not from the other edx libraries that have defined it (such as frontend-platform/logging).

It looks like in here specifically every subcomponent of the portal is being wrapped by  this ErrorBoundary class, which has: 
```
class ErrorBoundary extends React.Component {
  componentDidCatch(error) {
    NewRelicService.logError(error);
  }
```
I see the guide on Error Boundaries suggests try/catch doesn't actually use ErrorBoundaries, but I am unsure about the then/catch flow we're using.

src/components/SamlConfiguration/SamlProviderConfigForm.jsx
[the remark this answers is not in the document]
taliaedX: Neat.

src/components/SamlConfiguration/index.jsx
adamstankiewicz: If an API call is wrapped with a try/catch, you're catching the exception before it makes it's way to the ErrorBoundary. The ErrorBoundary is to catch any exceptions/errors that are not explicitly caught (e.g., in a try/catch).

I think actually should have mentioned it's the `logAPIErrorResponse` function, not `logError`. I'd say modify either of those functions as you see fit for your use case.

For context, this repo is using an outdated implementation. There is a ticket on the backlog to refactor to use `@edx/frontend-build` and `@edx/frontend-platform` like newer MFEs use today, which would give us the ability to use `logError` from `@edx/frontend-platform/logging`.

The migration to use those ^^ packages may be non-trivial so we haven't tackled that quite yet.

PR #292, opened 2020-07-09
ENT-2954: Connect remind endpoint for subscriptions frontend

sameenfatima78:
- When reminding users in the admin dash subscriptions page, the expected data is sent to the license manager service via API call.
- The “Remind” submit button in the modal has a spinner while the API call is pending, and the modal closes after successful reminder (logic should already be in place, just need to verify).
- The “remind” endpoint errors, an error alert is shown in the “Remind” modal (logic should already be in place, just need to verify).

### DIFF
--- a/src/components/LicenseRemindModal/index.jsx
+++ b/src/components/LicenseRemindModal/index.jsx
@@ -78,38 +78,31 @@ class LicenseRemindModal extends React.Component {
     const {
       isBulkRemind,
       user,
+      subscriptionUUID,
+      searchQuery,
       sendLicenseReminder,
       fetchSubscriptionDetails,
       fetchSubscriptionUsers,
-      searchQuery,
     } = this.props;
     // Validate form data
     this.validateFormData(formData);
     // Configure the options to send to the assignment reminder API endpoint
     const options = {
-      template: formData['email-template-body'],
-      template_greeting: formData['email-template-greeting'],
-      template_closing: formData['email-template-closing'],
+      greeting: formData['email-template-greeting'],
+      closing: formData['email-template-closing'],
     };
 
-    if (isBulkRemind) {
-      options.bulkRemind = true;
-    } else if (!isBulkRemind && user) {
-      options.email = user.emailAddress;
-      options.userId = user.userId;
+    if (!isBulkRemind && user) {
+      options.user_email = user.userEmail;
     }
-
-    return sendLicenseReminder(options)
+    return sendLicenseReminder(options, subscriptionUUID, isBulkRemind)
       .then(async (response) => {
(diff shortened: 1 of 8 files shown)


src/components/subscriptions/LicenseStatus.jsx
```diff
@@ -15,7 +15,6 @@ export default function LicenseStatus({ user }) {
           };
         case 'assigned':
           return {
-            iconClassName: 'fa-hourglass-half text-muted',
```
adamstankiewicz, line 18: Why are we removing the hourglass icon here?

src/components/LicenseRemindModal/index.jsx
```diff
@@ -78,38 +78,29 @@ class LicenseRemindModal extends React.Component {
...
     // Validate form data
     this.validateFormData(formData);
     // Configure the options to send to the assignment reminder API endpoint
     const options = {
-      template: formData['email-template-body'],
-      template_greeting: formData['email-template-greeting'],
-      template_closing: formData['email-template-closing'],
+      greeting: formData['email-template-greeting'],
+      closing: formData['email-template-closing'],
+      user_email: user.userEmail,
```
adamstankiewicz, line 93: Are we prematurely adding `user_email` to the `options` here? It's only relevant when `isBulkRemind` is false. See the below removed code email is only added to options if `isBulkRemind` is false. Is `user` defined when choosing the bulk remind option?

src/components/subscriptions/data/service.js
```diff
@@ -92,6 +92,13 @@ class LicenseManagerApiService {
     return apiClient.post(url, options, 'json');
   }
 
+  static licenseRemind(options, subscriptionUUID, bulkRemind) {
+    const remindUrl = bulkRemind ? 'remind-all' : 'remind';
+
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}subscriptions/${subscriptionUUID}/licenses/${remindUrl}/`;
```
zamanafzal, line 98: Please add a slash before subscriptions.`licenseManagerBaseUrl ` has been changed now.

src/components/subscriptions/LicenseStatus.jsx
sameenfatima78: Fixed.

src/components/LicenseRemindModal/index.jsx
sameenfatima78: Fixed.

src/components/subscriptions/data/service.js
sameenfatima78: Done.

PR #296, opened 2020-07-13
ENT-3166: Support form creates Zendesk ticket

anaidu0:
## Description
Submitting the support form creates a post request to the Zendesk proxy that currently resides in edx-platform. This way, users can submit a request without exposing API secrets. The Zendesk proxy takes the fields that the user enters and creates a ticket.

- https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/zendesk_proxy

## Ticket Information
The ticket is created with the inputted form fields. The tag "enterprise_admin_support" is associated with a trigger within the Zendesk Help Desk so that any ticket with that tag is sent to the edX for Business brand. Currently, ECS processes tickets that are in the edX for Business brand in Zendesk.

## Testing Considerations
- Utilized the Zendesk sandbox workflow and observed that tickets created with the "enterprise_admin_support" tag get routed to a specific brand
- However, we want to test the creation of tickets within stage with the prod Zendesk API keys/configuration and also ask ECS if there are any additional parameters they want to be added to the tickets

### DIFF
--- a/src/components/SupportPage/SupportForm.jsx
+++ b/src/components/SupportPage/SupportForm.jsx
@@ -88,6 +88,19 @@ class SupportForm extends React.Component {
                 disabled
               />
               <Field
+                name="subject"
+                type="text"
+                component={RenderField}
+                label={
+                  <React.Fragment>
+                    Subject
+                    <span className="required">*</span>
+                  </React.Fragment>
+                }
+                validate={[isRequired]}
+                required
+              />
+              <Field
                 name="notes"
                 type="text"
                 component={TextAreaAutoSize}
--- a/src/components/SupportPage/SupportForm.jsx
+++ b/src/components/SupportPage/SupportForm.jsx
@@ -56,10 +56,10 @@ class SupportForm extends React.Component {
     return (
       <React.Fragment>
         {submitFailed && error && this.renderErrorMessage()}
-        {submitSucceeded && this.renderSuccessMessage()}
+        {submitSucceeded && this.renderSuccessMessage}
         <div className="support-form row">
           <div className="col-12 col-md-6 col-lg-4">
-            <form onSubmit={handleSubmit(this.renderSuccessMessage)}>
+            <form onSubmit={handleSubmit}>
               <Field
                 name="emailAddress"
                 type="email"
@@ -91,8 +91,14 @@ class SupportForm extends React.Component {
                 name="notes"
                 type="text"
                 component={TextAreaAutoSize}
-                label="Notes"
-                validate={[maxLength512]}
+                label={
+                  <React.Fragment>
+                    Notes
+                    <span className="required">*</span>
+                  </React.Fragment>
+                }
+                validate={[isRequired, maxLength512]}
+                required
               />
               <Button
                 type="submit"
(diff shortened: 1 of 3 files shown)


src/data/services/ZendeskApiService.js
```diff
@@ -0,0 +1,25 @@
...
+class ZendeskApiService {
+    static baseUrl = configuration.LMS_BASE_URL;
+    static createZendeskTicket(options) {
+      const url = `${ZendeskApiService.baseUrl}/zendesk_proxy/v1`;
+      const data = {
+        requester: {
+          email: options.emailAddress,
+          name: options.enterpriseName,
+        },
+        subject: 'Enterprise Admin Support',
```
bbaker6225, line 13: Should we allow admins to customize the subject (through a new form field maybe)? I think it will get confusing for ECS if they have a number of tickets that all have the same subject

src/data/services/ZendeskApiService.js
```diff
@@ -0,0 +1,25 @@
+import apiClient from '../apiClient';
+import { configuration } from '../../config';
+
+class ZendeskApiService {
+    static baseUrl = configuration.LMS_BASE_URL;
+    static createZendeskTicket(options) {
+      const url = `${ZendeskApiService.baseUrl}/zendesk_proxy/v1`;
```
bbaker6225, line 7: I would just use configuration.LMS_BASE_URL instead of re-declaring it in the class

src/data/services/ZendeskApiService.js
```diff
@@ -0,0 +1,25 @@
+import apiClient from '../apiClient';
+import { configuration } from '../../config';
+
+class ZendeskApiService {
```
bbaker6225, line 4: docstring/comment might be useful for this class

src/data/services/ZendeskApiService.js
bbaker6225: Not saying we have to do it, but something to check with ECS / Joe / @brittneyexline about

src/components/SupportPage/SupportForm.jsx
```diff
@@ -56,10 +56,10 @@ class SupportForm extends React.Component {
     return (
       <React.Fragment>
         {submitFailed && error && this.renderErrorMessage()}
-        {submitSucceeded && this.renderSuccessMessage()}
+        {submitSucceeded && this.renderSuccessMessage}
```
bbaker6225, line 59: Is renderSuccessMessage a boolean? It looks like it's a function since it's bound above

src/data/services/ZendeskApiService.js
brittneyexline: yea, that seems reasonable to me as well, to let subject be a field the user enters.

src/components/SupportPage/SupportForm.jsx
bbaker6225: Looking at the code I think you'll want to make this a function call again so that https://github.com/edx/frontend-app-admin-portal/blob/063b5298745f3b10bc0e5c179a3a8d46b367920e/src/components/SupportPage/SupportForm.jsx#L33-L44 happens again

PR #297, opened 2020-07-14
show subscriptions for correct enterprise customer

muhammad-ammar:
**Description:** Show subscriptions of the currently selected enterprise customer.

**JIRA:** [ENT-3164](https://openedx.atlassian.net/browse/ENT-3164)


muhammad-ammar:
> Have you tested the scenario when there's an enterprise for which no subscription plan exist? Although, it won't be a problem since the subscription management page will only be available to enterprises that have a subscription plan. But we should still add a check just to make sure that user never lands on a 500 server error.
> 
> Other the PR looks good!

As discussed, https://openedx.atlassian.net/browse/ENT-3066 will address the issue when there is no subscription plan exist for an enterprise.

PR #299, opened 2020-07-15
ENT-3149-Pagination support for subscription work

zamanafzal:
Make sure the Pagination of Subscriptions Page works correctly.

### DIFF
--- a/src/components/subscriptions/LicenseAllocationNavigation.jsx
+++ b/src/components/subscriptions/LicenseAllocationNavigation.jsx
@@ -8,6 +8,7 @@ import {
   TAB_LICENSED_USERS,
   TAB_PENDING_USERS,
   TAB_DEACTIVATED_USERS,
+  DEFAULT_PAGE,
 } from './constants';
 
 export default function LicenseAllocationNavigation() {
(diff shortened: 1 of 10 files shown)


src/components/subscriptions/LicenseAllocationNavigation.jsx
```diff
@@ -50,7 +52,7 @@ export default function LicenseAllocationNavigation() {
                 'btn btn-link btn-block pl-0 text-left',
                 { 'font-weight-bold': activeTab === tab.key },
               )}
-              onClick={() => setActiveTab(tab.key)}
+              onClick={() => { setActiveTab(tab.key); setCurrentPage(DEFAULT_PAGE); }}
```
zamanafzal, line 55: This has been done so that when user changes the tab then he will be shown the records of the specific tab from page no1.

adamstankiewicz: Might recommend creating a new function that does both `setActiveTab` and `setCurrentPage`, and using that function here (to avoid putting 2 function calls in a single callback).

PR #302, opened 2020-07-27
ENT-3177 Show the error message only once the api call is completed.

zamanafzal:
Admin Dash Subscriptions page briefly displays empty/error state before loading all the data


src/components/subscriptions/hooks/licenseManagerHooks.js
```diff
@@ -12,6 +12,7 @@ export const useSubscriptions = (enterpriseId) => {
     next: null,
     previous: null,
   });
+  const [isSubscription, setIsSubscription] = useState(null);
```
irfanuddinahmad, line 15: isSubscription could be hasSubscription

PR #305, opened 2020-07-27
ENH- adding country attr to SAML config form

alex-sheehan-edx:
Related [backend PR](https://github.com/edx/edx-platform/pull/24560) -- **Merged**

Saml config [ground work PR]( https://github.com/edx/frontend-app-admin-portal/commit/49e2cfd2e7da36b01db3e210a3da2ee8f6205d22)    -- **Merged** 

We want an end user to have the ability to include their IDP's country attr URN along with the rest of their SAML config data and have the country be included in the backend Django Admin config creation. The linked backend PR above has backwards compatibly added country as a new field in the model. This PR adds country as an additional text input to the config form. 

Additionally added state change tests and slightly padded out the unit tests for Saml configs.

### DIFF
--- a/src/components/SamlConfiguration/index.test.jsx
+++ b/src/components/SamlConfiguration/index.test.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import SamlConfiguration from './index';
+import LmsApiService from './../../data/services/LmsApiService';
+import { REQUIRED_DATA_FIELDS } from './SamlProviderDataForm';
+
+const formData = new FormData();
+const responseData = { data: {} };
+const oldConfig = { };
+REQUIRED_DATA_FIELDS.forEach((field) => {
+  formData.append(field, 'testdata');
+  responseData.data[field] = 'testdata';
+  oldConfig[field] = 'oldtestdata';
+});
+
+describe('<SamlConfiguration /> ', () => {
+  it('creating provider config states', () => {
+    const wrapper = shallow(<SamlConfiguration enterpriseId="testEnterpriseId" />);
+    const spyPostNewProviderConfig = jest.spyOn(LmsApiService, 'postNewProviderConfig');
+
+    spyPostNewProviderConfig.mockImplementation(() => responseData);
+    wrapper.instance().createProviderConfig(formData, () => {
+      expect(wrapper.state().providerConfig).toEqual(responseData.data);
+    });
+  });
+
+  it('updating provider config states', () => {
+    const wrapper = shallow(<SamlConfiguration enterpriseId="testEnterpriseId" />);
+    const spyPostUpdatedProviderConfig = jest.spyOn(LmsApiService, 'updateProviderConfig');
+    wrapper.setState({ providerConfig: oldConfig });
+
+    spyPostUpdatedProviderConfig.mockImplementation(() => responseData);
+    wrapper.instance().updateProviderConfig(formData, () => {
+      expect(wrapper.state().providerConfig).toEqual(responseData.data);
+    });
+  });
+
+  it('deleting provider config resets states', () => {
+    const wrapper = shallow(<SamlConfiguration enterpriseId="testEnterpriseId" />);
+    const spyPostDeletedProviderConfig = jest.spyOn(LmsApiService, 'deleteProviderConfig');
+
+    wrapper.setState({ providerConfig: oldConfig });
+    spyPostDeletedProviderConfig.mockImplementation(() => {});
+    wrapper.instance().deleteProviderConfig(formData, () => {
+      expect(wrapper.state().providerConfig).toEqual(undefined);
+    });
+  });
+
+  it('handling response errors', () => {
+    const wrapper = shallow(<SamlConfiguration enterpriseId="testEnterpriseId" />);
+
+    const errorMessage = 'test error message.';
+    const spyPostUpdatedProviderConfig = jest.spyOn(LmsApiService, 'updateProviderConfig');
+    spyPostUpdatedProviderConfig.mockImplementation(() => {
+      throw new Error(errorMessage);
+    });
+    expect(wrapper.instance().updateProviderConfig(formData)).resolves.toEqual(errorMessage);
+  });
+});
--- a/src/components/SamlConfiguration/SamlProviderConfigForm.jsx
+++ b/src/components/SamlConfiguration/SamlProviderConfigForm.jsx
@@ -283,6 +283,22 @@ class SamlProviderConfigForm extends React.Component {
             </ValidationFormGroup>
           </div>
         </div>
+        <div className="row">
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="country"
+              helpText="URN of SAML attribute containing the user's country"
+            >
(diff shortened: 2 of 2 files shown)


src/components/SamlConfiguration/index.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+import Index from './index';
+import LmsApiService from './../../data/services/LmsApiService';
+import { REQUIRED_DATA_FIELDS } from './SamlProviderDataForm';
+
+const formData = new FormData();
+REQUIRED_DATA_FIELDS.forEach(field => formData.append(field, 'testdata'));
+
+describe('<SamlConfiguration /> ', () => {
+  it('creating provider config states', () => {
+    const wrapper = shallow(<Index enterpriseId="testEnterpriseId" />);
```
taliaedX, line 13: I'm confused by using Index here instead of `<SamlConfiguration`, Not really sure how a base Index component affects tests honestly.

src/components/SamlConfiguration/index.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    wrapper.setState({ providerConfig: toBeDeletedProviderConfigData });
+    spyPostDeletedProviderConfig.mockImplementation(() => {});
+    wrapper.instance().deleteProviderConfig(formData, () => {
+      expect(wrapper.state().providerConfig).toEqual(undefined);
+    });
+  });
+
+  it('handling response errors', () => {
```
taliaedX, line 62: So this particular test isn't super helpful. It is successfully testing that a handler got called, but I think I say it's not helpful because that handler contains logic I'd definitely want to test here if we have a test at all.

There are a couple things you'd need to do to have that functioning though:
1. Mock New Relic calls (there are some examples of that in this repo actually)
2. Use `toHaveReturned()`
3. Possibly change "errorMessage" - I don't know what properties or formatting this JS Error object has.

src/components/SamlConfiguration/index.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+    const spyPostUpdatedProviderConfig = jest.spyOn(LmsApiService, 'updateProviderConfig');
+
+    const oldProviderConfigData = { data: {} };
+    const mockResponseData = { data: {} };
+
+    formData.forEach((value, key) => {
+      oldProviderConfigData.data[key] = 'old '.concat(value);
+      mockResponseData.data[key] = value;
+    });
+    wrapper.setState({ providerConfig: oldProviderConfigData });
```
taliaedX, line 38: Provider Config shouldn't be getting saved as a nested dict of `data` so using oldProviderConfigData here could introduce weird bugs or errors since we don't expect providerConfig to be structured as `{data: {field:value}}`

src/components/SamlConfiguration/index.test.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import Index from './index';
+import LmsApiService from './../../data/services/LmsApiService';
+import { REQUIRED_DATA_FIELDS } from './SamlProviderDataForm';
+
+const formData = new FormData();
+REQUIRED_DATA_FIELDS.forEach(field => formData.append(field, 'testdata'));
```
taliaedX, line 9: Can I suggest doing
```
const responseData = { data: {} };
const oldConfig = { };
REQUIRED_DATA_FIELDS.forEach((field) => {
  formData.append(field, 'testdata');
  responseData.data[field] = 'testdata');
  oldConfig[field] = 'oldtestdata');
})
```
so that you don't have to loop through formData 3+ more times?

src/components/SamlConfiguration/index.test.jsx
taliaedX: I say "could introduce bugs" but I think that's a bit strong. We are rendering a saml provider config preview and a provider data preview which if we tested here (or would do in the future) the state being `{data: {}}` would be a problem.

src/components/SamlConfiguration/index.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+  it('updating provider config states', () => {
+    const wrapper = shallow(<Index enterpriseId="testEnterpriseId" />);
+    const spyPostUpdatedProviderConfig = jest.spyOn(LmsApiService, 'updateProviderConfig');
+
+    const oldProviderConfigData = { data: {} };
+    const mockResponseData = { data: {} };
+
+    formData.forEach((value, key) => {
+      oldProviderConfigData.data[key] = 'old '.concat(value);
```
taliaedX, line 35: If you don't do the top level thing I suggest, at the very least please change this name.
We have ProviderData and ProviderConfig objects so ProviderConfigData is just goofy lol.

src/components/SamlConfiguration/index.test.jsx
alex-sheehan-edx: Ahh, no good catch. That should probably be SamlConfiguration.

src/components/SamlConfiguration/index.test.jsx
alex-sheehan-edx: Agreed. Mix up with mock responses and what's set as the state. Will change back.

src/components/SamlConfiguration/index.test.jsx
alex-sheehan-edx: By `change "errorMessage"` do you mean change the hard coded string I have in the test or go about creating 
my own exception differently.?

alex-sheehan-edx: I'm also slightly confused as to what you're saying by "use tohavereturned"

taliaedX: So like I said, ideally this test would actually test the logic of the handleErrors which centers entirely around what error status code and message format is brought back.

So changing errorMessage would be to test an API error format, which I think is Json-ish. It might work as is with throwing an Error, but unclear how a JS error here differs from a response JSON blob originating from Python.

And all of the steps below that are what would be needed to test the actual logic. We return a value from handleErrors rather than set state, which is why toHaveReturned() would be used in the assertion.

I don't think fleshing out these tests is blocking or anything and I think it's light enough weight as is to be left. I just worry about either A. someone thinking we have total coverage (when we don't) or B. someone using this as a template of a comprehensive test. I think adding all these tests was probably out of scope for this task to begin with, so I can probably just create a task to flesh out these tests.

alex-sheehan-edx: To my knowledge, `toHaveReturned()` is used to test that a mock function successfully returned, and `toHaveReturnedWith()` is used to to ensure that a mock function returned a specific value. In this test, handleErrors ins't mocked or spied on, so I'm still not sure how to use `toHaveReturned`. If we want to confirm the value returned by `handleErrors` we could possibly do something like 
```
expect(
wrapper.instance().updateProviderConfig(formData)
).resolves.toEqual(errorMessage);
```

taliaedX: Ah! My bad. I was thinking we could use it against real functions.

Your suggestion looks apt here for sure.

PR #310, opened 2020-07-29
BUG: Fix saml response display

taliaedX:
While testing changes when fixing another bug, I realized provider config was not displaying if data was 404'ing. This is due to using promise.all instead of promise.settled.

Switching to promise.settled allows us to still resolve providerConfig and providerData gets simultaneously, but ensures we can display config even if the data call fails.

### Testing notes:
- Have latest master branch of edx-platform
- Make sure LMS is running.
- Make sure you have an enterprise user associated with an enterprise customer and that user has an operator or admin enterprise system wide role configured.
- Enable the saml portal in LMS admin:
  - `localhost:18000/admin/enterprise/enterprisecustomer`
  - Click on the name of an enterprise customer (thus navigating to the enterprise customer details).
  - Scroll towards the bottom and check the box next to `Enable portal saml configuration screen`
  - Make sure to click the save button at the very bottom.
- after downloading/installing this repo, run the server with `npm start`
- Navigate to localhost:1991 (if redirected to :18000 for login, just login with a user and navigate again to locahost:1991)
- Create a provider config in the admin portal (port 1991).
- Reload the admin portal and verify you can see the new config. (Previously it would not show if provider data is not present)

### DIFF
--- a/src/components/SamlConfiguration/index.test.jsx
+++ b/src/components/SamlConfiguration/index.test.jsx
@@ -1,47 +1,78 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 
 import SamlConfiguration from './index';
 import LmsApiService from './../../data/services/LmsApiService';
 import { REQUIRED_DATA_FIELDS } from './SamlProviderDataForm';
+import { REQUIRED_CONFIG_FIELDS } from './SamlProviderConfigForm';
+
+jest.mock('./../../data/services/LmsApiService');
 
 const formData = new FormData();
-const responseData = { data: {} };
+const configResponse = {
+  data: { results: [{}] },
+};
+const validDataResponse = {
+  data: { results: [{}] },
+};
+const invalidDataResponse = {
+  data: { results: [{}] },
+};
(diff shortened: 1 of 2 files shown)


src/components/SamlConfiguration/index.test.jsx
```diff
@@ -1,47 +1,78 @@
...
+      expect(wrapper.state().providerData).toEqual(validDataResponse.data.results[0]);
+    });
+  });
+
   it('creating provider config states', () => {
     const wrapper = shallow(<SamlConfiguration enterpriseId="testEnterpriseId" />);
-    const spyPostNewProviderConfig = jest.spyOn(LmsApiService, 'postNewProviderConfig');
 
-    spyPostNewProviderConfig.mockImplementation(() => responseData);
+    LmsApiService.postNewProviderConfig.mockResolvedValue(configResponse);
```
alex-sheehan-edx, line 55: oh yeah, I forgot `mock resolved value`. 👍

taliaedX: I previously did not know about it, but it makes sense.

PR #316, opened 2020-07-31
ENT-3194: refining the "Add User" email functionality

chavesj:
* allow the submission of emails from text and csv. include invalid email addresses in the request to LM

https://openedx.atlassian.net/browse/ENT-3194

### DIFF
--- a/src/components/AddUsersModal/index.jsx
+++ b/src/components/AddUsersModal/index.jsx
@@ -11,8 +11,10 @@ import {
   validateEmailTemplateFields,
   validateEmailAddresses,
   validateEmailAddressesFields,
-  mergeErrors, camelCaseObject,
+  mergeErrors,
+  camelCaseObject,
 } from '../../utils';
+import { EMAIL_ADDRESS_TEXT_FORM_DATA, EMAIL_ADDRESS_CSV_FORM_DATA } from '../../data/constants/addUsers';
 
 class AddUsersModal extends React.Component {
   constructor(props) {
@@ -54,8 +56,8 @@ class AddUsersModal extends React.Component {
   }
 
   validateFormData(formData) {
-    const userEmailsKey = 'email-addresses';
-    const emailsCSVKey = 'csv-email-addresses';
+    const userEmailsKey = EMAIL_ADDRESS_TEXT_FORM_DATA;
+    const emailsCSVKey = EMAIL_ADDRESS_CSV_FORM_DATA;
     const emailTemplateKey = 'email-template-body';
 
     /* eslint-disable no-underscore-dangle */
@@ -106,15 +108,11 @@ class AddUsersModal extends React.Component {
     // Validate email addresses from both the text area and the CSV file and
     //   submit to the backend for emails to be sent and/or error to be displayed
     const emails = [];
-    if (formData['email-addresses'] && formData['email-addresses'].length) {
-      formData['email-addresses'].split(/\r\n|\n/).forEach((email) => {
-        emails.push(email);
-      });
+    if (formData[EMAIL_ADDRESS_TEXT_FORM_DATA] && formData[EMAIL_ADDRESS_TEXT_FORM_DATA].length) {
+      emails.push(...formData[EMAIL_ADDRESS_TEXT_FORM_DATA].split(/\r\n|\n/));
     }
-    if (formData['csv-email-addresses'] && formData['csv-email-addresses'].length) {
-      formData['csv-email-addresses'].forEach((email) => {
-        emails.push(email);
-      });
+    if (formData[EMAIL_ADDRESS_CSV_FORM_DATA] && formData[EMAIL_ADDRESS_CSV_FORM_DATA].length) {
+      emails.push(...formData[EMAIL_ADDRESS_CSV_FORM_DATA]);
     }
     options.user_emails = validateEmailAddresses(emails).validEmails;
 
--- a/src/components/AddUsersModal/index.jsx
+++ b/src/components/AddUsersModal/index.jsx
@@ -105,34 +105,24 @@ class AddUsersModal extends React.Component {
 
     // Validate email addresses from both the text area and the CSV file and
     //   submit to the backend for emails to be sent and/or error to be displayed
-    const emailData = {
-      validEmails: [],
-      invalidEmails: [],
-    };
-    if (formData['email-addresses'].length > 0) {
+    const emails = [];
+    if (formData['email-addresses'] && formData['email-addresses'].length) {
       const validatedTextAreaEmailAddressData = validateEmailAddresses((
         formData['email-addresses'].split(/\r\n|\n/)
       ));
-      emailData.validEmails = emailData.validEmails.concat((
-        validatedTextAreaEmailAddressData.validEmails
-      ));
-      emailData.invalidEmails = emailData.invalidEmails.concat((
-        validatedTextAreaEmailAddressData.invalidEmails
-      ));
+      validatedTextAreaEmailAddressData.validEmails.forEach((email) => {
+        emails.push(email);
+      });
     }
-    if (formData['csv-email-addresses'].length > 0) {
+    if (formData['csv-email-addresses'] && formData['csv-email-addresses'].length) {
       const validatedCsvEmailAddressData = validateEmailAddresses((
         formData['csv-email-addresses']
       ));
-      emailData.validEmails = emailData.validEmails.concat((
-        validatedCsvEmailAddressData.validEmails
-      ));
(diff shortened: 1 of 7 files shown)


src/components/AddUsersModal/index.jsx
```diff
@@ -102,10 +102,37 @@ class AddUsersModal extends React.Component {
       greeting: formData['email-template-greeting'],
       closing: formData['email-template-closing'],
     };
-    const hasTextAreaEmails = !!formData['email-addresses'];
-    const emails = hasTextAreaEmails ? formData['email-addresses'].split(/\r\n|\n/) : formData['csv-email-addresses'];
-    const { validEmails } = validateEmailAddresses(emails);
-    options.user_emails = validEmails;
+
```
adamstankiewicz, line 105: Could this potentially help simplify the below logic?

```
const emails = [];

if (!!formData['email-addresses']) {
  emails.push(...formData['email-addresses'].split(/\r\n|\n/));
}

if (!!formData['csv-email-addresses']) {
  emails.push(...formData['csv-email-addresses']);
}

const { validEmails } = validateEmailAddresses(emails);
```

src/containers/AddUsersModal/index.jsx
```diff
@@ -19,4 +19,4 @@ const mapDispatchToProps = dispatch => ({
   }),
 });
 
-export default connect(mapStateToProps, mapDispatchToProps)(AddUsersModel);
+export default connect(mapStateToProps, mapDispatchToProps)(AddUsersModal);
```
kellmarie, line 22: I think this rename might have to also be updated in AddUsersModal.test.jsx

src/components/AddUsersModal/index.jsx
```diff
@@ -102,10 +102,27 @@ class AddUsersModal extends React.Component {
...
+    // Validate email addresses from both the text area and the CSV file and
+    //   submit to the backend for emails to be sent and/or error to be displayed
+    const emails = [];
+    if (formData['email-addresses'] && formData['email-addresses'].length) {
+      const validatedTextAreaEmailAddressData = validateEmailAddresses((
```
adamstankiewicz, line 110: If we concat all the emails (valid or otherwise) into the `emails` array, we could call `validateEmailAddresses` just once. See https://github.com/edx/frontend-app-admin-portal/pull/316/files#r463620542.

src/components/subscriptions/SubscriptionManagementPage.jsx
```diff
@@ -86,11 +86,11 @@ function SubscriptionManagementPage({ enterpriseSlug, enterpriseId }) {
                         </div>
                         <div className="col-12 col-lg-7">
                           <AddUsersButton
-                            onSuccess={() => {
+                            onSuccess={(response) => {
```
bbaker6225, line 89: Do we want to display any of the error response from the endpoint in the admin portal?

adamstankiewicz, line 89: Can we camelCase the `response` so we can access the keys without snake case (camelCase is convention in JS), e.g. `num_already_associated` --> `numAlreadyAssociated`?

I'd recommend doing that as soon as we get the data back from the API: https://github.com/edx/frontend-app-admin-portal/blob/master/src/components/AddUsersModal/index.jsx#L113

chavesj: That's outside the scope of this ticket, for now we're going with the success message that indicates already assigned and newly assigned. Updating the UI for 400 responses would have to be a different ticket

adamstankiewicz: As is, we should be covered with displaying an error alert with the `error.message` that comes back from the API call within the `AddUsersModal` (https://github.com/edx/frontend-app-admin-portal/blob/master/src/components/AddUsersModal/index.jsx#L117). Of course, if that message could be cleaned up or more useful, it probably makes sense to do in another ticket so we can get more input from Product/UX in terms of copy, etc.

adamstankiewicz: We could also only pass `response.data` into the `onSuccess` function as well. `.data` is a thing that comes with axios requests (the way we make API calls).

Example:
```
const result = camelCaseObject(response.data);
this.props.onSuccess(result);
```

See https://github.com/edx/frontend-app-admin-portal/blob/53c218b68f3d63360973e5df2ad110d0fe7ccb4d/src/utils.js#L123 for where `camelCaseObject` is defined.

chavesj: Yeah I know it wouldn't be hard; we had a pretty lengthy conversation about error handling in addition to what is covered by the AC for this ticket and it was decided that nothing was going to be done in this ticket with errors, only the success message. I'm not expanding the scope of this ticket to include something we directly talked about when grooming it

src/components/AddUsersModal/index.jsx
```diff
@@ -102,15 +102,27 @@ class AddUsersModal extends React.Component {
...
+      formData['email-addresses'].split(/\r\n|\n/).forEach((email) => {
+        emails.push(email);
+      });
+    }
+    if (formData['csv-email-addresses'] && formData['csv-email-addresses'].length) {
+      formData['csv-email-addresses'].forEach((email) => {
+        emails.push(email);
+      });
+    }
+    options.user_emails = validateEmailAddresses(emails).validEmails;
```
adamstankiewicz, line 119: tiny nit. it usually feels a bit odd to access an object property off a function directly. might recommend creating a variable for it.

```
const { validEmails } = validateEmailAddresses(emails);
options.user_emails = validEmails;
```

src/components/AddUsersModal/index.jsx
```diff
@@ -11,7 +11,7 @@ import {
   validateEmailTemplateFields,
   validateEmailAddresses,
   validateEmailAddressesFields,
-  mergeErrors,
+  mergeErrors, camelCaseObject,
```
adamstankiewicz, line 14: nit: new line per import

src/components/AddUsersModal/index.jsx
```diff
@@ -102,15 +102,27 @@ class AddUsersModal extends React.Component {
...
-    const hasTextAreaEmails = !!formData['email-addresses'];
-    const emails = hasTextAreaEmails ? formData['email-addresses'].split(/\r\n|\n/) : formData['csv-email-addresses'];
-    const { validEmails } = validateEmailAddresses(emails);
-    options.user_emails = validEmails;
+
+    // Validate email addresses from both the text area and the CSV file and
+    //   submit to the backend for emails to be sent and/or error to be displayed
+    const emails = [];
+    if (formData['email-addresses'] && formData['email-addresses'].length) {
+      formData['email-addresses'].split(/\r\n|\n/).forEach((email) => {
```
adamstankiewicz, line 110: tiny nit: could use the `emails.push(...formData['email-addresses'].split(/\r\n|\n/))` syntax here (and below) to save a few lines and avoid the `forEach` 🤷‍♂️

src/components/AddUsersModal/index.jsx
```diff
@@ -102,15 +102,27 @@ class AddUsersModal extends React.Component {
...
     };
-    const hasTextAreaEmails = !!formData['email-addresses'];
-    const emails = hasTextAreaEmails ? formData['email-addresses'].split(/\r\n|\n/) : formData['csv-email-addresses'];
-    const { validEmails } = validateEmailAddresses(emails);
-    options.user_emails = validEmails;
+
+    // Validate email addresses from both the text area and the CSV file and
+    //   submit to the backend for emails to be sent and/or error to be displayed
+    const emails = [];
+    if (formData['email-addresses'] && formData['email-addresses'].length) {
```
bbaker6225, line 109: Should the field names here (at least email-addresses and csv-email-addresses) be constants now that we're accessing them a fair bit more?

src/utils.js
```diff
@@ -9,6 +9,7 @@ import isEmpty from 'validator/lib/isEmpty';
 import isNumeric from 'validator/lib/isNumeric';
 
 import { EMAIL_TEMPLATE_FIELD_MAX_LIMIT, OFFER_ASSIGNMENT_EMAIL_SUBJECT_LIMIT } from '../src/data/constants/emailTemplate';
+import { EMAIL_ADDRESS_TEXT_FORM_DATA, EMAIL_ADDRESS_CSV_FORM_DATA } from '../src/data/constants/addUsers';
```
adamstankiewicz, line 12: nit: we're already in the `src` directory. these imports could be `from './data/constants/*`

PR #317, opened 2020-08-03
Reset form and helptext

raq929:
This adds a button to reset the filters form when filters are present ENT-3233
It also adds helptext icon with a tooltip to the Filter by Course Date label ENT-3235

I created a reusable component for the icon with a tooltip as I expect we'll need something similar again.
The looks of the icon is slightly different based on the free icons that were available from fontawesome-react

UX tickets: https://openedx.atlassian.net/browse/UXENT-6
https://openedx.atlassian.net/browse/UXENT-7

Full screen:
Screen Shot 2020-08-04 at 9 07 10 AM

Mobile/small screen:
Screen Shot 2020-08-04 at 9 09 17 AM


src/components/Admin/index.jsx
```diff
@@ -297,10 +311,21 @@ class Admin extends React.Component {
...
+                        <H2 className="table-title">{tableMetadata.title}</H2>
+                      </div>
+                      <div className="col-12 col-md-9 col-xl-10 mb-2 mb-md-0 mt-0 mt-md-1">
+                        {actionSlug && this.renderUrlResetButton()}
+                        {filtersActive && this.renderFiltersResetButton()}
+                      </div>
+                    </div>
+                    <div className="row">
+                      <div className="col">
+                        {tableMetadata.subtitle && <H3>{tableMetadata.subtitle}</H3>}
```
binodpant, line 325: maybe h3 instead of H3?

src/hooks/useWindowSize.jsx
```diff
@@ -0,0 +1,35 @@
+import { useState, useEffect } from 'react';
+
+// Hook
```
binodpant, line 3: this line seems not too useful, maybe instead a jsdoc doc like `/** kasjdf;asjdf. */` ? or just remove it since you have doc below (I would still convert that doc to jsdoc like while you are here)

src/components/IconWithTooltip/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+  altText: PropTypes.string.isRequired,
+  tooltipText: PropTypes.string.isRequired,
+  // These props have defaults above
+  /* eslint-disable react/require-default-props */
+  placementSm: PropTypes.oneOf(['top', 'right', 'left', 'bottom']),
+  placementLg: PropTypes.oneOf(['top', 'right', 'left', 'bottom']),
+  trigger: PropTypes.oneOfType(PropTypes.string, PropTypes.array),
+  breakpoint: PropTypes.number,
+  iconClassNames: PropTypes.string,
+  /* eslint-enable react/require-default-props */
```
binodpant, line 38: not sure what this is and why we need it?

src/components/IconWithTooltip/index.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { OverlayTrigger, Tooltip } from '@edx/paragon';
+import useWindowSize from '../../hooks/useWindowSize';
+
+
+const IconWithTooltip = ({
+  icon, altText, tooltipText, placementSm = 'right', placementLg = 'top', trigger = ['hover', 'focus'], breakpoint = 768, iconClassNames = 'ml-1',
```
binodpant, line 9: not sure if eslint cares, but perhaps better to have one prop per line?

src/hooks/useWindowSize.jsx
```diff
@@ -0,0 +1,35 @@
+import { useState, useEffect } from 'react';
```
binodpant, line 1: This hook looks very generic. Post this work, I would love it if we proposed this for a more generic place for commonly usable react hooks?

src/components/Admin/index.jsx
```diff
@@ -269,6 +281,8 @@ class Admin extends React.Component {
     } = this.props;
 
     const { params: { actionSlug } } = match;
+    const { location } = this.props;
+    const filtersActive = location.search;
```
binodpant, line 285: should this check be more specfic? meaning if there is anything in the url does that classify right now as filtersActive?

binodpant:
Generally small thoughts I added above.

On testing, Would be great to explore some kind of specific testing coverage for this: enzyme, react testing library or otherwise (but tbh rtl has worked the best for me, have not checked what this repo uses)

src/components/Admin/index.jsx
raq929: I think, because it worked, that it must be a Bootstrap thing. But I'll look into it.

src/hooks/useWindowSize.jsx
raq929: Oh yeah, I totally stole this hook from the internet, I'll remove it. Not sure what you mean by a jsdoc?

src/components/IconWithTooltip/index.jsx
raq929: So.... the eslint settings we have aren't quite right for function components. These props *have* defaults, in the code above, but I'm not using `defaultPropTypes`. Mostly because I think it's easier to read with the default where it currently is. We could turn the eslint setting off for function components; but that's not quite what I want to do either; I *want* these to have defaults; I just want them in a different location.

src/components/IconWithTooltip/index.jsx
raq929: Eslint does not care, but yeah, one per line can be nice. 👍

src/hooks/useWindowSize.jsx
raq929: AGREE. I put it in the most generic place I could for this repo. I'm going to move the whole icon component into paragon, so this hook will also go into paragon, so from there I think it's just a question of documenting it?

src/components/Admin/index.jsx
raq929: Soooo... I debated this one. Right now, we don't need it to be more specific; there's nothing else on this page that uses queries. In the future, we might benefit from it being more specific, or we might want to move this functionality out of the querystring altogether. (I'm in favor of the latter, tbh.)

raq929:
Agreed, I haven't done a lot of testing on this repo because it doesn't already have a lot of tests, but... that's not a great excuse. Do you have ideas for testing hooks and / or components that use hooks? I'm kind of looking for a good idea about how to test that the tooltip is in a certain location when it's a large screen and a different location when it's a smaller screen.

src/components/Admin/index.jsx
raq929: It's not bootstrap, it's specific to this repo. Maybe it has styles attached? Either way, I think it's the right component to use here.

src/components/Admin/index.jsx
binodpant: totally! That is in fact what I thought too, could we not achieve with by updating state instead? but perhaps the url is also a requirement if you want to send links etc .. so still keep that maybe? not sure at all, we can talk but does not block this pr I think

src/components/Admin/index.jsx
binodpant: ah I was confused then, this is not a regular h tag I see now!!! weird .. but okay.

src/components/IconWithTooltip/index.jsx
binodpant: for me prettier usually takes care of this

src/hooks/useWindowSize.jsx
binodpant: yup! paragaon may be the correct place perhaps, but let's check with fedx too yeah

src/hooks/useWindowSize.jsx
binodpant: oh for comments, it's typically awesome if we can use the jsdoc style 

e.g. stolen from https://jsdoc.app/about-getting-started.html :)

```
Use a JSDoc tag to describe your code
/**
 * Represents a book.
 * @constructor
 */
function Book(title, author) {
}
```

src/components/IconWithTooltip/index.jsx
binodpant: ah ok just never used this rule so was curious what it was about. looks like you got it! just maybe want to add a line on why this was added would be cool

src/components/Admin/index.jsx
binodpant: yeah resolving this one, since we discussed in testing too. right now urls seems like sufficient to do this, perhaps a rework using state + url in the future but not too worried about it right now

src/hooks/useWindowSize.jsx
raq929: I'm going to moving all of this to Paragon, so I'll make sure I do that when it's ported.

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
```diff
@@ -123,6 +125,7 @@ describe('EnterpriseApp', () => {
 
     // ensure focus is set on content wrapper
     expect(document.activeElement.className).toEqual('content-wrapper');
+    global.innerWidth = initialWidth;
```
binodpant, line 128: don't have context on this setting here, but in case this line does not execute and somehow affects other tests, is it better in a teardown or after method?

src/components/IconWithTooltip/IconWithTooltip.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+  });
+  [
+    { windowSize: 800, expectedLocation: 'right' },
+    { windowSize: 700, expectedLocation: 'top' },
+  ].forEach((data) => {
+    it(`renders the tooltip on the ${data.expectedLocation} for ${data.windowSize}px screen`, () => {
+      useWindowSize.mockReturnValue({ width: data.windowSize });
+      const { container } = render(<IconWithTooltip {...DEFAULT_PROPS} />);
+      const icon = container.querySelector(`[data-icon=${faInfoCircle.iconName}]`);
+      expect(icon).toBeTruthy();
```
binodpant, line 33: I think using the toBeInTheDocument() or toBeInTheDom() may be stronger tests in case these values are truthy but still somehow `invalid`

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
raq929: I didn't write this test; I just added this because it didn't looks like this was ever reset. I'm going to call it better than it used to be. ;)

src/components/IconWithTooltip/IconWithTooltip.test.jsx
raq929: Yeah. I spent more time than I had timeboxed for testing, so... next time!

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
binodpant: Totally! my point was mainly about the resetting of the innerWidth would ideally be in cleanup, but since you ran the test and it passed without side effects, no worries.

PR #321, opened 2020-08-11
remove subscription_management feature flag

adamstankiewicz:
Now that Subscriptions is "released", we can safely remove the associated feature flag. This means it won't be necessary to append `?features=SUBSCRIPTION_MANAGEMENT` to the end of the URL for the page to appear in the sidebar navigation.


src/config/index.js
```diff
@@ -28,7 +28,6 @@ const configuration = {
 const features = {
   CODE_MANAGEMENT: process.env.FEATURE_FLAGS.CODE_MANAGEMENT || hasFeatureFlagEnabled('CODE_MANAGEMENT'),
   REPORTING_CONFIGURATIONS: process.env.FEATURE_FLAGS.REPORTING_CONFIGURATIONS,
-  SUBSCRIPTION_MANAGEMENT: process.env.FEATURE_FLAGS.SUBSCRIPTION_MANAGEMENT || hasFeatureFlagEnabled('SUBSCRIPTION_MANAGEMENT'),
```
bbaker6225, line 31: Not your change, but isn't support enabled now? cc @anaidu0

bbaker6225: `  SUPPORT: process.env.FEATURE_FLAGS.SUPPORT || hasFeatureFlagEnabled('SUPPORT'),`

adamstankiewicz: @bbaker6225 It is. Technically all feature flags here are enabled. But from code management on, we never removed the feature flags and instead just default them to `true` in webpack.{dev/prod}.config.js. And that's been the convention since haha. But really, I think we could remove all these feature flags now since they're all defaulted to `true`.

I suppose I could just go ahead and do the same for the other feature flags in this PR as well.

bbaker6225: nah, I think it's fine. :shipit:

PR #326, opened 2020-08-31
Saml Config rename + actual saml config component

taliaedX:
This does the following:

- Rename component in a bunch of places for Saml provider from SamlConfiguration to SamlProviderConfiguration since we need a component to handle the _actual_ saml configuration objects.
- Creates SamlConfiguration component
- Adds tests for SamlConfiguration component (TBD: more?)

**TO DO:**
- Any style changes per design/miro board when that gets updated.
- linter fixes.


src/components/SamlConfiguration/index.jsx
```diff
@@ -1,279 +1,57 @@
...
+//import LMSApiService from '../../data/services/LmsApiService';
 
 class SamlConfiguration extends React.Component {
   state = {
-    providerConfig: undefined,
-    providerData: undefined,
     error: undefined,
-    loading: true,
-    deleteEnabled: false,
+    configs: ['test1', 'test2'],
```
taliaedX, line 9: TODO: don't forget to set this back to []

src/components/SamlConfiguration/index.jsx
```diff
@@ -1,279 +1,57 @@
...
-   */
-  deleteProviderData = async (pdid) => {
-    try {
-      await LmsApiService.deleteProviderData(pdid, this.props.enterpriseId);
-      this.setState({ providerData: undefined });
-      return undefined;
-    } catch (error) {
-      return this.handleErrors(error);
-    }
+    options.push({label: '-- choose a configuration --', value: '', hidden: true})
```
taliaedX, line 29: TODO: pull existing config in as a prop from ProviderConfig(parent) component and use that as the default here instead of this (unless their existing config doesn't get converted to "public" I guess)

binodpant (approved):
no blockers! this renaming lgtm

PR #327, opened 2020-09-01
add Enterprise Admin registration route

adamstankiewicz:
[ENT-3250](https://openedx.atlassian.net/browse/ENT-3250)

When an Enterprise Admin navigates to the registration route (e.g., `/:enterpriseSlug/admin/register`), unauthenticated users will be redirected to the enterprise proxy logistration page to complete their registration. When registration is complete, they will be appropriately linked to the Enterprise with an "enterprise_admin" system-wide role created for the user and redirected back to the registration route.

At this point, the user should be authenticated and they are redirected to the `/:enterpriseSlug/admin/register/activate` route, which checks whether the user has verified their email address (`is_active`).  If they have, redirect them to the default page ("Learner Report") in the app and display a Toast; otherwise, poll/refresh the user account data to re-check `is_active` (i.e., once their email is verified, they automatically get redirected into the app with the Toast).

This PR also has updates to other components and many snapshots due to upgrading Paragon.

Related PRs:
* https://github.com/edx/edx-enterprise/pull/940

#### Successfully registered and activated admin account
image

#### Used an invalid email to register/login
image

#### If user hasn't verified their email address
image


src/components/UserActivationPage/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+        if (timeout) {
+          clearInterval(timeout);
+        }
+      };
+    },
+    [authentication, userAccount],
+  );
+
+  if (authentication?.username) {
+    // user is authenticated, but doesn't any JWT roles so redirect the user to
```
iloveagent57, line 46: nit: "...doesn't _have_ any JWT roles..."

iloveagent57 (approved):

PR #330, opened 2020-09-14
Remove purchase date info

bbaker6225:
This field no longer exists on the Subscription model so don't display
anything

ENT-3409


chavesj (approved):
:shipit:

PR #334, opened 2020-09-30
Send enterprise learner portal url when assigning emails

raq929:
Allows emails to link to the branded search page rather than to edx.org


config/webpack.dev.config.js
```diff
@@ -111,6 +111,7 @@ module.exports = Merge.smart(commonConfig, {
       DATA_API_BASE_URL: 'http://localhost:8000',
       ECOMMERCE_BASE_URL: 'http://localhost:18130',
       LICENSE_MANAGER_BASE_URL: 'http://localhost:18170',
+      ENTERPRISE_LEARNER_PORTAL_HOSTNAME: 'http://localhost:8734',
```
binodpant, line 114: hmm, on naming. the name says hostname, but this is actually a full blown url.. should this be called `***_URL` instead?

src/config/index.js
```diff
@@ -24,6 +24,7 @@ const configuration = {
   NODE_ENV: process.env.NODE_ENV,
   CUSTOMER_SUPPORT_EMAIL: 'customersuccess@edx.org',
   TABLEAU_URL: process.env.TABLEAU_URL,
+  ENTERPRISE_LEARNER_PORTAL_HOSTNAME: process.env.ENTERPRISE_LEARNER_PORTAL_HOSTNAME,
```
binodpant, line 27: just a process question: so for prod env, how does this work? do we set en env var in the webpack.prod.config.js at some point or do we directly set an env var for prod somewhere outside of source control (hopefully not this latter one)?

src/components/CodeAssignmentModal/index.jsx
```diff
@@ -18,8 +19,9 @@ import { ONCE_PER_CUSTOMER, MULTI_USE, CSV_HEADER_NAME } from '../../data/consta
 import { EMAIL_ADDRESS_TEXT_FORM_DATA, EMAIL_ADDRESS_CSV_FORM_DATA } from '../../data/constants/addUsers';
 
 import './CodeAssignmentModal.scss';
+import { configuration } from '../../config';
 
-class CodeAssignmentModal extends React.Component {
+class BaseCodeAssignmentModal extends React.Component {
```
binodpant, line 24: sorry if I missed prior context, the reason for renaming this to Base* ?

raq929: Because I'm connecting it to Redux, and our linter rules won't let me rename the connected component the same thing as the base component, and I prefer this pattern to calling the connected component `ConnectedCodeAssignmentModel`. I'll often use the Base class for testing, because it takes away the complexity of the redux provider, though I didn't do that in this case.

src/config/index.js
raq929: @iloveagent57 assures me that this env variable is already provided in our prod environment. So, we test on stage, and that's also why I made sure this variable exists before using it.

config/webpack.dev.config.js
raq929: Gooood point, I should check with @iloveagent57 to be sure the env variable I have is what I'm expecting to have.

src/components/CodeAssignmentModal/index.jsx
binodpant: Ah I see, but I guess I did not see what was overriding or subclassing it, so did not see the 'Base'-ness of it is what I had meant ..

src/config/index.js
binodpant: ah ok that is the part I was not clear on, looks like we have a way to set the env vars for prod, was curious about that one ... but not an issue for this PR

config/webpack.dev.config.js
raq929: CHANGE NEEDED: The env variable on stage/prod is *just* the hostname, it doesn't include the `http://`. We need to change this to account for it.

src/containers/CodeAssignmentModal/CodeAssignmentModal.test.jsx
```diff
@@ -22,9 +22,14 @@ const initialState = {
       data: {
         count: 0,
         results: [],
+
       },
     },
   },
+  portalConfiguration: {
+    enterpriseSlug: 'bearsRus',
```
binodpant, line 30: love the bears! :bear:

config/webpack.dev.config.js
binodpant: ➕

src/components/CodeAssignmentModal/index.jsx
```diff
@@ -535,6 +547,13 @@ CodeAssignmentModal.propTypes = {
   }),
 };
 
+const mapStateToProps = state => ({
+  enterpriseSlug: state.portalConfiguration.enterpriseSlug,
+  enableLearnerPortal: state.portalConfiguration.enableLearnerPortal,
+});
+
+const CodeAssignmentModal = connect(mapStateToProps)(BaseCodeAssignmentModal);
```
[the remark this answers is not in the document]
binodpant: Aargh ok, HOCs honestly are not my fav React pattern but hey you are just working within the current setup so not a reflection on the PR! I think hooks can make everything cleaner! I am also never a fan of this connect() business

binodpant (approved):
lgtm! had already done a prior review so latest changes look good for next step

PR #339, opened 2020-10-05
ENT-3332: Adds in visibility for codes on admin dashboard

jlajoie:


binodpant (approved):
pulled code down, npm i and npm start, all works, opened the page at http://localhost:1991/test-enterprise/admin/coupons but not yet setup to fully test this locally, did a pass through the code. no changes I can suggest, looking great. The feature switch should take care nicely of us releasing this out

only a Q: any plan on when we would enable this in prod? any tickets needed for subsequent actions like that?

jlajoie:
> pulled code down, npm i and npm start, all works, opened the page at http://localhost:1991/test-enterprise/admin/coupons but not yet setup to fully test this locally, did a pass through the code. no changes I can suggest, looking great. The feature switch should take care nicely of us releasing this out
> 
> only a Q: any plan on when we would enable this in prod? any tickets needed for subsequent actions like that?

So we can pass in the features=CODE_VISIBILITY for testing then once we get signoff from ken it's just setting env vars in internal.

PR #342, opened 2020-10-08
Bug fixes for Visibility

jlajoie:
* Adds in informational alert for bulk assign when selecting both public & private codes
* Fixes success message when changing visibility of codes
* Changes Visibility column display name & value display
* Fixes swapping from Public/Private to Both from erroring out

Screen Shot 2020-10-08 at 7 56 59 AM

Screen Shot 2020-10-08 at 7 58 26 AM

### DIFF
--- a/src/components/CouponDetails/index.jsx
+++ b/src/components/CouponDetails/index.jsx
@@ -68,7 +68,7 @@ class CouponDetails extends React.Component {
 
     if (features.CODE_VISIBILITY) {
       tableColumns.splice(3, 0, {
-        label: 'Is Public?',
+        label: 'Visibility',
         key: 'is_public',
       });
     }
@@ -148,7 +148,7 @@ class CouponDetails extends React.Component {
   getTableFilterVisibilitySelectionOptions() {
     return [{
       label: 'Both',
-      value: 'both',
+      value: undefined,
     }, {
       label: 'Public',
       value: 'public',
@@ -350,6 +350,12 @@ class CouponDetails extends React.Component {
     );
   }
 
+  shouldShowVisibilityStatusAlert() {
+    const { selectedCodes } = this.state;
+    return features.CODE_VISIBILITY && (selectedCodes.some(code => code.is_public) &&
+      selectedCodes.some(code => !code.is_public));
+  }
+
   hasStatusAlert() {
     // The following are the scenarios where a status alert will be shown. Note, the coupon
     // details table must be finished loading for status alert to show:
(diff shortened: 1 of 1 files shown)


src/components/CouponDetails/index.jsx
```diff
@@ -463,7 +470,10 @@ class CouponDetails extends React.Component {
 
   handleVisibilitySelect(newValue) {
     const { tableColumns, visibilityToggle } = this.state;
-    const value = newValue || visibilityToggle;
+    // Paragon InputSelect will use the `label` if the value isn't defined
+    // this will intentionally keep the value set as undefined so that qs.stringify
+    // won't send along a value we didn't set to the API.
+    const value = newValue === 'Both' ? undefined : newValue || visibilityToggle;
```
raq929, line 476: I prefer `null` to `undefined` when it's an intentionally falsey value, but that's a to total nitpick.

src/components/CouponDetails/index.jsx
```diff
@@ -626,7 +631,7 @@ class CouponDetails extends React.Component {
       assignments_remaining: `${code.redemptions.total - code.redemptions.used - code.redemptions.num_assignments}`,
       assignment_date: `${code.assignment_date}`,
       last_reminder_date: `${code.last_reminder_date}`,
-      is_public: `${code.is_public}`,
+      is_public: code.is_public ? 'Public' : 'Private',
```
raq929, line 634: I'd rather have this be a boolean than strings, but it looks like you may have changed it from a boolean to not for what I'm sure is a good reason. But that tempts me to change the status to `publicStatus` rather than `isPublic`. 
But then it's not in line with the data from the backend, so, in conclusion, this is probably fine. I might leave a comment about why you're not leaving it as a Boolean.

src/components/CouponDetails/index.jsx
```diff
@@ -148,7 +148,7 @@ class CouponDetails extends React.Component {
   getTableFilterVisibilitySelectionOptions() {
     return [{
       label: 'Both',
-      value: 'both',
+      value: undefined,
```
raq929, line 151: I prefer `null` to `undefined` if something is intentionally falsey. (Though with React sometimes there's reasons that doesn't work.)

raq929 (approved):

src/components/CouponDetails/index.jsx
jlajoie: Would _also_ prefer it but the qs package will serialize null params but will drop undefined ones which is what we want here

src/components/CouponDetails/index.jsx
jlajoie: This is for display purposes, product request

PR #345, opened 2020-10-08
remove assigned license count from revoke modal

adamstankiewicz:


src/components/LicenseRevokeModal/LicenseRevokeModal.scss
```diff
@@ -11,7 +11,7 @@
     }
 
     .license-details {
-      padding: 1rem 1.75rem;
+      padding: 1rem;
```
adamstankiewicz, line 14: make padding consistent with the alert padding on the left/right per UX feedback.

src/components/LicenseRevokeModal/index.jsx
```diff
@@ -123,12 +122,6 @@ class LicenseRevokeModal extends React.Component {
               assign <strong>{user.userEmail}</strong> to another license, but they
               will need to re-enroll in any course after being assigned a new license.
             </p>
-            {this.shouldRenderRevocationCapMessaging() && licenseOverview.assigned > 0 && (
-              <p>
-                You have <strong>{licenseOverview.assigned}</strong> licenses that are
-                in pending status that could be removed or re-assigned.
-              </p>
-            )}
```
adamstankiewicz, line 131: After a conversation with UX, we are opting to remove this sentence for now rather than investing in refactoring to make the `licenseOverview.assigned` count used here not consider an active search query. If UX decides its crucial, we'll come back to it.

src/components/subscriptions/constants.js
```diff
@@ -19,5 +19,5 @@ export const SUBSCRIPTION_USERS_OVERVIEW = 'Subscription Users Overview';
 export const NETWORK_ERROR_MESSAGE = 'Error occurred while loading the data.';
 export const DEFAULT_PAGE = 1;
 
-// used to determine whether to show the revocation cap messaging
-export const SHOW_REVOCATION_CAP_PERCENT = 10;
+// used to determine whether to show the revocation cap messaging in the license revoke modal
+export const SHOW_REVOCATION_CAP_PERCENT = 80;
```
adamstankiewicz, line 23: After conversation with Joe, we're opting to raise this percent to extend number of revocations the customer must do before they would see the revocation cap alert.

chavesj (approved):
🎱  0️⃣

PR #347, opened 2020-10-13
show tooltips for New Email and From Template buttons

muhammad-ammar:
**Description:** Show tooltips for `New Email` and `From Template` buttons to reduce confusion in email template creation for code assignments.
 
**JIRA:** [ENT-3425](https://openedx.atlassian.net/browse/ENT-3425)


zamanafzal (approved):
👍

PR #349, opened 2020-10-22
Add subscription expiration experience for admins

bbaker6225:
Note that this doesn't provide any sort of post-expiration experience
and there isn't any backend validation preventing admins from doing
things once their subscription expires. That will come later.

ENT-3504

### Modals
**Pre-expiration modal**
image
**Post-expiration modal**
image

### Banners
**<= 120 day banner:**
image

**<= 60 day banner:**
image

**<= 30 day banner:**
image

### DIFF
--- a/src/components/subscriptions/SubscriptionExpirationBanner.jsx
+++ b/src/components/subscriptions/SubscriptionExpirationBanner.jsx
@@ -1,0 +1,49 @@
+import React, { useContext } from 'react';
+import { MailtoLink } from '@edx/paragon';
+
+import StatusAlert from '../../components/StatusAlert';
+import { SubscriptionContext } from './SubscriptionData';
+import {
+  SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD,
+  SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD,
+  SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD,
+} from './constants';
+
+export default function SubscriptionExpirationBanner() {
+  const { details } = useContext(SubscriptionContext);
+  const { daysUntilExpiration } = details;
+
+  const renderMessage = () => (
+    <React.Fragment>
+      Your subscription is {daysUntilExpiration} days from expiration.
+      Contact the edX Customer Success team at
+      {' '}
+      <MailtoLink to="customersuccess@edx.org">customersuccess@edx.org</MailtoLink>
+      {' '}
+      to extend your contract.
+    </React.Fragment>
+  );
+
+  if (daysUntilExpiration > SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD) {
+    return null;
+  }
+
+  let dismissible = true;
+  let alertType = 'info';
+  if (daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD) {
+    alertType = 'warning';
+  }
+  if (daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD) {
+    dismissible = false;
+    alertType = 'danger';
+  }
+
+  return (
+    <StatusAlert
+      className="mt-1"
+      alertType={alertType}
+      message={renderMessage(daysUntilExpiration)}
+      dismissible={dismissible}
+    />
+  );
+}
--- a/src/components/subscriptions/SubscriptionExpirationBanner.jsx
+++ b/src/components/subscriptions/SubscriptionExpirationBanner.jsx
@@ -40,7 +40,7 @@ export default function SubscriptionExpirationBanner() {
 
   return (
     <StatusAlert
-      className="mt-1"
+      className="expiration-alert mt-1"
       alertType={alertType}
       message={renderMessage(daysUntilExpiration)}
       dismissible={dismissible}
--- a/src/components/subscriptions/SubscriptionExpirationBanner.jsx
+++ b/src/components/subscriptions/SubscriptionExpirationBanner.jsx
@@ -4,9 +4,9 @@ import { MailtoLink } from '@edx/paragon';
 import StatusAlert from '../../components/StatusAlert';
 import { SubscriptionContext } from './SubscriptionData';
 import {
-  SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD,
-  SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD,
-  SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD,
+  SUBSCRIPTION_DAYS_REMAINING_MODERATE,
+  SUBSCRIPTION_DAYS_REMAINING_SEVERE,
+  SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL,
 } from './constants';
 
 export default function SubscriptionExpirationBanner() {
@@ -24,16 +24,16 @@ export default function SubscriptionExpirationBanner() {
     </React.Fragment>
   );
 
-  if (daysUntilExpiration > SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD) {
+  if (daysUntilExpiration > SUBSCRIPTION_DAYS_REMAINING_MODERATE) {
     return null;
   }
 
   let dismissible = true;
   let alertType = 'info';
-  if (daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD) {
+  if (daysUntilExpiration <= SUBSCRIPTION_DAYS_REMAINING_SEVERE) {
     alertType = 'warning';
   }
-  if (daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD) {
+  if (daysUntilExpiration <= SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL) {
     dismissible = false;
     alertType = 'danger';
   }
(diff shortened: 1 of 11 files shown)


src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  // If the subscription has already expired, we show a different un-dismissible modal
+  const subscriptionExpired = daysUntilExpiration <= 0;
+  if (subscriptionExpired) {
+    return (
+      <Modal
+        renderHeaderCloseButton={false}
+        renderDefaultCloseButton={false}
+        title={null}
```
bbaker6225, line 81: Paragon doesn't really support no title, but BCool said let's go with this for now and he'll mention updating the modal to not require a title to the paragon working group soon

src/components/subscriptions/constants.js
```diff
@@ -21,3 +21,11 @@ export const DEFAULT_PAGE = 1;
 
 // used to determine whether to show the revocation cap messaging in the license revoke modal
 export const SHOW_REVOCATION_CAP_PERCENT = 80;
+
+// Subscription expiration
+// Days until expiration threshold constants
+export const SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD = 120;
```
bbaker6225, line 27: I don't love these names I came up with

src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+  if (seenCurrentExpirationModal) {
+    return null;
+  }
+
+  return (
+    <Modal
+      renderHeaderCloseButton={false}
+      title={renderTitle()}
+      body={renderBody()}
+      closeText="Ok"
```
adamstankiewicz, line 112: The current Paragon modal, as you probably realized, doesn't allow modifying the class name of the default close button unfortunately 😞 

That said, since this is the only action in the modal, it should have the primary variant, matching what's in the wireframe. We might get around this for now by hiding the default close button, and passing our own "Ok" button as a prop, managing whether the modal is open via the Modal's `open` prop:

```jsx
const [isOpen, setIsOpen] = useState(false);

<Modal
  buttons={[
    <Button
      onClick={() => {
        cookies.set(...);
        setIsOpen(false);
      }
    >
      Ok
    </Button>
  ]}
  open={isOpen}
/>
```

Or, alternatively, we could also override the CSS of the default close button to use the `.btn-primary` styles.

src/components/subscriptions/constants.js
iloveagent57: `SUBSCRIPTION_DAYS_REMAINING_MODERATE`, ``SUBSCRIPTION_DAYS_REMAINING_SEVERE`, `SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL`.

https://droughtmonitor.unl.edu/About/AbouttheData/DroughtClassification.aspx

src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+    return null;
+  }
+
+  let expirationThreshold = SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD;
+  if (details.daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD) {
+    expirationThreshold = SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD;
+  }
+  if (daysUntilExpiration <= SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD) {
+    expirationThreshold = SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD;
+  }
```
iloveagent57, line 97:
```
const thresholds = [SUBSCRIPTION_EXPIRATION_FIRST_THRESHOLD, SUBSCRIPTION_EXPIRATION_SECOND_THRESHOLD, SUBSCRIPTION_EXPIRATION_THIRD_THRESHOLD]
const expirationThreshold = thresholds.find(element => element > details.daysUntilExpiration);
```

src/components/subscriptions/constants.js
iloveagent57: or maybe just `export const SUBSCRIPTIONS_REMAINING_DAYS_THRESHOLDS = [30, 60, 120]`?

src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+          To reactivate your subscriptions please cotact the edX Customer Success team at
+          {' '}
+          <MailtoLink to="customersuccess@edx.org">customersuccess@edx.org</MailtoLink>
+        </li>
+        <li>
+          View your learner progress in the <Link to={`/${enterpriseSlug}/admin/learners`}>learner management page</Link>
+        </li>
+        <li>
+          Manage your codes in the <Link to={`/${enterpriseSlug}/admin/coupons`}>code management page</Link>
+        </li>
```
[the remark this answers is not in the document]
bbaker6225: Good catch

src/components/subscriptions/constants.js
bbaker6225: Love the drought classification. Perfection

src/components/subscriptions/SubscriptionExpirationModal.jsx
bbaker6225: ah right I missed that. I think just overriding the close button to match `btn-primary` makes sense. Is there a way to get the styles from paragon's scss like bootstrap does here: https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_buttons.scss#L71-L73? If not I can copy the styles manually

src/components/subscriptions/SubscriptionExpirationModal.jsx
bbaker6225: nice

src/components/subscriptions/SubscriptionExpirationModal.jsx
bbaker6225: Updated the styles for the post expiration header. Personally I'm not sold on the event listener being worth it since (as you pointed out):
- No real subscriptions are set to expire this calendar year
- We have to add in backend validation anyway
- The existing modal is getting replaced in Paragon "soon"

I could be convinced if you/others feel really strongly, but I feel like this is ok for now

src/components/subscriptions/SubscriptionExpirationModal.test.jsx
```diff
@@ -0,0 +1,88 @@
...
+const history = createMemoryHistory({
+  initialEntries: ['/'],
+});
+
+jest.mock('universal-cookie', () => {
+  const mockCookie = {
+    get: jest.fn(() => true),
+  };
+  return jest.fn(() => mockCookie);
+});
```
bbaker6225, line 22: Mocking this cookie stuff for separate tests is not going to well for me. I can do what I have right now and get it to always be true or false, but I'm having trouble clearing the mocks so that both the cookie and non-cookie test work.
What I've tried with `doMock`, `setMockImplementation`, `jest.spyOn` hasn't worked, so please let me know if you all have other ideas

src/components/subscriptions/SubscriptionExpirationBanner.jsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  let dismissible = true;
+  let alertType = 'info';
+  if (daysUntilExpiration <= SUBSCRIPTION_DAYS_REMAINING_SEVERE) {
+    alertType = 'warning';
+  }
+  if (daysUntilExpiration <= SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL) {
+    dismissible = false;
+    alertType = 'danger';
```
kellmarie, line 38: Are these alertTypes not in a constants file anywhere? I would move them into a constants file and then import them if they aren't already existing in a constants file somewhere.

src/components/subscriptions/SubscriptionExpirationModal.test.jsx
kellmarie: Did you try not mocking it? If it's just a wrapper class around standard front-end cookies (i.e. document.cookie) then you might not need to mock it at all. And you could try manually clearing the cookies after each test as you would normally using the Cookie class.

src/components/subscriptions/SubscriptionExpirationBanner.jsx
bbaker6225: These are just the bootstrap "themes" https://getbootstrap.com/docs/4.0/utilities/colors/so I don't think it's especially worthwhile to use constants here whereas our other frontends do not

src/components/subscriptions/SubscriptionExpirationModal.test.jsx
bbaker6225: I have tried that, that gives the opposite issue. I have a different idea I'm trying however

bbaker6225:
Ok the test is now working for a full flow

src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -0,0 +1,148 @@
...
+      <p>
+        Your subscription license expired on {formatTimestamp({ timestamp: expirationDate })}.
+        To access your subscription management page contact edX and reactivate your subscriptions.
+      </p>
+      <p>
+        What to do next?
+      </p>
+      <ul>
+        <li>
+          To reactivate your subscriptions please contact the edX Customer Success team at
```
iloveagent57, line 61: Nit: "subscription" singular, or "subscription(s)", if you're afraid of commitment.

iloveagent57 (approved):

src/components/subscriptions/SubscriptionExpirationModal.jsx
bbaker6225: I am indeed afraid

PR #350, opened 2020-10-23
Reset course search filter label

raq929:
Course search filter was not properly resetting because it's value was undefined
This changes it to the empty string which is the default option of 'All courses'


alex-sheehan-edx (approved):
LGTM! 👍 .

PR #354, opened 2020-11-02
Clean up console errors on tests

raq929:
This PR is a first pass at removing console errors on tests, with the goal of eventually making a console error fail the test. 

This will both allow us to have more signal and less noise when looking at test results, and catch more errors when they occur, as React console errors are often indicative of actual problems and/or performance issues.

### DIFF
--- a/src/setupTest.js
+++ b/src/setupTest.js
@@ -1,9 +1,17 @@
 /* eslint-disable import/no-extraneous-dependencies */
-
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new Adapter() });
+
+
+const { error } = global.console;
+
+global.console.error = (...args) => {
+  error(...args);
+  throw new Error(args.join(' '));
+};
+
 
 // These configuration values are usually set in webpack's EnvironmentPlugin however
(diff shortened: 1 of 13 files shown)


src/setupTest.js
```diff
@@ -1,10 +1,19 @@
 /* eslint-disable import/no-extraneous-dependencies */
-
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new Adapter() });
 
+
+// TODO: Once there are no more console errors in tests, uncomment the code below
+// const { error } = global.console;
```
jlajoie, line 9: Is this worth uncommenting now or something we're tracking for subsequent work?

jlajoie (approved):
Don't have a lot of context on the jsdom stuff but seems sane, just the one question but looks good to go 👍

src/setupTest.js
raq929: I didn't end up having time to fix all the test errors, so there are still 9 tests that fail if I uncomment. I'm going to open a ticket for subsequent work and leave this here so I don't have to look it up again. ;) 
(I suspect some of the errors are there because they're tricky to get rid of; I'm *hoping* none of them are coming from third-party code.)

PR #359, opened 2020-11-23
ENT-3657: Ability to Revoke Code, but Not Email

saleem-latif:
JIRA Ticket: [ENT-3657](https://openedx.atlassian.net/browse/ENT-3657)

__Description:__
"As an admin, I want the ability to revoke my learner's code, without having to email them, to make the revoke a softer experience."

A/C:

1. Be able to revoke codes without having to email learners and notify them.

### DIFF
--- a/src/components/CodeRevokeModal/index.jsx
+++ b/src/components/CodeRevokeModal/index.jsx
@@ -1,13 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Field, reduxForm, SubmissionError } from 'redux-form';
-import { Button, Icon, Modal } from '@edx/paragon';
+import { Button, Icon, Input, Modal } from '@edx/paragon';
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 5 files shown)


src/components/CodeRevokeModal/index.jsx
```diff
@@ -140,13 +150,36 @@ class CodeRevokeModal extends React.Component {
     /* eslint-enable no-underscore-dangle */
   }
 
+  doNotEmailField({ input }) {
```
sameenfatima78, line 153: Will this checkbox appear in both "New Email" and "From Template" section?

sameenfatima78 (approved):
Just one question, otherwise LGTM

src/components/CodeRevokeModal/index.jsx
saleem-latif: My understanding is that we need to show it in both sections, but I will also confirm it with Nathan in the standup today.

zamanafzal (approved):

PR #360, opened 2020-11-23
Rebrand!

raq929:
- Import new fonts / logos
- Update scss to use paragon variables for font-size
- Use dotenv and dotenv-webpack to inject env variables
- Get favicon from webpack / env variable
- Make footer stick to the bottom of the screen


package.json
```diff
@@ -34,6 +34,10 @@
     "babel-eslint": "^10.1.0",
     "classnames": "^2.2.5",
     "core-js": "^3.6.5",
+    "dotenv": "^8.2.0",
+    "dotenv-webpack": "^5.1.0",
+    "eslint": "^7.14.0",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
```
adamstankiewicz, line 40: Should these be devDependencies since they are not used in production build?

raq929: Yes... I thought I had moved them, will do it now.

package.json
```diff
@@ -46,6 +50,7 @@
     "query-string": "^5.1.1",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
+    "react-focus-on": "^3.5.0",
```
adamstankiewicz, line 53: Do we need this as a dep? I think this is leftover from the Paragon issues in CI.

raq929: I couldn't get tests to pass without it as a dep, so yes.

config/webpack.common.config.js
```diff
@@ -6,7 +6,6 @@ module.exports = {
   entry: {
     segment: path.resolve(__dirname, '../src/segment.js'),
     app: path.resolve(__dirname, '../src/index.jsx'),
-    demoDataLoader: path.resolve(__dirname, '../src/demo/index.js'),
```
adamstankiewicz, line 9: ☠️

src/components/CouponDetails/index.jsx
```diff
@@ -969,25 +974,25 @@ class CouponDetails extends React.Component {
...
-                            </Button>
+                          <Button
+                            variant="link"
+                            className="p-0 pl-1 border-0"
+                            onClick={() => this.setState({
+                              hasAllCodesSelected: true,
+                            })}
+                          >
+                            {`Select all ${tableData.count} codes?`}
+                          </Button>
```
adamstankiewicz, line 988: nit: Is this indenting correct? It looks like the original indent spacing was previously correct.

src/components/CouponDetails/index.jsx
```diff
@@ -1002,29 +1007,32 @@ class CouponDetails extends React.Component {
                 columns={tableColumns}
                 formatData={this.formatCouponData}
               />
-              {modals.assignment &&
+              {modals.assignment
```
adamstankiewicz, line 1010: nit: 

```jsx
{modals.assignment && (
  // ...
)}
```

And similar cases below...

src/components/ErrorPage/index.jsx
```diff
@@ -6,16 +6,15 @@ import StatusAlert from '../StatusAlert';
 import NotFoundPage from '../NotFoundPage';
 import ForbiddenPage from '../ForbiddenPage';
 
-
 function renderErrorComponent(status, message) {
   const errorMessage = message || 'An unknown error has occured.';
   if (status === 404) {
     return <NotFoundPage />;
-  } else if (status === 403) {
+  } if (status === 403) {
```
adamstankiewicz, line 13: nit: 2nd `if` on its own line

src/components/Header/index.jsx
```diff
@@ -2,15 +2,17 @@ import React from 'react';
...
 
 import SidebarToggle from '../../containers/SidebarToggle';
 import Img from '../Img';
 
 import apiClient from '../../data/apiClient';
 
-import EdxLogo from '../../images/edx-logo.png';
 import './Header.scss';
 
+const logo = configuration.LOGO_URL;
```
adamstankiewicz, line 14: nit: It feels a bit odd to create this variable outside of the component. Any reason not to pull this line into the `renderLogo()` function?

src/components/Hero/index.jsx
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import edXLogoForBanner from '../../images/edx-logo-for-banner.png';
 import './Hero.scss';
 
+import { configuration } from '../../config';
+
+const edxWhiteSemiTransparentLogo = configuration.LOGO_WHITE_URL;
```
adamstankiewicz, line 8: nit: It feels a bit odd to create this variable outside of the component. Any reason not to reference `configuration.LOGO_WHITE_URL` directly on line 18?

src/components/Hero/index.jsx
```diff
@@ -12,7 +15,7 @@ const Hero = props => (
           <h1>{props.title}</h1>
         </div>
         <div className="col-12 col-md-3 text-md-right mt-2">
-          <img src={edXLogoForBanner} alt="edX logo" />
+          <img src={edxWhiteSemiTransparentLogo} alt="edX logo" />
```
adamstankiewicz, line 18: Does the LOGO_WHITE_URL have any transparency baked in? The original white logo used here was semi-transparent and partially revealed the background so as to be more subtle than solid white.

Do we need to add transparency (through CSS) to the new white logo here?

src/components/NumberCard/index.jsx
```diff
@@ -187,19 +185,22 @@ class NumberCard extends React.Component {
               <span>
                 {this.formatTitle(title)}
               </span>
-              {iconClassName &&
+              {iconClassName
```
adamstankiewicz, line 188: nit:

```jsx

{iconClassName && (
  // ...
)}

File: src/components/ReportingConfig/EmailDeliveryMethodForm.jsx [hunk 11]
```diff
@@ -22,13 +22,12 @@ const EmailDeliveryMethodForm = ({ invalidFields, config, handleBlur }) => {
             id="email"
             name="emailRaw"
             defaultValue={config ? config.email.join('\n') : undefined}
-            onBlur={e => handleBlur(e, () =>
-              !isEmpty(e.target.value.split('\n')
-                .filter(email => !isEmail(email))))
-            }
+            onBlur={e => handleBlur(e, () => !isEmpty(e.target.value.split('\n')
```
adamstankiewicz, line 25: nit: something like this might be a bit more readable?

```jsx
onBlur={e => handleBlur(e, () => {
  const rows = e.target.value.split('\n');
  const emails = rows.filter(email => !isEmail(email));
  return !isEmpty(emails)
})}
```

src/components/SamlProviderConfiguration/SamlProviderConfigForm.jsx
```diff
@@ -330,15 +330,16 @@ class SamlProviderConfigForm extends React.Component {
             />
           </div>
           <div className="col col-2">
-            {config && deleteEnabled &&
+            {config && deleteEnabled
```
adamstankiewicz, line 333: nit: 

```jsx
{config && deleteEnabled && (
  // ...
)}
```

src/components/SupportPage/__snapshots__/SupportPage.test.jsx.snap
```diff
@@ -23,7 +23,6 @@ Array [
         >
           <img
             alt="edX logo"
-            src="test-file-stub"
```
adamstankiewicz, line 26: Looks like jest doesn't have a `LOGO_URL` environment variable defined?

src/components/SupportPage/index.jsx
```diff
@@ -48,16 +48,18 @@ class SupportPage extends React.Component {
...
                   match={match}
                 />
-              ) :
-                <p>
-                  For assistance, please contact the edX Customer Success team at
-                  {' '}
-                  <MailtoLink to="customersuccess@edx.org">customersuccess@edx.org</MailtoLink>.
-                </p>}
+              )
+                : (
```
adamstankiewicz, line 52: nit: these 2 lines can be combined into one.

src/components/TemplateSourceFields/index.jsx
```diff
@@ -166,26 +169,28 @@ class TemplateSourceFields extends React.Component {
...
-          <Field
-            name="template-name-select"
-            component={this.selectRenderField}
-            templateOptions={this.state.templateOptions}
-            changeFromEmailTemplate={this.changeFromEmailTemplate}
-            disabled={disabled}
-          />
-        }
-      </React.Fragment>
+        {emailTemplateSource === newEmail
```
adamstankiewicz, line 172: nit:

```jsx
{emailTemplateSource === newEmail ? (
  // ...
) : (
  // ...
)}
```

src/components/subscriptions/LicenseActions.jsx
```diff
@@ -25,6 +25,7 @@ export default function LicenseAction({ user }) {
 
   const hasNoRevocationsRemaining = useHasNoRevocationsRemaining(details);
   const noActionsAvailable = [{ key: 'no-actions-here', text: '-' }];
+  const { user } = props;
```
adamstankiewicz, line 28: nit: curious the reason for moving the props destructing for `user`  from the function argument to here?

src/components/subscriptions/SubscriptionExpirationModal.jsx
```diff
@@ -12,14 +12,13 @@ import {
...
-import Img from '../../components/Img';
-import edxLogo from '../../images/edx-logo.png';
+import { configuration } from '../../config';
+import Img from '../Img';
 import { formatTimestamp } from '../../utils';
 
 export const MODAL_DIALOG_CLASS_NAME = 'subscription-expiration';
 
-
+const edxLogo = configuration.LOGO_URL;
```
adamstankiewicz, line 21: nit: Should this be defined where the variable is used in `renderExpiredBody()`?

src/config/index.js
```diff
@@ -1,4 +1,5 @@
 import qs from 'query-string';
+// import { getConfig } from '@edx/frontend-platform';
```
adamstankiewicz, line 2: nit: unused (commented) import

src/config/index.js
```diff
@@ -26,6 +27,8 @@ const configuration = {
   CUSTOMER_SUPPORT_EMAIL: 'customersuccess@edx.org',
   TABLEAU_URL: process.env.TABLEAU_URL,
   ENTERPRISE_LEARNER_PORTAL_URL: process.env.ENTERPRISE_LEARNER_PORTAL_URL,
+  LOGO_URL: process.env.LOGO_URL,
+  LOGO_WHITE_URL: process.env.LOGO_WHITE_URL,
```
adamstankiewicz, line 31: nit: might include `LOGO_TRADEMARK_URL` in here too so it's accessible if/when needed.

src/components/Sidebar/index.jsx
```diff
@@ -140,7 +140,6 @@ class Sidebar extends React.Component {
         onFocus={() => !this.isSidebarExpanded() && expandSidebar()}
         onMouseLeave={() => this.shouldSidebarCollapse() && collapseSidebar()}
         onBlur={() => this.shouldSidebarCollapse() && collapseSidebar()}
-        aria-expanded={this.isSidebarExpanded()}
```
adamstankiewicz, line 143: Was this intentionally removed? If so, may I ask why? I believe `aria-expanded` is needed for a11y purposes.

config/webpack.prod.config.js
```diff
@@ -1,18 +1,28 @@
...
+const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 const OptimizeCssnanoPlugin = require('@intervolga/optimize-cssnano-plugin');
+const dotenv = require('dotenv');
+const Dotenv = require('dotenv-webpack');
+const commonConfig = require('./webpack.common.config.js');
+
+// Add process env vars. Currently used only for setting the
+// server port and the publicPath
+const result = dotenv.config({
+  path: path.resolve(process.cwd(), '.env.development'),
```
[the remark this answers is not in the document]
raq929: YES YES IT SHOULD, THANK YOU FOR THIS IMPORTANT CATCH

package.json
```diff
@@ -8,30 +8,30 @@
...
     "@edx/frontend-auth": "5.3.6",
+    "@edx/frontend-build": "5.5.1",
     "@edx/frontend-logging": "2.1.0",
-    "@edx/paragon": "12.0.4",
+    "@edx/paragon": "12.2.0",
     "@fortawesome/fontawesome-svg-core": "1.2.30",
     "@fortawesome/free-solid-svg-icons": "5.14.0",
     "@fortawesome/react-fontawesome": "0.1.11",
-    "@fullstory/browser": "1.4.5",
+    "@fullstory/browser": "1.4.4",
```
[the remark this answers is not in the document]
raq929: Yeeeaaah the rebase was annoying; I'll fix it

package.json
```diff
@@ -65,51 +65,32 @@
...
+  "peerDependencies": {
+    "dotenv": "^8.2.0",
+    "dotenv-webpack": "^3.0.0",
+    "webpack": "^4.44.2",
+    "webpack-merge": "^5.4.0",
+    "mini-css-extract-plugin": "^0.11.2",
+    "html-webpack-new-relic-plugin": "^2.0.0",
+    "html-webpack-plugin": "4.5.0",
+    "clean-webpack-plugin": "^3.0.0"
+  },
```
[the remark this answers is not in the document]
raq929: Yes to all. These can be removed as soon as we're fully using frontend-build

package.json
```diff
@@ -65,51 +65,32 @@
...
+    "enzyme-adapter-react-16": "1.15.1",
     "es-check": "5.1.0",
-    "eslint-config-edx": "4.0.4",
-    "file-loader": "1.1.11",
-    "gh-pages": "1.2.0",
-    "html-webpack-exclude-assets-plugin": "0.0.7",
-    "html-webpack-harddisk-plugin": "0.2.0",
-    "html-webpack-new-relic-plugin": "1.1.0",
-    "html-webpack-plugin": "3.2.0",
+    "gh-pages": "1.1.0",
```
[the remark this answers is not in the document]
raq929: I was wondering about that! I'll take it out.

src/components/subscriptions/LicenseActions.jsx
raq929: I think this was an attempt to see if moving it would get rid of that buggy eslint error.

src/components/subscriptions/SubscriptionExpirationModal.jsx
raq929: Yeah, I think this was from my initial pass when I was just seeing what worked and not really paying attention to how the variable was being used. I'll see about cleaning it up.

src/config/index.js
raq929: Yeah, makes sense.

package.json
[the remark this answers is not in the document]
raq929: Aaaaand I really wish I could leave a comment to that effect in the JSON, but alas.

config/webpack.prod.config.js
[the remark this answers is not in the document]
raq929: 😱

package.json
```diff
@@ -65,51 +65,32 @@
...
-    "redux-mock-store": "1.5.4",
-    "sass-loader": "6.0.7",
-    "style-loader": "0.20.3",
-    "webpack": "4.44.1",
-    "webpack-cli": "3.3.12",
-    "webpack-dev-server": "3.11.0",
-    "webpack-merge": "4.2.2"
+    "redux-mock-store": "1.5.1"
   },
   "jest": {
```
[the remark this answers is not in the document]
raq929: Frontend-build appears to be doing all the magic! 🧙 
Or at least, all the tests still pass when I take this out, and I'm counting that as all the magic I need.

src/components/Sidebar/index.jsx
raq929: It was, eslint told me I didn't need it here because it was implicit for nav components? (Or something, I don't remember the details). Seemed a bit weird to me. I could suppress the error instead?

adamstankiewicz: Did some additional reading. `aria-expanded` is typically used on interactive elements like a button, not a `nav`. I think eslint is correct here. 

Based on what I was reading, it seems `aria-expanded` is typically used to indicate that an interactive element like a button can be used to expand/collapse additional content. However, since the `nav` is technically always visible, either showing just the icons or the full menu item, I don't think `aria-expanded` actually makes sense in this case.

Let's keep it removed! The sidebar toggle in the header still refers to the `nav` sidebar through `aria-controls`.

adamstankiewicz (approved):

PR #361, opened 2020-11-25
feat: ENT- Bulk Enrollment screen on admin portal

binodpant:
Checklist before merge ready

- [X] Feature switch for browser
- [ ] Feature switch for enable_portal_bulk_enrollment_screen from backend (not using it until switch is available from/enterprise/api/v1/enterprise-customer/dashboard_list endpoint)
- [ ] BulkEnrollmentPage component to grab and (basic) display the data in a basic way (without needing the Table component Lael is working on)
- [ ] Tests


package.json
```diff
@@ -14,7 +14,7 @@
     "prepublishOnly": "npm run build",
     "start": "NODE_ENV=development BABEL_ENV=development webpack-dev-server --config=config/webpack.dev.config.js --hot --inline --progress",
     "debug-test": "node --inspect-brk node_modules/.bin/jest --runInBand --coverage",
-    "test": "jest --coverage --runInBand",
+    "test": "jest --coverage --maxWorkers=50%",
```
binodpant, line 17: this speeds up tests significantly and also makes use of half of the cores available which optimizes jest's use of resources

src/config/index.js
```diff
@@ -26,10 +26,16 @@ const configuration = {
...
+  ALGOLIA: {
+    APP_ID: process.env.ALGOLIA_APP_ID,
+    SEARCH_API_KEY: process.env.ALGOLIA_SEARCH_API_KEY,
+    INDEX_NAME: process.env.ALGOLIA_INDEX_NAME,
+  },
 };
 
 const features = {
   CODE_MANAGEMENT: process.env.FEATURE_FLAGS.CODE_MANAGEMENT || hasFeatureFlagEnabled('CODE_MANAGEMENT'),
+  BULK_ENROLLMENT: process.env.FEATURE_FLAGS.BULK_ENROLLMENT || hasFeatureFlagEnabled('BULK_ENROLLMENT'),
```
raq929, line 38: Yeah, so this is the only place you need the feature flag; in .env variables. Heads up the env variable process is going to change a bit today when I merge my PR.
There's no need to add this to the backend; we don't need enterprises to be able to set or unset this feature. When we're done with it, we'll remove the flag and everyone will have it.

binodpant:
closing this in favor of a new PR

PR #365, opened 2020-11-25
feat: ENT-3742 upgrade react-router ecosystem

binodpant:
does not work yet but made a lot of progress


raq929:
This has now been done via other PRs.

PR #378, opened 2020-11-30
ENT-3614: Adds in Canvas Form

jlajoie:


src/containers/MoodleIntegrationConfig/index.jsx
```diff
@@ -0,0 +1,9 @@
+import { connect } from 'react-redux';
```
taliaedX, line 1: I think I removed this on my branch, but if not I should.

This redux container only gets triggered in the enterpriseApp page and since we are only referencing LmsConfigurations, Moodle/CanvasIntegrationConfig containers are useless. So for those they should just involve standard prop passing from LmsConfigurations (which I know I recently did for Moodle)

taliaedX (commented):
there's a couple things I'm not so sure about in terms of what data we want to require, but otherwise it's fine.

src/components/LmsConfigurations/CanvasIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,273 @@
...
+import { snakeCaseFormData } from '../../utils';
+import StatusAlert from '../StatusAlert';
+import LmsApiService from '../../data/services/LmsApiService';
+import NewRelicService from '../../data/services/NewRelicService';
+
+export const REQUIRED_CONFIG_FIELDS = [
+  'clientId',
+  'clientSecret',
+  'canvasAccountId',
+  'active',
```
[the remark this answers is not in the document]
taliaedX: oh seeing down further where there is no invalid label on the active field, I'm not sure this would ever get triggered anyways. You probably should remove active from this list.

src/containers/MoodleIntegrationConfig/index.jsx
taliaedX: Point here was I don't think we need the Canvas container either.

src/components/LmsConfigurations/CanvasIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,275 @@
...
+        invalidFields: {
+          ...invalidFields,
+        },
+        submitState: SUBMIT_STATES.DEFAULT,
+      });
+      return;
+    }
+
+    formData.append('enterprise_customer', this.props.enterpriseId);
+    formData.set('active', this.state.active);
```
taliaedX, line 75: I think I said some useless things in my previous comment.
What I was trying to understand is if you create a config, don't refresh, and then submit again, does the formData *only* include fields changed or does it just pass everything along again? Are the fields passed along based on the data actually entered in or does it trigger the props again (and thus everything resets?)

I think if formData only passes along fields modified, then you don't need to do anything with this active field. Only in the case where a user actually changes that flag would it need to be passed along and in that case it would be as part of formData.

However, if formData just passes everything along and uses whatever values were in the props, then yeah not only am I concerned about this field, but about every field, because there wouldn't be any props on initial creation.

If I'm sounding confusing (which is very likely) we can chat more offline. I want to clarify I'm not dictating a change here, I'm actually just unsure what formData is passing in this case. I've found it passes everything, but you're using an onChange override that updates state you don't otherwise seem to be using, so I'm not sure if formData gets updated in the same way....

src/components/LmsConfigurations/index.jsx
```diff
@@ -10,13 +11,15 @@ import ErrorPage from '../ErrorPage';
 class LmsConfigurations extends React.Component {
   state = {
     moodleConfig: null,
+    canvasConfig: null,
     error: null,
     loading: true,
-  }
+  };
```
taliaedX, line 17: surprised it didn't yell at me on my pr for this style error.

taliaedX (approved):
LGTM!

PR #388, opened 2020-12-02
ENT-3671 Keep email field populated while switching in between new and old templates

zamanafzal:
Here is the [Jira Ticket](https://openedx.atlassian.net/browse/ENT-3671) for this issue.


moconnell1453 (approved):

irfanuddinahmad (approved):

sameenfatima78 (changes requested):
Changes look good but shouldn't we add the same changes on Remind and Revoke modals too?

zamanafzal:
@sameenfatima78 For remind and revoke email field is not editable. So, we need this change for only assign modals.

sameenfatima78 (approved):

PR #390, opened 2020-12-02
ENT-3694: Batch subscriptions UI

chavesj:
Multiple Subscription Admin Experience
-
### Ticket
[ENT-3694](https://openedx.atlassian.net/browse/ENT-3694)

### Testing Instructions

I suggest configuring test data through LMS and license-manager Django admin as follows:
- Enterprise Customer A
    - 1 SubscriptionPlan that has not had any licenses assigned or activated
- Enterprise Customer B
    - 2 SubscriptionPlans

Use Customer A to test the redirect from the list view to the detail view when there is only one available subscription plan. While on the detail view you will be able to test that the zero-state functionality is still in tact since none of the licenses for Customer A's subscription plan have been assigned or activated. 

Use Customer B to test the list view which displays a card for each subscription plan available to the given customer.

Significant refactoring was done to the Subscriptions Management Page to make this possible. Poking around to ensure all other functionality for the detail view is still in tact is highly encouraged. I also invite you to critique the new directory structure within the subscriptions folder.

Note: A bug exists in which the overview (and sometimes detail) context doesn't actually refresh after a successful license assignment or revocation. See [comment](https://github.com/edx/frontend-app-admin-portal/pull/390#issuecomment-745460170) below for details. My recommendation is to address that bug in a separate PR and avoid any further delays.


src/components/subscriptions/SubscriptionDetailContextProvider.jsx
```diff
@@ -0,0 +1,52 @@
...
+    activeTab,
+    currentPage,
+    hasMultipleSubscriptions,
+    overview,
+    searchQuery,
+    setActiveTab,
+    setCurrentPage,
+    setSearchQuery,
+    subscription,
+  }), [activeTab, currentPage, hasMultipleSubscriptions, overview, searchQuery]);
```
binodpant, line 36: not clear if subscription should be in this list ...

src/components/subscriptions/SubscriptionDetailPage.jsx
```diff
@@ -0,0 +1,42 @@
...
+import { SubscriptionContext } from './SubscriptionData';
+import SubscriptionDetailContextProvider from './SubscriptionDetailContextProvider';
+import { NotFound } from '../NotFoundPage';
+
+const SubscriptionDetailPage = ({ match }) => {
+  const { params: { subscriptionUUID } } = match;
+  const { data: subscriptions } = useContext(SubscriptionContext);
+  if (!subscriptions?.count) {
+    return (
+      <NotFound />
```
binodpant, line 16: unrelated random thought: I have been thinking that at some point we should enhance the messaging of this page or add more similar pages to indicate better what the cause of the 404 page was, but not for this PR!

src/components/subscriptions/SubscriptionManagementPage.jsx
```diff
@@ -1,115 +1,47 @@
...
+            />
+          </Switch>
+        </Container>
+      </main>
+    </SubscriptionData>
   );
 }
 
 SubscriptionManagementPage.propTypes = {
-  enterpriseSlug: PropTypes.string.isRequired,
```
binodpant, line 107: I like this! was not sure earlier when I looked at this code why both were being used here

src/components/subscriptions/SubscriptionZeroStateMessaging.jsx
```diff
@@ -0,0 +1,32 @@
+import React, { useContext } from 'react';
+import { Card } from '@edx/paragon';
+import InviteLearnersButton from './buttons/InviteLearnersButton';
+import { TAB_PENDING_USERS } from './data/constants';
+import { SubscriptionContext } from './SubscriptionData';
+import { ToastsContext } from '../Toasts';
+import { SubscriptionDetailContext } from './SubscriptionDetailContextProvider';
+
+const SubscriptionZeroStateMessaging = () => {
```
binodpant, line 9: just a pref thing I suppose, but perhaps this name can be more component like such as SubscriptionZeroStateCTA or SubscriptionZeroStateMessage ? not too strong a feeling though.

src/components/subscriptions/data/hooks.js
```diff
@@ -0,0 +1,159 @@
...
+    count: 0,
+    next: null,
+    previous: null,
+  });
+
+  const forceRefresh = () => {
+    setSubscriptions({ ...subscriptions });
+  };
+
+  useMemo((page = 1) => {
```
binodpant, line 29: will `page` ever be set/ changed by anyone? if so, where would that be?

src/components/subscriptions/data/hooks.js
```diff
@@ -0,0 +1,159 @@
...
+      .then((response) => {
+        const { data: subscriptionsData } = response;
+        const subscriptionsList = subscriptionsData?.results || [];
+        subscriptionsData.results = subscriptionsList.filter((
+          subscription => subscription.is_active
+        ));
+        setSubscriptions(camelCaseObject(subscriptionsData));
+      })
+      .catch((err) => {
+        NewRelicService.logAPIErrorResponse(err);
```
binodpant, line 40: in general, I thought a good improvement here (again not for this PR!) will be to call a more abstract method like logErrorResponse() or something like it and have it handle NewRelic-ness but we can have that be considered separately

src/components/subscriptions/data/hooks.js
```diff
@@ -0,0 +1,159 @@
...
+            [SUBSCRIPTION_USERS_OVERVIEW]: NETWORK_ERROR_MESSAGE,
+          });
+        });
+    }
+  }, [subscriptionUUID, search]);
+
+  return subscriptionUsersOverview;
+};
+
+/*
```
binodpant, line 98: nit: ideally jsdoc style comments? like 

```
/**
* abc
* def
*/
```

src/components/subscriptions/SubscriptionDetailContextProvider.jsx
chavesj: yes it should! good catch

src/components/subscriptions/data/hooks.js
chavesj: I don't believe it will. Page number was part this hook beforehand and has always been used as a query param in the initial call to license-manager so I left it in there. I agree that it doesn't seem needed though

src/components/subscriptions/data/hooks.js
binodpant: thanks filed https://openedx.atlassian.net/browse/ENT-3852

src/components/subscriptions/data/hooks.js
```diff
@@ -0,0 +1,159 @@
...
+*/
+export const useSubscriptionData = ({ enterpriseId }) => {
+  const [errors, setErrors] = useState({});
+  const {
+    subscriptions,
+    forceRefresh,
+  } = useSubscriptions({ enterpriseId, errors, setErrors });
+
+  return {
+    subscriptions,
```
binodpant, line 154: really nitty: maybe group subscriptions and forceRefresh together? to separate out subscriptions data vs the useState outputs?

binodpant (approved):
Thanks for letting us review this. It may also be helpful/instructive for us, as we are planning to add bulk enrollment features here. Generally looks good to me! I guess I am not the best reviewer right now overall due to my lack of context on some of the code, but I like the general direction and improvements! Just dropped some minor comments.

src/components/subscriptions/buttons/InviteLearnersButton.jsx
```diff
@@ -0,0 +1,41 @@
...
+import InviteLearnersModal from '../../../containers/InviteLearnersModal';
+import ActionButtonWithModal from '../../ActionButtonWithModal';
+import { SubscriptionDetailContext } from '../SubscriptionDetailContextProvider';
+
+const InviteLearnersButton = ({ onSuccess, onClose }) => {
+  const { overview, subscription } = useContext(SubscriptionDetailContext);
+  return (
+    <ActionButtonWithModal
+      buttonLabel="Invite learners"
+      buttonClassName="add-users-btn"
```
adamstankiewicz, line 13: nit: along the lines of renaming AddUsers to InviteLearners, the `buttonClassName` could be `invite-learners-btn` instead.

src/components/subscriptions/data/hooks.js
adamstankiewicz: Just to confirm, does the API get called with the appropriate page number during pagination of the TabContentTable data?

src/components/subscriptions/tests/TestUtilities.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+export const mockUseSubscriptionData = (state) => ({
+  subscriptions: {
+    count: 1,
+    next: null,
+    previous: null,
+    results: [subscriptionPlan(state)],
+  },
+  errors: {},
+  setErrors: () => null,
```
adamstankiewicz, line 108: super nit: usually a no-op function is written as `() => {}`

src/components/subscriptions/tests/licenses/LicenseAllocationDetails.test.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import {
+  beforeEach,
+  describe,
+  jest,
+  test,
+} from '@jest/globals';
```
adamstankiewicz, line 9: Wondering about why `@jest/globals` is needed since they should already be globals?

package.json
```diff
@@ -30,6 +30,7 @@
     "@fortawesome/free-solid-svg-icons": "5.15.1",
     "@fortawesome/react-fontawesome": "0.1.12",
     "@fullstory/browser": "1.4.5",
+    "@jest/globals": "^26.6.2",
```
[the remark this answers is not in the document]
chavesj: my IDE linter was complaining about it but that makes sense given the use of setupTest.js. will remove

iloveagent57 (approved):
This looks good from my manual testing.  Is the bug you mentioned in your description still an issue?  The subscription details seemed to update fine for me when I assigned/revoked a user.

chavesj:
> This looks good from my manual testing. Is the bug you mentioned in your description still an issue? The subscription details seemed to update fine for me when I assigned/revoked a user.

Yes it is. What you should be seeing is the table contents are successfully updated but the overview data (number of learners for each license status) displayed in the tabs on the left is not updated. Similarly, the "X of Y licenses allocated" text is not appropriately updated after successful action. That's because the user data hook has `activeTab` as a dependency and the tab is set to revoked/assigned after either of those actions are successfully performed. The overview and license data hooks don't. I'll update the bug description above to be more specific.

adamstankiewicz (approved):

src/components/subscriptions/data/hooks.js
chavesj: I'm inclined to keep this as is and avoid adding another level of nesting

src/components/subscriptions/SubscriptionDetailPage.jsx
```diff
@@ -0,0 +1,42 @@
...
+import SubscriptionDetails from './SubscriptionDetails';
+import LicenseAllocationDetails from './licenses/LicenseAllocationDetails';
+import { SubscriptionContext } from './SubscriptionData';
+import SubscriptionDetailContextProvider from './SubscriptionDetailContextProvider';
+import { NotFound } from '../NotFoundPage';
+
+const SubscriptionDetailPage = ({ match }) => {
+  const { params: { subscriptionUUID } } = match;
+  const { data: subscriptions } = useContext(SubscriptionContext);
+  if (!subscriptions?.count) {
```
kellmarie, line 14: I'd move line 19 and 20 to be with the other consts and update this if statement to also check for subscription. 
`if (!subscriptions?.count || !subscription) {`

Currently when an admin who has subscriptions tries to load one that doesn't exist, like a mistyped uuid, the NotFound page doesn't display and the SubscriptionDetailContextProvider gets returned and it causes a whole bunch of errors and the entire page goes blank.

src/components/subscriptions/data/hooks.js
kellmarie: I confirmed that it does

kellmarie (approved):

src/components/subscriptions/SubscriptionDetailPage.jsx
chavesj: good catch!

chavesj: I'm also relocating this logic to be inside `SubscriptionDetailContextProvider`, there was no need to have it split between the parent component and context provider

chavesj: edit: apparently there is a good reason to return the 404 from the parent component - React doesn't like it when a component conditionally calls hooks. `hasMultipleSubscriptions` ended up being the only part of this moved into the context provider

adamstankiewicz: Ah, eyes. That is one of the rules of React hooks, that you can't conditionally call hooks as they much execute in the exact same order on every render of the component!

PR #399, opened 2020-12-03
fix column widths for new theme

adamstankiewicz:
image


raq929 (approved):
Fix whatever is failing on CI (probably snapshots?) and you're good to go. :)

PR #400, opened 2020-12-03
feat: ENT-3725 : barebones url feature switched bulk enrollment page

binodpant:
In scope:
  A page that simply fetches all courses for this enterprise is subscribed to and lists them in some simple manner. without attempting to render correctly. course_run and pacing type are not yet available in algolia so will ticket that out

Future work here will be to make courses filterable and selectable so we can enroll students in them

As of this PR, the feature switch will be used to control if the bulk enrollment tab / page can be used. We will remove the feature flag or make it true by default once we are ready with the features


binodpant:
zero tests right now... I think it makes sense to test this once we are using the filtering and display logic... but open to suggestions

.gitignore
```diff
@@ -10,4 +11,4 @@ dist/
 
 # emacs
 *~
-*#
\ No newline at end of file
+*#
```
raq929, line 14: uh, what's going on here?

.nvmrc
```diff
@@ -0,0 +1,2 @@
+12.10.0
```
raq929, line 1: 🎉

src/components/Sidebar/IconLink.jsx
```diff
@@ -2,36 +2,28 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { NavLink } from 'react-router-dom';
 import classNames from 'classnames';
-import { Icon } from '@edx/paragon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 const IconLink = props => (
```
raq929, line 7: nit: I have a slight preference for object destructuring, so instead of props, 
`({ isExpanded, icon, title, to })`

But, you didn't write this originally, so nbd.

src/components/Sidebar/index.jsx
```diff
@@ -2,6 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import { faFile, faIdCard, faLifeRing } from '@fortawesome/free-regular-svg-icons';
```
raq929, line 5: Yes, I definitely like importing these rather than using strings.

src/index.jsx
```diff
@@ -32,6 +36,9 @@ import { configuration } from './config';
 
 import './index.scss';
 
+// font awesome
+library.add(fab, faBookOpen);
```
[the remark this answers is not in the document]
binodpant: oh wait, now that I am importing directly I don't need this sorry. this is useful if we don't want to import those in every file and want to use strings, removing this

src/components/Sidebar/IconLink.jsx
binodpant: good point let me change this as well!

.gitignore
binodpant: hmm no idea, will remove it

PR #405, opened 2020-12-04
remove small subtitles from modals

adamstankiewicz:


raq929 (approved):
Looks good code-wise. I'm a little worried because I don't know what the titles say that it won't be clear to people what they are doing?
But if you think that's clear enough without the subtitles, you're good to go.

PR #406, opened 2020-12-04
Add zero state for subscription with no allocations

kellmarie:
### **Changes Overview:**
Add getting started messaging for when admins view All Users on a subscription that has no allocated licenses on it.

### **Messaging with wide screen:**
Screen Shot 2020-12-08 at 10 09 36 PM

### **Messaging on a smaller screen:**
Screen Shot 2020-12-08 at 10 09 43 PM

### **Messaging on a tiny screen:**
Screen Shot 2020-12-08 at 10 09 51 PM

### **Tabs that aren't All Users remain unchanged from how they currently are:**
Screen Shot 2020-12-08 at 10 10 08 PM

### DIFF
--- a/src/components/subscriptions/LicenseAllocationNavigation.jsx
+++ b/src/components/subscriptions/LicenseAllocationNavigation.jsx
@@ -52,6 +52,7 @@ export default function LicenseAllocationNavigation() {
         {tabs.map(tab => (
           <li key={tab.key}>
             <button
+              id={`navigation-${tab.key}`}
               className={classNames(
                 'btn btn-link btn-block pl-0 text-left',
                 { 'font-weight-bold': activeTab === tab.key },
--- a/src/components/subscriptions/SubscriptionManagementPage.test.jsx
+++ b/src/components/subscriptions/SubscriptionManagementPage.test.jsx
@@ -1,0 +1,138 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { createMemoryHistory } from 'history';
+import { Router, Route } from 'react-router-dom';
+import { Provider } from 'react-redux';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import * as hooks from './hooks/licenseManagerHooks';
+import { ToastsContext } from '../Toasts';
+import {
+  ASSIGNED, TAB_LICENSED_USERS, TAB_PENDING_USERS, TAB_REVOKED_USERS,
+} from './constants';
+import SubscriptionManagementPage from './SubscriptionManagementPage';
+
+const mockStore = configureMockStore([thunk]);
+const store = mockStore({
+  authentication: {
+    username: 'edx',
+    roles: ['enterprise_admin:*'],
+  },
+  userAccount: {
+    loaded: true,
+    isActive: true,
+  },
+  portalConfiguration: {
+    enterpriseSlug: 'test-enterprise',
+    enterpriseId: 'test-enterprise-id',
+    enableSubscriptionManagementScreen: true,
+  },
+});
+
+const initialHistory = createMemoryHistory({
+  initialEntries: ['/'],
+});
+
+/* eslint-disable react/prop-types */
+const ManagementPageWithContext = () => (
+  <Router history={initialHistory}>
+    <Provider store={store}>
+      <ToastsContext.Provider value={{ addToast: () => {} }}>
+        <Route
+          exact
+          path="/test-enterprise/admin/subscriptions"
+        >
+          <SubscriptionManagementPage
+            enterpriseSlug="test-enterprise"
+            enterpriseId="a test id"
+          />
+        </Route>
+      </ToastsContext.Provider>
+    </Provider>
+  </Router>
+);
+
+/* eslint-enable react/prop-types */
+
+describe('SubscriptionManagementPage', () => {
+  let mockLoadSubscriptionData = {};
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockLoadSubscriptionData = {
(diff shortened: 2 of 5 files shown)


src/components/subscriptions/LicenseAllocationNavigation.jsx
```diff
@@ -48,22 +48,25 @@ export default function LicenseAllocationNavigation() {
...
-              )}
-              onClick={() => updateTabWithDefaultPage(tab.key)}
-              type="button"
-            >
-              {tab.text}
-            </button>
-          </li>
-        ))}
-      </ul>
+      {overview.assigned + overview.activated + overview.revoked > 0
```
adamstankiewicz, line 51: nit: maybe do an early return instead of adding this as a conditional.

```jsx

const overviewCount = overview.assigned + overview.activated + overview.revoked;
if (!overviewCount) {
  return null;
)

return (
  <nav>
    <ul className="list-unstyled w-100">
      {tabs.map(tab => (
        // ...
      ))}
  </nav>
);

```

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -131,49 +117,84 @@ function TabContentTable({ enterpriseSlug }) {
...
-              </div>
-              <div className="mt-3 d-flex justify-content-center">
-                <Pagination
-                  onPageSelect={page => fetchSubscriptionUsers({ searchQuery, page })}
-                  pageCount={users.numPages}
-                  currentPage={currentPage}
-                  paginationLabel={activeTabData.paginationLabel}
-                />
+          {activeTab === TAB_ALL_USERS && tableData?.length === 0 ? (
+            <div className="container py-4 col-md-5 border shadow">
```
adamstankiewicz, line 121: I'm not sure `.container` is needed here. Also, fwiw, with Bootstrap, you typically would nest a column within a row.

This example uses a centered column in a Bootstrap row:

```html
<div class="row">
    <div class="col-6 offset-md-3">
        This is horizontally centered.
    </div>
</div>
```

Also, Paragon now exports some layout components from react-bootstrap for rows and columns, too if thats more convenient:

```jsx
<Row>
  <Col xs={12} md={{ span: 6, offset: 3 }}>
    This is horizontally centered.
  </Col>
</Row>
```

As is, you have a container that is also being treated as a column, and then columns nested inside of columns without a parent row in between first.

src/components/subscriptions/SubscriptionManagementPage.test.jsx
```diff
@@ -0,0 +1,140 @@
...
+        subscriptionUsers: {
+          numPages: 1, next: null, previous: null, results: [],
+        },
+      },
+      isLoading: false,
+      hasSubscription: true,
+    };
+  });
+
+  test('dispalys the zero state on the default All Users tab with no allocations', () => {
```
adamstankiewicz, line 101: nit: "displays"

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -101,6 +104,27 @@ function TabContentTable({ enterpriseSlug }) {
     [users],
   );
 
+  const SubscriptionZeroStateMessaging = () => (
+    <div className="py-4 border shadow">
```
adamstankiewicz, line 108: Could we use the `Card` component here in place of explicitly using these padding, border, and shadow classes?

```jsx
<Card className="text-center">
  <Card.Body>
    <h2>Get Started</h2>
    <p className="py-2 lead">Assign your learners...</p>
    <AddUsersButton />
  </Card.Body>
</Card>
```

src/components/subscriptions/TabContentTable.jsx
```diff
@@ -19,6 +21,7 @@ import {
   TAB_PENDING_USERS,
   TAB_REVOKED_USERS,
 } from './constants';
+import AddUsersButton from './AddUsersButton';
```
adamstankiewicz, line 24: Per the mocks, we should rename `AddUsersButton` to `InviteLearnersButton` and change the button text to "Invite learners" from "Add Users". This applies to both places where this component is now used.

iloveagent57 (approved):
LGTM

src/components/subscriptions/SubscriptionManagementPage.test.jsx
```diff
@@ -0,0 +1,140 @@
...
+    enterpriseSlug: 'test-enterprise',
+    enterpriseId: 'test-enterprise-id',
+    enableCodeManagementScreen: true,
+    enableSubscriptionManagementScreen: true,
+    enableAnalyticsScreen: true,
+  },
+});
+
+const initialHistory = createMemoryHistory({
+  initialEntries: ['/test-enterprise/admin/register/activate'],
```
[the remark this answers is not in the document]
kellmarie: Copy pasta, I was trying to figure out what the UserActivationPage.test.jsx was doing that was making the multiple providers and router stuff working. I'll remove it, unless it breaks things.

src/components/subscriptions/SubscriptionManagementPage.test.jsx
```diff
@@ -0,0 +1,140 @@
...
+});
+
+/* eslint-disable react/prop-types */
+const ManagementPageWithContext = () => (
+  <Router history={initialHistory}>
```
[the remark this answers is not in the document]
kellmarie: It doesn't really help that much. It doesn't return an enzyme wrapper so I'd have to figure out how to do all of the assertions and how to do click simulating in the test. The query options that `screen` gives me doesn't seem to work for finding `SubscriptionZeroStateMessaging`

adamstankiewicz (approved):

PR #413, opened 2020-12-10
fix capitalization on invite learners button and make sure the zero s…

kellmarie:
…tate messaging doesn't display when a search is made


adamstankiewicz (approved):
:shipit: Cool, thanks for addressing this! 🎉

PR #415, opened 2020-12-11
Add Success Factor to LMS Configurations

taliaedX:
TODO:
- Tests
- Verify functionality (update in particular)
- Add a feature gate (maybe?)

### DIFF
--- a/src/components/LmsConfigurations/index.test.jsx
+++ b/src/components/LmsConfigurations/index.test.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { REQUIRED_MOODLE_CONFIG_FIELDS } from './MoodleIntegrationConfigForm';
 import { REQUIRED_CANVAS_CONFIG_FIELDS } from './CanvasIntegrationConfigForm';
+import { REQUIRED_SUCCESS_FACTOR_CONFIG_FIELDS } from './SuccessFactorsIntegrationConfigForm';
 import LmsConfigurations from './index';
 import LmsApiService from '../../data/services/LmsApiService';
 import { REQUIRED_BLACKBOARD_CONFIG_FIELDS } from './BlackboardIntegrationConfigForm';
@@ -11,10 +12,22 @@ jest.mock('./../../data/services/LmsApiService');
 const moodleResponse = {
   data: { results: [{}] },
 };
+const successFactorsResponse = {
+  data: { results: [{}] },
+};
+
 const notFoundMoodleResponse = {
   request: { status: 404, statusText: 'Not Found' },
 };
+const notFoundSuccessFactorsResponse = {
+  request: { status: 404, statusText: 'Not Found' },
+};
+
 const badMoodleResponse = {
+  request: { status: 400, statusText: 'Bad Request' },
+};
+
+const badSuccessFactorsResponse = {
   request: { status: 400, statusText: 'Bad Request' },
 };
 
@@ -23,6 +36,10 @@ const canvasResponse = {
 };
 const notFoundCanvasResponse = {
   request: { status: 404, statusText: 'Not Found' },
+};
+
+const badCanvasResponse = {
+  request: { status: 400, statustext: 'Bad Request' },
 };
 
 const blackboardResponse = {
@@ -35,32 +52,44 @@ const badBlackboardResponse = {
   request: { status: 400, statusText: 'Bad Request' },
 };
 
-const formData = new FormData();
-
 REQUIRED_MOODLE_CONFIG_FIELDS.forEach((field) => {
-  formData.append(field, 'testdata');
   moodleResponse.data.results[0][field] = 'testdata';
 });
 
 REQUIRED_CANVAS_CONFIG_FIELDS.forEach((field) => {
-  formData.append(field, 'testdata');
   canvasResponse.data.results[0][field] = 'testdata';
 });
 
 REQUIRED_BLACKBOARD_CONFIG_FIELDS.forEach((field) => {
-  formData.append(field, 'testdata');
   blackboardResponse.data.results[0][field] = 'testdata';
+});
+
+REQUIRED_SUCCESS_FACTOR_CONFIG_FIELDS.forEach((field) => {
+  successFactorsResponse.data.results[0][field] = 'testdata';
 });
 
 const waitForAsync = () => new Promise(resolve => setImmediate(resolve));
 
 describe('<LmsConfigurations /> ', () => {
+  it('get list of forms when all configs return 404s', async () => {
+    LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.find('ErrorPage').length).toBe(0);
+    expect(wrapper.find('CollapsibleAdvanced').length).toBeGreaterThan(0);
+  });
   it('get the Moodle LMS configuration when present', async () => {
     LmsApiService.fetchMoodleConfig.mockResolvedValue(moodleResponse);
 
     // Mock the fetch of the other configs to 404, not found
     LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
     LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
 
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
@@ -75,6 +104,7 @@ describe('<LmsConfigurations /> ', () => {
     // Mock the fetch of the other configs to 404, not found
     LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
     LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
 
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
@@ -83,21 +113,51 @@ describe('<LmsConfigurations /> ', () => {
     expect(wrapper.find('BlackboardIntegrationConfigForm').props().config).toEqual(wrapper.state().blackboardConfig);
   });
 
+  it('get the Canvas LMS configuration when present', async () => {
+    LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockResolvedValue(canvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.state().canvasConfig).toEqual(canvasResponse.data.results[0]);
+    expect(wrapper.find('CanvasIntegrationConfigForm').props().config).toEqual(wrapper.state().canvasConfig);
+  });
+
+  it('get the Success Factors LMS configuration when present', async () => {
+    LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockResolvedValue(successFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.state().sapsfConfig).toEqual(successFactorsResponse.data.results[0]);
+    expect(wrapper.find('SuccessFactorsIntegrationConfigForm').props().config).toEqual(wrapper.state().sapsfConfig);
+  });
+
   it('return Error Page when the Moodle LMS fetch request fails with error (not 404)', async () => {
     LmsApiService.fetchMoodleConfig.mockRejectedValue(badMoodleResponse);
 
     // We want to ensure that the Moodle fetch config method will cause the error page to display
     // if it encounters an unexpected error while the other config fetch methods succeed.
     LmsApiService.fetchBlackboardConfig.mockResolvedValue(blackboardResponse);
-    LmsApiService.fetchCanvasConfig.mockResolvedValue(canvasResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
 
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
     wrapper.update();
 
-    // Expect the moodle config to be false, while the others to be true
     expect(wrapper.state().moodleConfig).toBeFalsy();
     expect(wrapper.find('ErrorPage').length).toBe(1);
+    expect(wrapper.find('MoodleIntegrationConfigForm').length).toBe(0);
+    expect(wrapper.find('SuccessFactorsIntegrationConfigForm').length).toBe(0);
+    expect(wrapper.find('CanvasIntegrationConfigForm').length).toBe(0);
+    expect(wrapper.find('BlackboardIntegrationConfigForm').length).toBe(0);
   });
 
   it('return Error Page when Blackboard LMS fetch request fails with an error (not 404)', async () => {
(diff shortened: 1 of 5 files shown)


src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
```diff
@@ -0,0 +1,374 @@
...
+   * Returns if there is an error.
+   * @param {FormData} formData
+   */
+  createSuccessFactorsConfig = async (formData) => {
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('enterprise_customer', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.postNewSuccessFactorsConfig(transformedData);
+      this.setState({ config: response.data });
+      return undefined;
```
raq929, line 49: The linter made you do this? I'd prefer to return the previous line, I guess?

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
```diff
@@ -0,0 +1,374 @@
...
+    }
+  }
+
+  updateSuccessFactorsConfig = async (formData, configId) => {
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('enterprise_customer', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.updateSuccessFactorsConfig(transformedData, configId);
+      this.setState({ config: response.data });
+      return undefined;
```
raq929, line 61: Same as above.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
```diff
@@ -0,0 +1,374 @@
...
+              }}
+              disabledStates={[SUBMIT_STATES.PENDING]}
+              variant="primary"
+              className="ml-3 col"
+            />
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-3 mt-3">
+            {errorAlert}
```
raq929, line 347: A more "React-y" way to do this would be to have this line be
```
{error && (<div className="form-group is-invalid align-items-left">
          <StatusAlert
            alertType="danger"
            iconClassName="fa fa-times-circle"
            title="Unable to submit config form:"
            message={error}
          />
        </div>)
        }
        ```

Comment (raq929) [reply src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx:347]: (But formatted correctly)

File: src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx [hunk 4]
```diff
@@ -0,0 +1,374 @@
...
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="secret"
+              invalid={invalidFields.key}
+              invalidMessage="Success Factors' Client Secret is required."
+              helpText="OAuth client secret."
+            >
+              <label htmlFor="secret">Client Secret</label>
+              <Input
+                type="text"
```
raq929, line 237: Should this be "password" instead of text, if it's a secret?

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
```diff
@@ -0,0 +1,374 @@
...
+export const REQUIRED_SUCCESS_FACTOR_CONFIG_FIELDS = [
+  'sapsfBaseUrl',
+  'sapsfCompanyId',
+  'sapsfUserId',
+  'key',
+  'secret',
+  'userType',
+];
+
+class SuccessFactorsIntegrationConfigForm extends React.Component {
```
raq929, line 22: Don't bother with this one because it's not worth your time to refactor, but we're moving away from class components on the frontend in favor of function components, so future work should be functional React components. Happy to chat about what that looks like next time you pick up a frontend ticket!

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
```diff
@@ -0,0 +1,374 @@
...
+            iconClassName="fa fa-times-circle"
+            title="Unable to submit config form:"
+            message={error}
+          />
+        </div>
+      );
+    }
+
+    return (
+      <form
```
raq929, line 148: I'm not sure from this PR exactly how the component is used, but if it's basically the whole page, then it needs a wrapper to give it the proper padding / margins etc that the other pages have. (Ignore me if that's already done!)

raq929 (changes requested):
I'd love to see some tests added to this PR, basically:

- It shows the form with correct defaults on load
- the proper function is called on submit
- errors are properly displayed if they are received

Happy to chat about frontend testing with you!

raq929:
Ha, sorry I didn't see that this was WIP and you already had plans for testing. 🎉

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: "return the previous line" ? 
Can you return setting state?

And yes the linter made me do this. It didn't like `return` nor did it like not having a return at all.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: yeah I've done purely functional components before. This style was also already more functional to me in comparison to previous class component work I've done at other places (there is no constructor at all here) so I wasn't sure if this constructor-less class component was the preferred style. Guess not!

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: I don't know what you mean full page. There is separate header, sidebar, and footer components that are always there. If there is some margin-ing to do for the sidebar components that I haven't done here, definitely let me know. 

It might be obvious, but most of what I have done is modeled off of the Saml Config forms which was modeled off the ReportingConfig form. I tried to go with consistency when first implementing this stuff how many months ago now? If we have a standard I can look at (or a form that has already been updated to match some standard) I'm more than happy to update these components to match. I just know me and 99.9% of other devs not focused on front end are just gonna copy and paste stuff that seems to work without paying attention to style updates/preferences (I think I recall there being some bug or something because the reporting config didn't match something in the rebrand and thus all these other forms that were based on that also don't match)

taliaedX: Sorry the TL;DR is: I don't know what standard I'm missing. If there is 1 (or more) sidebar components that have been updated in some consistent manner, definitely point me at them and I'll make sure this (and all the other LMS configs) match. I just didn't find consistency in how they were implemented 5+ months ago and have been just been basing off of one mostly relevant form (Reporting Config), so they may be currently consistently wrong.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: TIL that was an input type.
Does that affect anything besides masking the field? Just curious because we don't actually want it encoded/hidden in general.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
raq929: yea, `this.setState` probably also returns undefined, but that seems nicer to me somehow? Not critical, though.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
raq929: It just masks the field. Basically it only protects against over-the-shoulder and screen capture type attacks. It won't encode the input in anyway, your form should pick it up as a regular form field.

raq929: Eventually I want a Paragon component that allows the user to show or hide a field like this, but that's a dream for the future.... :)

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
raq929: Yeah, no worries, no way you could have known; just wanted to pass info along for the future.
Personally, I think that even though React says they'll never get rid of class components, they are eventually going to ditch support for class components.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
raq929: No worries. Can you throw a screenshot of this page up? I'm not even 100% sure this is still an issue (I noticed it in the demo, but I've got devstack shut down right now for Paragon work.) 
I'm happy to open it up as a small follow-on ticket if the margin is looking off.

raq929:
To be clear, most of these comments are nits, the only thing blocking is tests. 😄

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: What do you think about me throwing in a ticket to refactor this repo to use functional components?
I think it's totally fine to want to move to something we decide is better stylistically/programmatically, I just also want us to standardize. From a glance it's very hard to gather what "way" we want our frontend components done since they all end up different lol.

I'd love to tackle this all at once.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: Yeah I mean if you saw an issue in the demo, I definitely haven't done anything with extra divs/css or Paragon itself that would fix it lol, but I've got some screenshots here:
Screen Shot 2020-12-16 at 9 57 07 AM
Screen Shot 2020-12-16 at 9 57 16 AM

raq929: Yeah, it needs a bit more margin. I'm gonna open a ticket on it, because it affects both your and @alex-sheehan-edx PRs.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
raq929: Honestly, I don't think refactoring to functional components is *ever* worth the time it takes. For new components, I think we should write them as functional components. But it's a different enough way of thinking about things that in general I don't think it's worth our time until React's inevitable betrayal.
My take is, if the class component works, use it.
The refactor I *do* want to open a ticket for is a refactor to share the error handling and other code that is re-used between this and Alex's PR. And that's a good opportunity to move the form part of this to a functional component.

src/components/LmsConfigurations/SuccessFactorsConfigForm.jsx
taliaedX: Thank you!

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -35,32 +52,44 @@ const badBlackboardResponse = {
...
+    LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.find('ErrorPage').length).toBe(0);
+    expect(wrapper.find('CollapsibleAdvanced').length).toBeGreaterThan(0);
```
raq929, line 84: I feel like this isn't testing much? It could be displaying multiple copies of each form and we wouldn't know. I'd prefer a test that proves that each form we expect to be there is actually there.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -83,21 +113,48 @@ describe('<LmsConfigurations /> ', () => {
...
+    LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockResolvedValue(canvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.state().canvasConfig).toEqual(canvasResponse.data.results[0]);
+    expect(wrapper.find('CanvasIntegrationConfigForm').props().config).toEqual(wrapper.state().canvasConfig);
```
raq929, line 126: I'd also love to add a test to show that the config data is actually being displayed on the page rather than just in the state. 
I would do something like:
`expect(wrapper.find('CanvasIntegrationConfigForm').text()).toContain(<some string from your mock response>)`

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -83,21 +113,48 @@ describe('<LmsConfigurations /> ', () => {
...
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockResolvedValue(successFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+    expect(wrapper.state().sapsfConfig).toEqual(successFactorsResponse.data.results[0]);
+    expect(wrapper.find('SuccessFactorsIntegrationConfigForm').props().config).toEqual(wrapper.state().sapsfConfig);
+  });
```
raq929, line 140: Same as above; I'd love a test that confirms what the user is seeing is correct.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -83,21 +113,48 @@ describe('<LmsConfigurations /> ', () => {
...
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
 
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
     wrapper.update();
 
-    // Expect the moodle config to be false, while the others to be true
     expect(wrapper.state().moodleConfig).toBeFalsy();
     expect(wrapper.find('ErrorPage').length).toBe(1);
+    expect(wrapper.find('MoodleIntegrationConfigForm').length).toBe(0);
```
raq929, line 157: What are we expecting the canvas and SuccessFactors forms to look like here? Are they supposed to be present, or is it just an error page? Can we add expects for that?

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -107,25 +164,44 @@ describe('<LmsConfigurations /> ', () => {
...
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
 
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
     wrapper.update();
 
-    // Expect the moodle config to be false, while the others to be true
     expect(wrapper.state().blackboardConfig).toBeFalsy();
     expect(wrapper.find('ErrorPage').length).toBe(1);
+    expect(wrapper.find('BlackboardIntegrationConfigForm').length).toBe(0);
```
raq929, line 175: Same as above.

src/components/LmsConfigurations/index.test.jsx
raq929: (Perhaps that should be a different test/tests)

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -107,25 +164,44 @@ describe('<LmsConfigurations /> ', () => {
...
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(badCanvasResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
+
+    const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
+    await waitForAsync();
+    wrapper.update();
+
+    expect(wrapper.state().canvasConfig).toBeFalsy();
+    expect(wrapper.find('ErrorPage').length).toBe(1);
+    expect(wrapper.find('CanvasIntegrationConfigForm').length).toBe(0);
```
raq929, line 190: Same as above.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -107,25 +164,44 @@ describe('<LmsConfigurations /> ', () => {
...
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
     wrapper.update();
-    expect(wrapper.state().moodleConfig).toBeFalsy();
-    expect(wrapper.find('ErrorPage').length).toBe(0);
-    expect(wrapper.find('CollapsibleAdvanced').length).toBe(3);
+
+    expect(wrapper.state().sapsfConfig).toBeFalsy();
+    expect(wrapper.find('ErrorPage').length).toBe(1);
+    expect(wrapper.find('SuccessFactorsIntegrationConfigForm').length).toBe(0);
```
raq929, line 205: Same as above.

raq929 (commented):
Looks good in general; I'm wanting tests to be testing the UI that the user is seeing, not just internal state, and it looks like there's some state that's not being tested. Let me know if my comments are unclear. Happy to jump on a call if that's helpful.

src/components/LmsConfigurations/index.test.jsx
taliaedX: That's a fair assessment. The goal was just to verify the logic for displaying the partial/collapsible forms was working as expected (ie when we don't have a config present). I felt like using an exact amount might be incredibly tedious and annoying as the test won't be able to be dynamic and every time we add a new form to the page, this count would have to be manually updated (maybe there is a better way to do this, I'm unsure honestly :()

Finding the underlying components could work. (ie in this particular test we should see every form). I think I avoided that initially since at least 1 underlying form should be present in both cases (ie if we have a SuccessFactor config we should see the SuccessFactors component, but if we have no config, we should also see the form)

I didn't even think of a possibility of multiples of the same form to be present.

src/components/LmsConfigurations/index.test.jsx
taliaedX: I feel like that's already covered by the component tests. Ie if you look at the CanvasIntegrationConfigForm tests they will comprehensively cover the fields therein and making sure they auto-populate based on props/state.

The goal of this test is just to make sure props are being passed. Like the other test above using `CollapsibleAdvanced`, the differentiator in the logic here isn't exclusively/necessarily around which forms we display, but also what data we pass through to it.

taliaedX: We may want a full end-to-end test somewhere, but I don't feel this test suite is the place to put it.
Logic-wise if we verify the props are getting sent from the index and we thoroughly test the form data/passed prop logic in the components it uses then we have 100% coverage. There is certainly an argument that can be made about an end-to-end test that tests that flow truly and completely working, but similar to your overall PR comment, I'd say that's a topic for debate and addressing at a later time.

src/components/LmsConfigurations/index.test.jsx
taliaedX: We're not expecting them that's why the length is 0. I was playing around with the logic here to get this expect logic to work in the intended way which is: "I should not see a form here. I should see an error page and only an error page"

taliaedX: I think I struggled getting any sort of "exists" logic instead of this length usage.

src/components/LmsConfigurations/index.test.jsx
raq929: Yeah, I think a test that confirms each form we want to be there is there would be great, that way, if one of them breaks, we'll know.

src/components/LmsConfigurations/index.test.jsx
raq929: I don't consider testing the UI to be an end-to-end test (though I agree that is also useful). In general, the React community is moving **strongly** away from testing what's in state or props and instead testing what's in the UI, and I do think we should be testing what the user sees. So, yes, it's an integration test rather than a unit test, but I think the test should exist. I'm OK with us opening a ticket for it.

src/components/LmsConfigurations/index.test.jsx
raq929: This appears to only test that the length of the Moodle form is zero. Could we add some expects to test that the Canvas and SuccessFactors forms are also not present, despite positive API results?

src/components/LmsConfigurations/index.test.jsx
taliaedX: ah I completely forgot that type of test ("integration"). We can open a ticket as I think we should add these for all sidebar items that don't already have them. If you happen to have - and don't feel obligated to go searching - some links to standards being discussed or the community talking about the testing strategies here I'd love to take a look. I've used snapshots as well as trying to dig into the DOM to do actual integration tests and I've found them really flaky to the point of being useless, however React moves fast, so there may be much better solutions now.

raq929:
Approving provided there's a ticket opened for improving testing.

raq929 (approved):

taliaedX:
https://openedx.atlassian.net/browse/ENT-3908

is the ticket. Will need to groom a bit there.

PR #416, opened 2020-12-14
add temporary maintenance alert to Learner Progress Report

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-3850

Adds a temporary alert banner to inform Enterprise Admins of planned maintenance on the data analytics api which means the data for the "Learner Progress Report" page won't load.

We are OK with the rest of the page showing a never-ending loading state and an error alert at the bottom during this maintenance period so long as the maintenance messaging is visible. This will be removed after the maintenance period is over.

It's behind a configuration setting so it won't show by default should there be any Open edX consumers of admin portal.

image


.env
```diff
@@ -1,24 +1,25 @@
 NODE_ENV='production'
-BASE_URL=null
```
adamstankiewicz, line 2: If you provide any value other than an empty string (`''`), it parses as a string when reading the `process.env` variable. Null and falsey values have been changed to an empty string.

src/components/Sidebar/index.jsx
```diff
@@ -55,7 +55,7 @@ class Sidebar extends React.Component {
 
     return [
       {
-        title: 'Learner Report',
+        title: 'Learner Progress Report',
```
adamstankiewicz, line 58: making this consistent with the Hero banner page title.

src/config/index.js
```diff
@@ -34,6 +34,7 @@ const configuration = {
   LOGO_URL: process.env.LOGO_URL,
   LOGO_WHITE_URL: process.env.LOGO_WHITE_URL,
   LOGO_TRADEMARK_URL: process.env.LOGO_TRADEMARK_URL,
+  SHOW_LEARNER_REPORT_MAINTENANCE_ALERT: process.env.SHOW_LEARNER_REPORT_MAINTENANCE_ALERT,
```
adamstankiewicz, line 37: https://github.com/edx/edx-internal/pull/3939

georgebabey (approved):

PR #417, opened 2020-12-14
Adding frontend configuration for blackboard

alex-sheehan-edx:
Modeled after existing Moodle and Canvas work.

### DIFF
--- a/src/components/LmsConfigurations/BlackboardIntegrationConfigForm.jsx
+++ b/src/components/LmsConfigurations/BlackboardIntegrationConfigForm.jsx
@@ -1,0 +1,290 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import isEmpty from 'lodash/isEmpty';
+import {
+  ValidationFormGroup, Input, StatefulButton, Icon,
+} from '@edx/paragon';
+import { snakeCaseFormData } from '../../utils';
+import LmsApiService from '../../data/services/LmsApiService';
+import StatusAlert from '../StatusAlert';
+import NewRelicService from '../../data/services/NewRelicService';
+import SUBMIT_STATES from '../../data/constants/formSubmissions';
+
+export const REQUIRED_BLACKBOARD_CONFIG_FIELDS = [
+  'blackboardBaseUrl',
+  'blackboardClientId',
+  'blackboardClientSecret',
+];
+
+class BlackboardIntegrationConfigForm extends React.Component {
+  state = {
+    invalidFields: {},
+    submitState: SUBMIT_STATES.DEFAULT,
+    active: this.props.config?.active,
+    error: null,
+  }
+
+  handleErrors = (error) => {
+    const errorMsg = error.message || error.response?.status === 500
+      ? error.message : JSON.stringify(error.response.data);
+    NewRelicService.logAPIErrorResponse(errorMsg);
+    return errorMsg;
+  }
+
+  /**
+   * Creates a new third party provider configuration, then updates this list with the response.
+   * Returns if there is an error.
+   * @param {FormData} formData
+   */
+  createBlackboardConfig = async (formData) => {
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('enterprise_customer', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.postNewBlackboardConfig(transformedData);
+      this.setState({ config: response.data });
+      return undefined;
+    } catch (error) {
+      return this.handleErrors(error);
+    }
+  }
+
+  updateBlackboardConfig = async (formData, configId) => {
+    const transformedData = snakeCaseFormData(formData);
+    transformedData.append('enterprise_customer', this.props.enterpriseId);
+    try {
+      const response = await LmsApiService.updateBlackboardConfig(transformedData, configId);
+      this.setState({ config: response.data });
+      return undefined;
+    } catch (error) {
+      return this.handleErrors(error);
+    }
+  }
+
+  /**
+   * Validates this form. If the form is invalid, it will return the fields
+   * that were invalid. Otherwise, it will return an empty object.
+   * @param {FormData} formData
+   * @param {[String]} requiredFields
+   */
+  validateBlackboardConfigForm = (formData, requiredFields) => {
+    const invalidFields = requiredFields
+      .filter(field => !formData.get(field))
+      .reduce((prevFields, currField) => ({ ...prevFields, [currField]: true }), {});
+    return invalidFields;
+  }
+
+  /**
+   * attempt to submit the form data and show any error states or invalid fields.
+   * @param {FormData} formData
+   */
+  handleSubmit = async (formData, config) => {
+    this.setState({ submitState: SUBMIT_STATES.PENDING, error: null });
+    let requiredFields = [];
+    requiredFields = [...REQUIRED_BLACKBOARD_CONFIG_FIELDS];
+    // validate the form
+    const invalidFields = this.validateBlackboardConfigForm(formData, requiredFields);
+    if (!isEmpty(invalidFields)) {
+      this.setState({
+        invalidFields: {
+          ...invalidFields,
+        },
+        submitState: SUBMIT_STATES.default,
+      });
+      return;
+    }
+
+    if (config) {
+      const err = await this.updateBlackboardConfig(formData, config.id);
+      if (err) {
+        this.setState({
+          submitState: SUBMIT_STATES.ERROR,
+          error: err,
+        });
+      } else {
+        this.setState({ submitState: SUBMIT_STATES.COMPLETE });
+      }
+    } else {
+      // ...or create a new configuration
+      const err = await this.createBlackboardConfig(formData);
+      if (err) {
+        this.setState({
+          submitState: SUBMIT_STATES.ERROR,
+          error: err,
+        });
+      } else {
+        this.setState({ submitState: SUBMIT_STATES.COMPLETE });
+      }
+    }
+  }
+
+  render() {
+    const {
+      invalidFields,
+      submitState,
+      active,
+      error,
+    } = this.state;
+    const { config } = this.props;
+    let errorAlert;
+    if (error) {
+      errorAlert = (
+        <div className="form-group is-invalid align-items-left">
+          <StatusAlert
+            alertType="danger"
+            iconClassName="fa fa-times-circle"
+            title="Unable to submit config form:"
+            message={error}
+          />
+        </div>
+      );
+    }
+
+    return (
+      <form
+        onSubmit={(e) => {
+          e.preventDefault();
+          const formData = new FormData(e.target);
+          this.handleSubmit(formData, this.state.config ? this.state.config : config);
+        }}
+        onChange={() => this.setState({ submitState: SUBMIT_STATES.DEFAULT })}
+      >
+        <div className="row">
+          <div className="col col-6">
+            <ValidationFormGroup
+              for="active"
+            >
+              <label htmlFor="active">Active</label>
+              <Input
+                type="checkbox"
+                id="active"
+                name="active"
+                className="ml-3"
+                checked={active}
+                onChange={() => this.setState(prevState => ({ active: !prevState.active }))}
+              />
+            </ValidationFormGroup>
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="blackboardBaseUrl"
+              invalid={invalidFields.blackboardBaseUrl}
+              invalidMessage="Blackboard Instance URL is required."
+              helpText="Your Blackboard instance URL. Make sure to include the protocol (ie https/http)"
+            >
+              <label htmlFor="blackboardBaseUrl">Blackboard Instance URL</label>
+              <Input
+                type="text"
+                id="blackboardBaseUrl"
+                name="blackboardBaseUrl"
+                defaultValue={config ? config.blackboardBaseUrl : null}
+              />
+            </ValidationFormGroup>
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="blackboardClientId"
+              invalid={invalidFields.blackboardClientId}
+              invalidMessage="Blackboard client ID is required."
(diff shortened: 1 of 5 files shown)


src/components/LmsConfigurations/index.jsx
```diff
@@ -74,6 +79,14 @@ class LmsConfigurations extends React.Component {
...
+    if (blackboardConfig) {
+      return (
+        <BlackboardIntegrationConfigForm
+          config={camelCaseObject(blackboardConfig)}
+          enterpriseId={this.props.enterpriseId}
+        />
+      );
+    }
     return (
       <>
```
raq929, line 91: This should likely be a div with the "container" class, or whatever class other pages use to get the correct margin / padding.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -45,6 +55,7 @@ describe('<LmsConfigurations /> ', () => {
   it('get LMS configurations when present', async () => {
     LmsApiService.fetchMoodleConfig.mockResolvedValue(moodleResponse);
     LmsApiService.fetchCanvasConfig.mockResolvedValue(canvasResponse);
+    LmsApiService.fetchBlackboardConfig.mockResolvedValue(blackboardResponse);
```
raq929, line 58: Please move to a separate test, and test that the Blackboard config results are present. The current test seems to only look for the Moodle test results on the actual page.
(Canvas should also be moved to a separate test.)

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -55,6 +66,7 @@ describe('<LmsConfigurations /> ', () => {
   it('return Error Page when any LMS fetch request fails with error (not 404)', async () => {
     LmsApiService.fetchMoodleConfig.mockRejectedValue(badMoodleResponse);
     LmsApiService.fetchCanvasConfig.mockRejectedValue(badCanvasResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(badBlackboardResponse);
```
raq929, line 69: Same as above.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -65,11 +77,12 @@ describe('<LmsConfigurations /> ', () => {
   it('return Fragment/Collapsible components when all have 404', async () => {
     LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
     LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
```
raq929, line 80: Same as above.

src/components/LmsConfigurations/BlackboardIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,284 @@
...
+import NewRelicService from '../../data/services/NewRelicService';
+import SUBMIT_STATES from '../../data/constants/formSubmissions';
+
+export const REQUIRED_BLACKBOARD_CONFIG_FIELDS = [
+  'blackboardBaseUrl',
+  'clientId',
+  'clientSecret',
+];
+
+class BlackboardIntegrationConfigForm extends React.Component {
```
raq929, line 19: Same as I wrote to Talia; do not refactor this component, but for the future: we're moving away from React class components in favor of functional components. All new React code we write should be functional components unless there's a compelling reason not to (In this case, it's very similar to another component.)

raq929 (changes requested):
Please update your tests so that:
1. We are testing one and only one thing at a time
2. You're actually testing blackboard

Please open a ticket to do the same for the canvas tests.

What I'm seeing here is that this code and @taliaedX's code are essentially the same except for the form fields. I assume the Moodle and Canvas code is similar as well. It is *definitely* time to refactor this component.
If error handling, etc is all the same across multiple components, there should be a parent component that handles that and takes the form fields (or the whole form) as children.
I'm happy to sync with you on how to do that, but it's time to get away from our habit of copy-pasting React code.

alex-sheehan-edx:
@raq929 thanks for the feedback. You make some very valid points on issues that I agree need addressing.

Firstly, to your comments on testing. I agree that the three configurations, as they currently stand in my PR, should not be mocked and invoked in the same tests and I feel it's easy enough to pull them apart. I will say that it feels slightly uncomfortable testing that the three fetch methods return what seem to be almost [identically mocked values](https://github.com/edx/frontend-app-admin-portal/pull/417/files#diff-cd6932df673db71655a514d34f387fa428aac8ccabc3f18afad7eeb94965fb74R10-R39) and ultimately set similar states in the Lms Configuration component. I would love your feedback, but perhaps there's a more efficient (generalized?) way to test that LMS configurations appear when present? Although I fear that might mean some refactoring of the base component.

To your point on error handling (as well as validating, updating and creating?). The reuse of identical handling does lend itself to the use of a parent component. I'm happy to spend some time on refactoring, and maybe we should discuss off or online about strategies. But perhaps also we should include @taliaedX and @jlajoie in the design and implementation as they built the corresponding Canvas and Moodle components? 

I will keep in mind the comment on trending away from using React class components. Obviously was not aware of the behavioral shift and will be sure to follow the pattern in future development on the frontend. As you stated, I don't think there are any good, actionable items for this particular PR but I would love to be included in on future discussions over/poked about instances of good examples of functional components/this pattern so I have something to go off of.

src/components/LmsConfigurations/index.jsx
alex-sheehan-edx: Let's find a time to discuss in more detail. My guess is that this should probably be applied to Canvas and Moodle as well

raq929:
Yeah, in light of @taliaedX's similar PR, I think we should open a ticket for refactor rather than fixing it here. 

There haven't really been *discussions* per se about moving away from class components, it's just that Stank and I are doing the bulk of frontend work and have silently agreed not to write them. 😂 So, no worries about not knowing that. I'll try to figure out a way we can talk about that as a larger group at some point. If there are actual discussions, I'll be sure to loop you in!

Let's chat a bit today about the tests; I don't think refactoring the component will be necessary.

src/components/LmsConfigurations/index.test.jsx
taliaedX: I agree with also testing specifically Blackboard or Canvas.
However, there are a couple things we _should_ be testing with the multiple fetches at once:
1. If one 404's, it doesn't throw an error - particularly if the first call 404's, it's good to see that we still get past that conditional. (I added this in my PR)
2. If we get all 3 (we actually shouldn't outside of internal folks messing about), it still only swaps to 1 form (we don't support multiple LMS configurations, but at the same time we also don't break on it either).
3. Error page when one or more configs fails to be retrieved.

src/components/LmsConfigurations/index.jsx
raq929: Agree, I think let's open a ticket for it, since multiple people are working on this.

src/components/LmsConfigurations/index.test.jsx
raq929: So, sorry. I'm not suggesting we not test multiple fetches at once. 
But the bottom of this test only tested that the Moodle config was fetched successfully:
```
expect(wrapper.state().moodleConfig).toEqual(moodleResponse.data.results[0]);
    expect(wrapper.find('MoodleIntegrationConfigForm').props().config).toEqual(wrapper.state().moodleConfig);	    expect(wrapper.find('MoodleIntegrationConfigForm').props().config).toEqual(wrapper.state().moodleConfig)
    ```
    
    I'm saying that in separate tests, we should also test that the Canvas and Blackboard configs were successful.
    The boiler plate of the fetches can go in a `beforeEach`, and there can be a separate test with expects similar to the above to ensure that everything is working correctly.

Comment (raq929) [reply src/components/LmsConfigurations/index.test.jsx:58]: I'm saying that in separate tests, we should add similar expects for the Canvas and Blackboard configs.
We can do a `beforeEach` for the setup so we don't have to repeat a bunch of code.

Comment (raq929) [reply src/components/LmsConfigurations/index.test.jsx:58]: If while we're at it we can also add an `expect` that tests what the user is seeing rather than simply that state and props have been properly updated, that would also be great. Enzyme makes it pretty easy to test state and props; but the React community is kind of moving away from that in favor of testing actual UI output, which I'm in favor of.

Comment (taliaedX) [reply src/components/LmsConfigurations/index.test.jsx:58]: That all makes sense and is good to add. The UI stuff I've had some issues trying to implement (submit doesn't behave as expected in a testing environment for some reason), however I've been able to still utilize finds for components. Will have this up soon for you to review and Alex to base off/crib.

File: src/components/LmsConfigurations/BlackboardIntegrationConfigForm.jsx [hunk 6]
```diff
@@ -0,0 +1,284 @@
...
+              />
+            </ValidationFormGroup>
+          </div>
+        </div>
+        <div className="row">
+          <div className="col col-4">
+            <ValidationFormGroup
+              for="refreshToken"
+              helpText="The Blackboard API's refresh token token. This should be automatically propagated once you visit the oauth complete endpoint."
+              // invalid={invalidFields.refreshToken}
```
taliaedX, line 220: make sure to just delete this and the line below it (I'm guessing we don't want to require this field?)

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -39,37 +47,85 @@ REQUIRED_CANVAS_CONFIG_FIELDS.forEach((field) => {
...
     const wrapper = mount(<LmsConfigurations enterpriseId="testEnterpriseId" />);
     await waitForAsync();
     wrapper.update();
+
+    // Expect the moodle config to be false, while the others to be true
     expect(wrapper.state().moodleConfig).toBeFalsy();
     expect(wrapper.find('ErrorPage').length).toBe(1);
   });
 
+  it('return Error Page when Blackboard LMS fetch request fails with an error (not 404)', async () => {
```
taliaedX, line 103: oh I think this was my b, but I really don't see a reason to have individual tests for 1 missing config. The result is the same regardless of which config is missing (an error page is shown), so it seems a bit redundant.

When a config is present however, the actual result of that should be different for each LMS, which is why individual tests to verify the Canvas/Moodle/Blackboard/etc. component is rendered with the config are valuable.

But this? We can just do one test together to verify it shows an error when any config fails with a 400/>404
However, if it's a pain I can just clean up in my PR. I really don't care.

taliaedX (approved):
outside of whatever style/test unification preference we want here, this looks fine to me. There's some commented out lines to be cleaned up, but otherwise fine from my perspective.

src/components/LmsConfigurations/index.test.jsx
taliaedX: Assuming this is insinuating we should have a separate test for each config returning an error and an error page being generated, I'm going to respectively disagree. It's overkill.

With individual configs, we produce different components based on which config is in state, so testing that each component is rendered correctly there individually I completely understand. However, we are taking ANY 400/>404 and returning an error page component. We don't need to test that functionality 3+ times.

src/components/LmsConfigurations/index.test.jsx
taliaedX: Again, assuming this is in reference to individual tests for this, I am again disagreeing, but this time because it actually just functionally does not work.

We *only* return a list of forms if *every* config 404s. There is nothing to test individually here. If you were successful on any one of them, we'd return that component. If any of them fail, we return the error page. It's only if all have 404s that we show the list which is what this is testing.

taliaedX: We will be essentially testing "individual" 404s by doing a successful config test for each config (because the others would have to be 404s instead of other error codes)

src/components/LmsConfigurations/BlackboardIntegrationConfigForm.jsx
alex-sheehan-edx: lol whoops, good catch!

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -39,37 +47,85 @@ REQUIRED_CANVAS_CONFIG_FIELDS.forEach((field) => {
...
 
 describe('<LmsConfigurations /> ', () => {
-  it('get LMS configurations when present', async () => {
+  it('get the Moodle LMS configuration when present', async () => {
     LmsApiService.fetchMoodleConfig.mockResolvedValue(moodleResponse);
-    LmsApiService.fetchCanvasConfig.mockResolvedValue(canvasResponse);
+
+    // Mock the fetch of the other configs to 404, not found
+    LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
+    LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
```
raq929, line 63: So, not 100% necessary for now, but it would be nice, if we're repeating these calls for each test, to move these into a `beforeEach` function. (That being said, I've had some trouble with that with Jest, so, if you try it and it doesn't work, go back to this.)

src/components/LmsConfigurations/index.test.jsx
raq929: Right, but we want to verify that the response actually *is* the same if any config is missing. Also, if we decide to refactor and change how we do things, testing each one will ensure that we don't do something that lets one of them slide. Having individual tests lets us easily identify where the problem is. It is a bit redundant, but personally I like redundancy in tests.

raq929 (approved):

src/components/LmsConfigurations/index.test.jsx
taliaedX: I really don't like redundancy in tests mostly because of it slowing down feedback from CI and I feel like the way we created these APIs there's not really a way that the results would vary at all between the API endpoints, but regardless, I'd rather not keep having PRs blocked while arguing about it and our build times for this repo are quite fast at the moment, so we should just do it @alex-sheehan-edx.

alex-sheehan-edx:
@raq929 I promised screen shots, and here they are!! 

The full form:
image (8)

Example of a field being changed on the frontend
image (9)

Corresponding backend changes show!:
image (10)

PR #421, opened 2020-12-20
fix: force logout to refresh JWT roles for admin registration

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-3884

The previous implementation of the logic for admin registration broke down when a user is authenticated but without any JWT roles, causing a danger alert to be rendered.

The issue is that we need their JWT to refresh to bring in the new JWT roles they were assigned through the pending admin user configured in Django Admin. To get around this, we now force a logout with a redirect to the branded logistration page. When they login again, they will have the correct JWT roles added and be able to continue with their admin registration.

Related PR: https://github.com/edx/edx-enterprise/pull/1093

### DIFF
--- a/src/components/AdminRegisterPage/index.jsx
+++ b/src/components/AdminRegisterPage/index.jsx
@@ -1,13 +1,15 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Redirect } from 'react-router-dom';
-import {
-  Container, Row, Col, Alert, MailtoLink,
-} from '@edx/paragon';
+import { Container, Row, Col } from '@edx/paragon';
 
 import LoadingMessage from '../LoadingMessage';
 
-import { getProxyLoginUrl, redirectToProxyLogin } from '../../utils';
+import {
+  getProxyLoginUrl,
+  redirectToProxyLogin,
+  hasEnterpriseAdminRole,
+} from '../../utils';
 import apiClient from '../../data/apiClient';
 
 const AdminRegisterPage = ({ authentication, match }) => {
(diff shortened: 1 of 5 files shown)


kellmarie (approved):

src/utils.js
```diff
@@ -275,15 +276,25 @@ const mergeErrors = (object, other) => {
   return mergeWith(object, other, customizer);
 };
 
-const redirectToProxyLogin = (enterpriseSlug) => {
+const getProxyLoginUrl = (enterpriseSlug) => {
```
brittneyexline, line 279: are these util functions something that should be moved to frontend-enterprise? i assume there is similar functionality in the learner portal

src/components/AdminRegisterPage/index.jsx
```diff
@@ -1,46 +1,51 @@
...
+  useEffect(
+    () => {
+      if (!authentication.username) {
+        // user is not authenticated
+        return;
+      }
+
+      if (!hasEnterpriseAdminRole(authentication.roles)) {
+        // user is authenticated but doesn't have the `enterprise_admin` JWT role; force a log out so their
+        // JWT roles gets refreshed. on their next login, the JWT roles will be updated.
```
brittneyexline, line 27: i'm curious if there is a way to refresh the JWT without forcing the user to log out and log back in? like just have the frontend re-request the JWT? (doesn't need to be addressed here if that's too much/not possible)

brittneyexline (approved):

src/components/AdminRegisterPage/index.jsx
adamstankiewicz: Unfortunately, there's no easy way given our use of the legacy edx/frontend-auth package to manage who the authenticated user is.

The `authentication` variable above comes from the Redux store and edx/frontend-auth doesn't expose a way to update/modify what is stored there. 

I tried to utilize the `login_refresh` endpoint to avoid a forced log out but without the user going through the logistration flow again (and triggering the appropriate Django signal), the user won't have any additional roles that may have been added after the user initially logged in. Even if that did work, there's no way given edx/frontend-auth to update the `authentication` value stored in Redux at the moment.

src/utils.js
adamstankiewicz: We do compute the proxy login URL in the learner portal where these util function could be reused but I'm hesitant to consolidate in frontend-enterprise because the code duplication here is mostly negligible and IMO not worth the cost of having to upgrade dependencies in admin/learner portals anytime the logic changes in frontend-enterprise.

Plus, in a world where the admin/learner portals are consolidated in a single experience, much of the logic around redirecting to the proxy-login could be done consistently in one place across both admin + learners as opposed to implementing and sharing duplicate logic for each distinct user role / portal.

PR #431, opened 2021-01-12
Degreed Configuration Form

taliaedX:
This adds the new form for Degreed Customer Configuration.

This also attempts to set some precedent in this repo with the following:
- Moving the error alert to the top of the form and triggering focus on it to align with Accessibility guidelines.
- Uses React Testing Library on every test on the form component
- Instantiates the form component as a functional component and utilizes `useState` and `useReducer`

This also moves 2 functions we were word for word reusing across LmsConfiguration forms into a common.jsx file (those are `handleErrors` and `validate[LMSName]ConfigForm` (which has been renamed `validateLmsConfigForm`))

Oh it also fixes 2 visual bugs:
- Blackboard was labeled as Canvas
- Success Factors wasn't properly wrapped in a div.

What this does ***not*** do:
- Update existing LMS Configuration forms to this format [there is a Jira ticket for this](https://openedx.atlassian.net/browse/ENT-3908)
- Does not update the testing in the index file to use this format (also should covered by the same Jira ticket)

It should be fully ready for review.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3049,6 +3049,129 @@
         }
       }
     },
+    "@testing-library/jest-dom": {
+      "version": "5.11.9",
+      "resolved": "https://registry.npmjs.org/@testing-library/jest-dom/-/jest-dom-5.11.9.tgz",
+      "integrity": "sha512-Mn2gnA9d1wStlAIT2NU8J15LNob0YFBVjs2aEQ3j8rsfRQo+lAs7/ui1i2TGaJjapLmuNPLTsrm+nPjmZDwpcQ==",
+      "dev": true,
+      "requires": {
+        "@babel/runtime": "^7.9.2",
+        "@types/testing-library__jest-dom": "^5.9.1",
+        "aria-query": "^4.2.2",
+        "chalk": "^3.0.0",
+        "css": "^3.0.0",
+        "css.escape": "^1.5.1",
+        "lodash": "^4.17.15",
+        "redent": "^3.0.0"
+      },
+      "dependencies": {
+        "ansi-styles": {
+          "version": "4.3.0",
+          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
+          "integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
+          "dev": true,
+          "requires": {
+            "color-convert": "^2.0.1"
+          }
+        },
+        "aria-query": {
+          "version": "4.2.2",
+          "resolved": "https://registry.npmjs.org/aria-query/-/aria-query-4.2.2.tgz",
+          "integrity": "sha512-o/HelwhuKpTj/frsOsbNLNgnNGVIFsVP/SW2BSF14gVl7kAfMOJ6/8wUAUvG1R1NHKrfG+2sHZTu0yauT1qBrA==",
+          "dev": true,
+          "requires": {
+            "@babel/runtime": "^7.10.2",
+            "@babel/runtime-corejs3": "^7.10.2"
+          }
+        },
+        "chalk": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-3.0.0.tgz",
+          "integrity": "sha512-4D3B6Wf41KOYRFdszmDqMCGq5VV/uMAB273JILmO+3jAlh8X4qDtdtgCR3fxtbLEMzSx22QdhnDcJvu2u1fVwg==",
+          "dev": true,
+          "requires": {
+            "ansi-styles": "^4.1.0",
+            "supports-color": "^7.1.0"
+          }
+        },
+        "color-convert": {
+          "version": "2.0.1",
+          "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-2.0.1.tgz",
+          "integrity": "sha512-RRECPsj7iu/xb5oKYcsFHSppFNnsj/52OVTRKb4zP5onXwVF3zVmmToNcOfGC+CRDpfK/U584fMg38ZHCaElKQ==",
+          "dev": true,
+          "requires": {
+            "color-name": "~1.1.4"
+          }
+        },
+        "color-name": {
+          "version": "1.1.4",
+          "resolved": "https://registry.npmjs.org/color-name/-/color-name-1.1.4.tgz",
+          "integrity": "sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==",
+          "dev": true
+        },
+        "css": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/css/-/css-3.0.0.tgz",
+          "integrity": "sha512-DG9pFfwOrzc+hawpmqX/dHYHJG+Bsdb0klhyi1sDneOgGOXy9wQIC8hzyVp1e4NRYDBdxcylvywPkkXCHAzTyQ==",
+          "dev": true,
+          "requires": {
+            "inherits": "^2.0.4",
+            "source-map": "^0.6.1",
+            "source-map-resolve": "^0.6.0"
+          }
+        },
+        "has-flag": {
+          "version": "4.0.0",
+          "resolved": "https://registry.npmjs.org/has-flag/-/has-flag-4.0.0.tgz",
+          "integrity": "sha512-EykJT/Q1KjTWctppgIAgfSO0tKVuZUjhgMr17kqTumMl6Afv3EISleU7qZUzoXDFTAHTDC4NOoG/ZxU3EvlMPQ==",
+          "dev": true
+        },
+        "indent-string": {
+          "version": "4.0.0",
+          "resolved": "https://registry.npmjs.org/indent-string/-/indent-string-4.0.0.tgz",
+          "integrity": "sha512-EdDDZu4A2OyIK7Lr/2zG+w5jmbuk1DVBnEwREQvBzspBJkCEbRa8GxU1lghYcaGJCnRWibjDXlq779X1/y5xwg==",
+          "dev": true
+        },
+        "redent": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/redent/-/redent-3.0.0.tgz",
+          "integrity": "sha512-6tDA8g98We0zd0GvVeMT9arEOnTw9qM03L9cJXaCjrip1OO764RDBLBfrB4cwzNGDj5OA5ioymC9GkizgWJDUg==",
+          "dev": true,
+          "requires": {
+            "indent-string": "^4.0.0",
+            "strip-indent": "^3.0.0"
+          }
+        },
+        "source-map-resolve": {
+          "version": "0.6.0",
+          "resolved": "https://registry.npmjs.org/source-map-resolve/-/source-map-resolve-0.6.0.tgz",
+          "integrity": "sha512-KXBr9d/fO/bWo97NXsPIAW1bFSBOuCnjbNTBMO7N59hsv5i9yzRDfcYwwt0l04+VqnKC+EwzvJZIP/qkuMgR/w==",
+          "dev": true,
+          "requires": {
+            "atob": "^2.1.2",
+            "decode-uri-component": "^0.2.0"
+          }
+        },
+        "strip-indent": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/strip-indent/-/strip-indent-3.0.0.tgz",
+          "integrity": "sha512-laJTa3Jb+VQpaC6DseHhF7dXVqHTfJPCRDaEbid/drOhgitgYku/letMUqOXFoWV0zIIUbjpdH2t+tYj4bQMRQ==",
+          "dev": true,
+          "requires": {
+            "min-indent": "^1.0.0"
+          }
+        },
+        "supports-color": {
+          "version": "7.2.0",
+          "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-7.2.0.tgz",
+          "integrity": "sha512-qpCAvRl9stuOHveKsn7HncJRvv501qIacKzQlO/+Lwxc9+0q2wLyv4Dfvt80/DPn2pqOBsJdDiogXGR9+OvwRw==",
+          "dev": true,
+          "requires": {
+            "has-flag": "^4.0.0"
+          }
+        }
+      }
+    },
     "@testing-library/react": {
       "version": "10.4.8",
       "resolved": "https://registry.npmjs.org/@testing-library/react/-/react-10.4.8.tgz",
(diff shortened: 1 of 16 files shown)


src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+      fireEvent.change(screen.getByLabelText('API Client ID'), {
+        target: { value: 'testing' },
+      });
+      fireEvent.change(screen.getByLabelText('API Client Secret'), {
+        target: { value: 'testing' },
+      });
+      fireEvent.click(screen.getByText('Submit'));
+    });
+
+    expect(LmsApiService.postNewDegreedConfig).toHaveBeenCalled();
```
taliaedX, line 65: So this test technically passes, but it is throwing a console error:
```
console.error
    Warning: An update to DegreedIntegrationConfigForm inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in DegreedIntegrationConfigForm
```

I _could_ ignore it, but I feel like maybe there is a better way to do this. I tried just using the regular ol' mounting and spying, but it's just fundamentally not working and I don't know if it's because of the functional component aspect or what. We don't have a lot of great examples in the codebase and functional components + testing functionality do not seem to go hand in hand on the interwebs. :(

src/components/LmsConfigurations/SuccessFactorsIntegrationConfigForm.test.jsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
-import SuccessFactorsConfigForm, { REQUIRED_SUCCESS_FACTOR_CONFIG_FIELDS } from './SuccessFactorsIntegrationConfigForm';
+import SuccessFactorsIntegrationConfigForm, { REQUIRED_SUCCESS_FACTOR_CONFIG_FIELDS } from './SuccessFactorsIntegrationConfigForm';
```
taliaedX, line 3: How the eff this was passing before I have no idea.

src/components/LmsConfigurations/common.jsx
```diff
@@ -0,0 +1,21 @@
...
+  return errorMsg;
+};
+
+/**
+   * Validates this form. If the form is invalid, it will return the fields
+   * that were invalid. Otherwise, it will return an empty object.
+   * @param {FormData} formData
+   * @param {[String]} requiredFields
+   */
+export const validateLmsConfigForm = (formData, requiredFields) => {
```
taliaedX, line 16: I plan on updating the other LMS configurations to use these, but want to get my own component fully functioning first.

src/components/LmsConfigurations/index.jsx
```diff
@@ -142,11 +161,23 @@ class LmsConfigurations extends React.Component {
           <Collapsible
             styling="card"
             className="shadow"
-            title="Canvas"
+            title="Blackboard"
```
taliaedX, line 164: Fixing a typo I found when initially rendering the page in stage.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+  'degreedBaseUrl',
+  'degreedUserId',
+  'degreedUserPassword',
+  'secret',
+  'degreedCompanyId',
+  'key',
+];
+
+function DegreedIntegrationConfigForm({ enterpriseId, config }) {
+  const [invalidFields, setInvalidFields] = useState({});
```
taliaedX, line 23: These names really highlight how we reuse names elsewhere (I had to change what invalidFields somewhere else to something else. :()

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+export const REQUIRED_DEGREED_CONFIG_FIELDS = [
+  'degreedBaseUrl',
+  'degreedUserId',
+  'degreedUserPassword',
+  'secret',
+  'degreedCompanyId',
+  'key',
+];
+
+function DegreedIntegrationConfigForm({ enterpriseId, config }) {
```
taliaedX, line 22: Definitely wouldn't mind a general look over of this entire function structure, because yeaaaah

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
raq929: Oh this has been PLAUGING me lately. Wrapping it in `act` doesn't work. You might try making the test function async and `await`ing the fireEvent, but fireEvent is supposedly wrapped in act anyway. I've got other tests with this issue; I think let's ask about it on fedex and see if anyone has a good solution and if not, leave it.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(screen.getByLabelText('Degreed Organization Code')).toHaveClass('is-invalid');
+    expect(screen.getByLabelText('API Client ID')).toHaveClass('is-invalid');
+    expect(screen.getByLabelText('API Client Secret')).toHaveClass('is-invalid');
+  });
+
+  test('submit calls postNewDegreedConfig when config is not present', () => {
+    render(<DegreedIntegrationConfigForm enterpriseId={enterpriseId} />);
+    act(() => {
+      fireEvent.change(screen.getByLabelText('Degreed User ID'), {
+        target: { value: 'testing' },
```
raq929, line 45: I think it's worth changing these values to be different for each one, otherwise you don't know if the form has the correct value or if it mistakenly has a value that's supposed to be in a different field.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
raq929, line 65: I think it's worth testing that the API service was called with the correct information.

raq929 (approved):

raq929:
Looked at the existing tests and the function you commented on but didn't look at other code fyi

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
taliaedX: Hmm yeah I'll try the await/async in the meantime. I tried to wrap every fireEvent in act lmao and it still didn't work.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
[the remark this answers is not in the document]
taliaedX: I'll look into that then thanks!

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+    transformedData.append('enterprise_customer', enterpriseId);
+    try {
+      const response = await LmsApiService.postNewDegreedConfig(transformedData);
+      return setConfig(response.data);
+    } catch (responseError) {
+      return handleErrors(responseError);
+    }
+  };
+
+  const updateDegreedConfig = async (formData, configId) => {
```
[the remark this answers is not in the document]
taliaedX: I had thought about making the createLmsConfig functions to be reused across LMS configurations, but with them all using different LmsService API calls, it seemed like a headache to add that logic.

That said, the current function is *not* reused outside of this specific form. With the functional structure I even thought about not having a separate function and just using the logic where needed since we only use it once, but it's still cleaner this way.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
taliaedX: Oops. I actually meant to do that (needed to confirm the format first). I just completely forgot while trying to get the rest of this functioning. Will look into that now.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+            }}
+            disabledStates={[SUBMIT_STATES.PENDING]}
+            variant="primary"
+            className="ml-3 col"
+          />
+        </div>
+      </div>
+      <div className="row">
+        <div className="col col-3 mt-3">
+          {errorAlert}
```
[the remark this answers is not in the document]
taliaedX: I thought I remembered seeing something about this in the quiz and yeah I finally found it: 
>Error messages should appear at the beginning of the form, and focus should be moved there so that the user can proceed through the form again to correct their mistakes.

taliaedX:
Tagging @alex-sheehan-edx as reviewer to highlight the changes here since he may also be new to the particular libraries and processes here.

package-lock.json
```diff
@@ -11211,7 +11359,7 @@
     },
     "is-data-descriptor": {
       "version": "0.1.4",
-      "resolved": "https://registry.npmjs.org/is-data-descriptor/-/is-data-descriptor-0.1.4.tgz",
+      "resolved": "http://registry.npmjs.org/is-data-descriptor/-/is-data-descriptor-0.1.4.tgz",
```
taliaedX, line 11362: why would you do this npm?

src/components/LmsConfigurations/index.jsx
taliaedX, line 164: This was a typo in master. Just tacked on here since I noticed it when I tested something on stage.

src/components/LmsConfigurations/index.jsx
```diff
@@ -115,6 +129,11 @@ class LmsConfigurations extends React.Component {
           >
             <MoodleIntegrationConfigForm enterpriseId={this.props.enterpriseId} />
           </Collapsible>
+        </div>
+        <div
+          key="successfactors-form-link"
```
taliaedX, line 134: The lack of this was causing the LMS items to not line up.

package-lock.json
edx-abolger: So the npm maintainers have a proposed solution to this: https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285
Can you see if using "npm install --prefer-online" makes this go away? If so, we may wish to open an edX level ticket to fix it across some repos.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,356 @@
...
+  'degreedUserId',
+  'degreedUserPassword',
+  'secret',
+  'degreedCompanyId',
+  'key',
+];
+
+function configFormReducer(state, action) {
+  switch (action.type) {
+    case 'PENDING': {
```
edx-abolger, line 24: Question mostly for my understanding: if you don't update the invalidFields on these other states they retain previous value, correct? Is that an issue?

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,356 @@
...
+        ...state,
+        submitState: SUBMIT_STATES.ERROR,
+        invalidFields: action.value,
+        error: 'Form could not be submitted as there are fields with invalid values. Please correct them below.',
+      };
+    }
+    case 'COMPLETE': {
+      return {
+        ...state,
+        invalidFields: {},
```
edx-abolger, line 41: Don't these items here match the default state already? Do you have to explicitly set them to invalid?

edx-abolger (approved):

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
taliaedX: If I follow correctly, it seems like the flow you'd expect would be starting with all required fields being invalid and then instead of adding to an object/array containing those fields, you would remove them when they are valid? Or am I completely misunderstanding?

In the current implementation we start with no field being invalid and iterate through the required fields and the fields in the form data to determine what is missing and error accordingly, so yes they have to be explicitly set as invalid. While typing this I was thinking through the reverse and I'm not sure that'd be a very easy proposition.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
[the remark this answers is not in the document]
raq929: I wonder if it would make sense to have a generic LmsServiceApi (hook? module?) that accepts the endpoint, so the function itself doesn't have to make that choice. 🤔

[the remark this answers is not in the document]
raq929: (Not a requested change, just something to think about)

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,292 @@
...
+          iconClassName="fa fa-times-circle"
+          title="Unable to submit config form:"
+          message={error}
+        />
+      </div>
+    );
+  }
+
+  return (
+    <form
```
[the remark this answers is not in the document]
raq929: Ha, so there are some Form components that do exist that would be good to use in the future, for styling consistency if nothing else: https://edx.github.io/paragon/components/forms
Doesn't do what I want it to do, though. I want to have a list of my form fields, give it to a Form component and call it a day.
I think for now this is OK, in the future, use the Paragon components.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+    });
+    fireEvent.click(screen.getByText('Submit'));
+
+    const postedFormData = Array.from(LmsApiService.postNewDegreedConfig.mock.calls[0][0].entries())
+      .reduce((acc, f) => ({ ...acc, [f[0]]: f[1] }), {});
+
+    expect(postedFormData).toMatchObject(expectedFormData);
+  });
+
+  test('submit calls updateDegreedConfig when config is not present', () => {
```
raq929, line 84: The only other test I would like is one that verifies that the initial config is being displayed to the user. So, rendering with an initialConfig and checking the text of `screen.getByLabelText('Active');` to make sure it's showing the initial config.

src/components/LmsConfigurations/index.jsx
raq929: Good catch.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -139,6 +159,19 @@ describe('<LmsConfigurations /> ', () => {
     expect(wrapper.find('SuccessFactorsIntegrationConfigForm').props().config).toEqual(wrapper.state().sapsfConfig);
   });
 
+  it('get the Degreed LMS configuration when present', async () => {
```
raq929, line 162: nitiest of nits: `gets` instead of `get` so it can be read as a sentence: "It gets the..."

raq929 (approved):
Looking good! There's ooone more test that I'd like. :)

src/components/LmsConfigurations/index.jsx
alex-sheehan-edx: *eye twitches*

src/components/LmsConfigurations/DegreedIntegrationConfigForm.jsx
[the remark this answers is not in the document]
taliaedX: Yeah I think there's definitely room for refactoring down the line. May be something we end up doing anyways as part of the test/formate updates.

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
taliaedX: Ah gotcha. Yeah I can add that.

package-lock.json
taliaedX: hmmm yeah that didn't seem to fix it. I also tried clearing the cache and removing the modules before running it again just in case. :/
I'm just going to manually find & replace the http for now.

src/components/LmsConfigurations/index.test.jsx
alex-sheehan-edx: FWIW, seems like all the methods use `get`, so my only ask is that we stay consistent!

raq929 (approved):
LGTM!

alex-sheehan-edx (approved):
A lot of good improvements from when I was working on the Blackboard configuration. Thanks for keeping me in the loop 👍  LGTM

PR #432, opened 2021-01-14
ENT-3883 Disable bulk enrollment when the subscription is invalid

edx-abolger:
Added flags so that Bulk Enrollment depends on same subscription management screen, with the idea being that if your enterprise can manage subscriptions it should be able to enroll learners.

### DIFF
--- a/src/components/EnterpriseApp/index.jsx
+++ b/src/components/EnterpriseApp/index.jsx
@@ -163,7 +163,7 @@ class EnterpriseApp extends React.Component {
                       render={routeProps => <RequestCodesPage {...routeProps} />}
                     />,
                   ]}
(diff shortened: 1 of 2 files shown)


raq929 (approved):
LGTM! This may be overkill, but I kinda want some tests that prove that the tab is not in the nav when the feature is off or the `enableSubscriptionManagementScreen` is false. 
It's the kind of thing where, if there are existing tests for this component, I would definitely add tests, but if there's not, I might open a tech-debt ticket to add tests to the component, because it's significantly more work to set up initial tests than to add one.

edx-abolger:
Per @raq929 request: https://openedx.atlassian.net/browse/ENT-3958 opened since PR

PR #437, opened 2021-01-19
ENT-3835: Adds in Bulk Enrollment Modal

jlajoie:
Without error
Screen Shot 2021-01-19 at 7 30 02 AM

With error
Screen Shot 2021-01-19 at 7 43 09 AM

With failed learners message (no dialog, also nothing powering this from the backend yet
Screen Shot 2021-01-19 at 7 59 22 AM


src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+class BulkEnrollmentModal extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.errorMessageRef = React.createRef();
+    this.modalRef = React.createRef();
+
+    this.handleModalSubmit = this.handleModalSubmit.bind(this);
+
+    this.state = { open: true };
```
raq929, line 19: I think current best practice is to put this below the constructor as
`state = { open: true };`
But I also think Stank and I disagree about that, so, take it or leave it.

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+      selectedCourseRunKeys,
+      submitFailed,
+    } = this.props;
+
+    return (
+      <>
+        {submitFailed && this.renderErrorMessage()}
+        {failedLearners.length > 0 && this.renderFailedLearnersMessage(failedLearners)}
+        <p className="courses-selected">Courses selected: {selectedCourseRunKeys.length}</p>
+        <form onSubmit={e => e.preventDefault()}>
```
raq929, line 130: Can you use the Paragon Form components here? I think that'll make styling in line with what we want without thinking about it.
https://edx.github.io/paragon/components/forms
And likely
https://edx.github.io/paragon/components/validationformgroup

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+            <h3>Add Learners</h3>
+            <div>
+              <Field
+                id="bulk-enrollment-email-addresses"
+                name="bulk-enrollment-email-addresses"
+                component={TextAreaAutoSize}
+                label="Email addresses"
+                type="text"
+              />
+              <p>To add more than one user, enter one email address per line</p>
```
raq929, line 141: Check with Ange, but I think we probably want periods at the end of this microcopy.

src/components/CsvUpload/index.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Input } from '@edx/paragon';
+
+class CsvUpload extends React.Component {
```
raq929, line 5: You may already be doing this, I know you said you were gonna push some tests, but can you add some tests for this component? Basically I want to test that it properly encodes / decodes the value.

raq929 (approved):

raq929:
Approved pending some extra testing and obvs CI has to pass.

src/components/BulkEnrollmentModal/index.jsx
jlajoie: Spent some time messing with this, doesn't look like there's an easy way to integrate the paragon forms with redux-forms (or at least I can't figure it out), going to punt on this and leave it consistent with the rest of the repo

PR #441, opened 2021-01-20
minor UI/bug fixes for LMS config forms

taliaedX:
This just cleans up some graphical issues and backend bugs:
- fixed active flag in some of the config forms was being sent as "undefined" which was causing an error in backend validation (field is boolean and does not support undefined)
- fixed error state not clearing after resubmission.
- moved Active box to top for Canvas form for ~ consistency ~
- used put instead of patch, because form value was being interpreted as no change instead of "false"
- Fixed the error format of the error we send to NewRelic as it wasn't matching the actual NR method logic.

### DIFF
--- a/src/data/services/LmsApiService.js
+++ b/src/data/services/LmsApiService.js
@@ -137,7 +137,7 @@ class LmsApiService {
   }
 
   static updateCanvasConfig(formData, id) {
-    return apiClient.patch(`${LmsApiService.lmsIntegrationUrl}/canvas/configuration/${id}/`, formData, 'json');
+    return apiClient.put(`${LmsApiService.lmsIntegrationUrl}/canvas/configuration/${id}/`, formData, 'json');
   }
 
   static fetchBlackboardConfig(uuid) {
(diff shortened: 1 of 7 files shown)


src/data/services/LmsApiService.js
```diff
@@ -137,7 +137,7 @@ class LmsApiService {
   }
 
   static updateCanvasConfig(formData, id) {
-    return apiClient.patch(`${LmsApiService.lmsIntegrationUrl}/canvas/configuration/${id}/`, formData, 'json');
+    return apiClient.put(`${LmsApiService.lmsIntegrationUrl}/canvas/configuration/${id}/`, formData, 'json');
```
binodpant, line 140: is the formData here always going to replace the entire config, I had thought the patch previously was about only updating some of the fields,,, if this has the desired effect in the backend, we are all set!

binodpant (approved):
looks like we are not adding a new feature or fixing a bug necessarily so probably don't need new tests, made a minor comment/question on put vs patch but otherwise looks like a good set of changes to me

src/data/services/LmsApiService.js
taliaedX: The way it is setup it will always send the full form data anyways.
The reason I swapped this - and apologies, I meant to comment on this line - is because of these 2 things combining together:
- The form component is storing the value as "null" instead of "false"
- sending a null value via the patch endpoint says "don't change this" rather than "this is changing to false"

binodpant: got it, makes sense now, just checking on accidental overwriting/losing of existing data, since that won't happen we are good, thanks!

taliaedX: no worries. I meant to explain this since it was definitely a weird change.

PR #444, opened 2021-01-21
Cornerstone Config Form

taliaedX:
Adds Cornerstone to the LMS config forms.
I also fixed a state warning (something about "uncontrolled input") that was happening with Degreed.
Also made tests more consistent with what they were supposed to be doing (ie they previously only passed due to order of operations)

### DIFF
--- a/src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
+++ b/src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
@@ -10,13 +10,8 @@ import StatusAlert from '../StatusAlert';
 import SUBMIT_STATES from '../../data/constants/formSubmissions';
 import { handleErrors, validateLmsConfigForm } from './common';
 
-export const REQUIRED_DEGREED_CONFIG_FIELDS = [
-  'degreedBaseUrl',
-  'degreedUserId',
-  'degreedUserPassword',
-  'secret',
-  'degreedCompanyId',
-  'key',
+export const REQUIRED_CORNERSTONE_CONFIG_FIELDS = [
+  'cornerstoneBaseUrl',
 ];
 
 function configFormReducer(state, action) {
@@ -61,7 +56,7 @@ function configFormReducer(state, action) {
   }
 }
 
-function DegreedIntegrationConfigForm({ enterpriseId, config }) {
+function CornerstoneIntegrationConfigForm({ enterpriseId, config }) {
   const [state, dispatch] = React.useReducer(
     configFormReducer,
     {
@@ -78,11 +73,11 @@ function DegreedIntegrationConfigForm({ enterpriseId, config }) {
    * Returns if there is an error.
    * @param {FormData} formData
    */
-  const createDegreedConfig = async formData => {
+  const createCornerstoneConfig = async formData => {
     const transformedData = snakeCaseFormData(formData);
     transformedData.append('enterprise_customer', enterpriseId);
     try {
-      const response = await LmsApiService.postNewDegreedConfig(transformedData);
+      const response = await LmsApiService.postNewCornerstoneConfig(transformedData);
       return dispatch({
         type: 'SET_CONFIG',
         value: response.data,
@@ -92,11 +87,11 @@ function DegreedIntegrationConfigForm({ enterpriseId, config }) {
     }
   };
 
-  const updateDegreedConfig = async (formData, configId) => {
+  const updateCornerstoneConfig = async (formData, configId) => {
     const transformedData = snakeCaseFormData(formData);
     transformedData.append('enterprise_customer', enterpriseId);
     try {
-      const response = await LmsApiService.updateDegreedConfig(transformedData, configId);
+      const response = await LmsApiService.updateCornerstoneConfig(transformedData, configId);
       return dispatch({
         type: 'SET_CONFIG',
         value: response.data,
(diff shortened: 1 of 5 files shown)


src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,256 @@
...
+    }
+    case 'ERROR': {
+      return { ...state, error: action.value, submitState: SUBMIT_STATES.ERROR };
+    }
+    case 'INVALID': {
+      return {
+        ...state,
+        submitState: SUBMIT_STATES.ERROR,
+        invalidFields: action.value,
+        error: 'Form could not be submitted as there are fields with invalid values. Please correct them below.',
```
edx-abolger, line 30: Quick check- does this message meet these guidelines? https://openedx.atlassian.net/wiki/spaces/BPL/pages/1966375321/Microcopy+writing+guidelines

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,256 @@
...
+      };
+    }
+    case 'BASE_SUBMIT': {
+      return {
+        ...state,
+        submitState: SUBMIT_STATES.DEFAULT,
+      };
+    }
+    default: {
+      return { state };
```
edx-abolger, line 54: Does this one not need a spread operator for some reason? (I could be wrong, I'm not amazing at React. yet.)

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
taliaedX: I think so? The only thing I could think to change is writing "your form" (which sounds bizarre to me since it's not their form) or "The form". I see something about avoiding punctuation for helper text, but I don't count this as "short".

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
taliaedX: That's for modifying an iterable isn't it? We're not modifying anything here.

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  let errorAlert;
+  if (state.error) {
+    errorAlert = (
+      <div className="form-group is-invalid align-items-left">
+        <StatusAlert
+          alertType="danger"
+          iconClassName="fa fa-times-circle"
+          title="Unable to submit config form:"
```
edx-abolger, line 158: Same thought here about the microcopy. Although if all of these match the other LMS forms, maybe a ticket to change all the LMS microcopy to meet standards is more sensible than changing here.

src/components/LmsConfigurations/index.test.jsx
```diff
@@ -163,15 +183,30 @@ describe('<LmsConfigurations /> ', () => {
     LmsApiService.fetchMoodleConfig.mockRejectedValue(notFoundMoodleResponse);
     LmsApiService.fetchBlackboardConfig.mockRejectedValue(notFoundBlackboardResponse);
     LmsApiService.fetchCanvasConfig.mockRejectedValue(notFoundCanvasResponse);
-    LmsApiService.fetchSuccessFactorsConfig.mockResolvedValue(successFactorsResponse);
+    LmsApiService.fetchSuccessFactorsConfig.mockRejectedValue(notFoundSuccessFactorsResponse);
```
edx-abolger, line 186: Was this just wrong before?

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
taliaedX: They all just match the Django Admin helper text, but slightly modified for user legibility. I can improve it, but I'm not 100% sure what piece of the microcopy it doesn't match. I guess punctuation? Not using second person pronouns?

edx-abolger (approved):
Looks good to me, but if you don't want to change microcopy in this ticket to keep it matching the microcopy in the existing LMS configs, please file a new ticket.

src/components/LmsConfigurations/index.test.jsx
taliaedX: Sort of. Ultimately this test should have failed, but because it's not a proper integration test, it passed.
Basically, you _could_ have multiple different integration configs (if configured on the backend) and we would store it in the state. However, what the user would see is the first config in the list (which would have been SuccessFactors here). 
Since we aren't currently wanting users to create multiple integration config forms at this point, this test case should just test the expectation of 1 integration returning true.

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
raq929: Yeah, this definitely qualifies as needing punctuation.

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
```diff
@@ -0,0 +1,256 @@
...
+        ...state,
+        invalidFields: {},
+        error: null,
+        submitState: SUBMIT_STATES.COMPLETE,
+      };
+    }
+    case 'SET_CONFIG': {
+      return {
+        ...state,
+        config: action.value,
```
raq929, line 44: should setting the config make error states go away?

raq929 (approved):
LGTM!

src/components/LmsConfigurations/CornerstoneIntegrationConfigForm.jsx
taliaedX: It could, but the flow should be as follows:
Click submit > trigger "PENDING" state (which sets errors to null) > hit validation failure > error state is set > user fixes and hits submit again > trigger PENDING state again (thus clearing errors), etc.

So every time they hit submit we're clearing the errors so clearing it here would be a bit redundant.

PR #447, opened 2021-01-25
Allow access to LMS Configuration screen without feature flag

taliaedX:
This is for ENT-3418.

This now enables the LMS Configuration screen such that [**if enabled in LMS admin**] an enterprise customer can now access the sidebar item without the feature flag query string.


alex-sheehan-edx (approved):
LGTM

PR #449, opened 2021-01-28
Upgrade to frontend-platform

raq929:
Upgrades us away from the deprecated frontend-auth repo to `frontend-platform`

https://openedx.atlassian.net/browse/ENT-2561

**Things to note:** 

- The Header component is what has changed the most; I decided for that one it was worth it to convert to a functional component. 
- A lot of containers that served to just pass in auth credentials have gone away
- We're mocking some things in setupTest, because that was the fastest way to get tests functional, and we definitely don't want to make real API calls and logs when we test. These can be unmocked as needed in individual tests.
- I'm skipping one test that I couldn't get working; I confirmed manually that what it's testing for works; ideas on that are welcome.


src/components/AdminRegisterPage/index.jsx
```diff
@@ -26,14 +28,14 @@ const AdminRegisterPage = ({ authentication, match }) => {
         // user is authenticated but doesn't have the `enterprise_admin` JWT role; force a log out so their
         // JWT roles gets refreshed. on their next login, the JWT roles will be updated.
         const logoutRedirectUrl = getProxyLoginUrl(enterpriseSlug);
```
edx-abolger, line 30: Does it make sense to update this variable name for readability? I really don't understand at first glance how the proxy login url is also the logout url, nor how adding a url to a history forces a log out of any kind.

src/components/App/index.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
edx-abolger, line 1: Does App as a new component not get tests?

src/components/AdminRegisterPage/index.jsx
```diff
@@ -66,12 +68,11 @@ const AdminRegisterPage = ({ authentication, match }) => {
 };
 
 AdminRegisterPage.propTypes = {
-  authentication: PropTypes.shape().isRequired,
```
binodpant, line 69: __worth a quick check:__ not 100% on this but I thought there was some eslint rule or something else maybe that suggested using shape({}) instead of just shape() but don't take my word for it

src/components/App/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
+import EnterpriseApp from '../../containers/EnterpriseApp';
+import Header from '../../containers/Header';
+import Footer from '../../containers/Footer';
+import EnterpriseIndexPage from '../../containers/EnterpriseIndexPage';
+import AdminRegisterPage from '../AdminRegisterPage';
+import UserActivationPage from '../UserActivationPage';
+import NotFoundPage from '../NotFoundPage';
+import SupportPage from '../SupportPage';
+// import { withErrorBoundary } from '../ErrorBoundary';
```
binodpant, line 16: remove this line?

src/components/App/index.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { Switch, Router } from 'react-router-dom';
+
+import { Helmet } from 'react-helmet';
+import { AuthenticatedPageRoute, PageRoute, AppProvider } from '@edx/frontend-platform/react';
+
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
```
binodpant, line 7: minor suggestion: group imports together, e.g. this @edx could go with the one above, in-project imports can be one group next and so on

binodpant: I typically do (third party on top, external edx imports next, then in-project imports)

src/components/ErrorBoundary/index.jsx
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
-import NewRelicService from '../../data/services/NewRelicService';
+import { logError } from '@edx/frontend-platform/logging';
```
edx-abolger, line 3: @binodpant does this mean @raq929 also just fixed https://openedx.atlassian.net/browse/ENT-3852?  (*hopeful face*)

src/components/App/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+            <PageRoute component={NotFoundPage} />
+          </Switch>
+          <Footer />
+        </ToastsProvider>
+      </Router>
+    </AppProvider>
+  );
+};
+
+// const PortalAppWrapper = withErrorBoundary(AppWrapper);
```
binodpant, line 73: can be removed

src/components/EnterpriseList/__snapshots__/EnterpriseList.test.jsx.snap
```diff
@@ -396,35 +396,8 @@ Array [
                 placeholder="Search by enterprise name..."
                 role="searchbox"
                 type="text"
-                value="enterprise name"
+                value=""
               />
-              <button
```
binodpant, line 401: `CLARIFICATION`: not clear on this change, so the button that used to be there will no longer render?

src/components/ErrorBoundary/index.jsx
edx-abolger: @raq929 this may mean you should delete the service itself?

src/components/EnterpriseList/index.jsx
```diff
@@ -182,4 +182,4 @@ EnterpriseList.propTypes = {
   error: PropTypes.instanceOf(Error),
 };
 
-export default EnterpriseList;
+export default withRouter(EnterpriseList);
```
binodpant, line 185: `CLARIFICATION`: so we now return a withRouter surrounded component instead of the regular component, okay, I believe this means it's an in place replacement for wherever this component is used, just that it has access to extra props right?

src/components/ErrorBoundary/ErrorBoundary.test.jsx
```diff
@@ -27,14 +29,13 @@ describe('<ErrorBoundary />', () => {
...
-    NewRelicService.logError = jest.fn();
 
     const wrapper = shallow((
       <ErrorBoundary>
         <Welcome />
       </ErrorBoundary>
     ));
     wrapper.find(Welcome).simulateError(welcomeError);
-    expect(NewRelicService.logError).toHaveBeenCalledWith(new Error('Catch me if you can!'));
+    expect(logError).toHaveBeenCalledWith(new Error('Catch me if you can!'));
```
binodpant, line 39: this means the imported logger takes care of New Relic logging too? we have a ticket for abstracting out new relic, perhaps this pattern can be re used in that

src/components/Header/index.jsx
```diff
@@ -1,102 +1,120 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
edx-abolger, line 1: nit: do we have normal standards related to 1 component per file? If so, please file ticket. if that's not a thing we do, ignore me.

src/components/Header/Header.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+describe('<Logo />', () => {
+  it('renders enterprise logo correctly', () => {
+    const props = {
+      enterpriseLogo: 'https://test.url/image/1.png',
+      enterpriseName: 'Test Enterprise',
+    };
+
+    const wrapper = mount(<Logo {...props} />);
```
binodpant, line 37: nit: would love to get rid of enzyme at some point but step by step! not for this PR already quite big!

src/components/InviteLearnersModal/index.jsx
```diff
@@ -12,7 +13,6 @@ import {
   validateEmailAddresses,
```
edx-abolger, line 13: Is there a future ticket to be filed to move these validators to frontend-platform?

src/components/Header/index.jsx
```diff
@@ -1,102 +1,120 @@
...
+};
+
+const Header = ({
+  hasSidebarToggle, enterpriseName, enterpriseLogo,
+}) => {
+  const user = getAuthenticatedUser();
+  useEffect(() => {
+    // if there is no email, the user data has not been hydrated
+    if (user && !user.email) {
+      hydrateAuthenticatedUser();
```
binodpant, line 82: dont know how hydrateAuthentication works exactly, there is hopefully not going to be a re execution of this useEffect once this is called?

binodpant: I guess even if it does , this check means it won't try to rehydrate.. so might be okay

src/components/UserActivationPage/index.jsx
```diff
@@ -16,50 +17,39 @@ const USER_ACCOUNT_POLLING_TIMEOUT = 5000;
...
-      return () => {
-        if (timeout) {
-          clearInterval(timeout);
-        }
-      };
-    },
-    [authentication, userAccount],
-  );
+  useInterval(() => {
+    if (username && !isActive) {
```
edx-abolger, line 33: Why do hydration criteria differ from component to component? The previous one uses email for this?

src/components/ErrorBoundary/index.jsx
binodpant: yes I believe so! as long as we can confirm the new relic posting is covered in that logger call, seems like it should

src/components/UserActivationPage/index.jsx
```diff
@@ -16,50 +17,39 @@ const USER_ACCOUNT_POLLING_TIMEOUT = 5000;
...
+    if (!roles?.length) {
       return (
         <Redirect to={`/${enterpriseSlug}/admin/register`} />
       );
     }
 
     // user data is hydrated with a verified email address, so redirect the user
     // to the default page in the Admin Portal.
-    if (userAccount?.loaded && userAccount?.isActive) {
+    if (isActive !== undefined && isActive) {
```
edx-abolger, line 52: Does this work??? I thought an undefined variable in javascript throws an error no matter what you try to compare it to?

edx-abolger: My normal workaround: (typeof isActive !== 'undefined')

src/components/UserActivationPage/index.jsx
```diff
@@ -72,7 +62,7 @@ const UserActivationPage = ({
     }
 
     // user data is hydrated with a unverified email address, so display a warning message
-    const isUserLoadedAndInactive = !!(userAccount?.loaded && !userAccount?.isActive);
+    const isUserLoadedAndInactive = !!(isActive !== undefined && !isActive);
```
edx-abolger, line 65: same question as above

src/containers/Header/__snapshots__/Header.test.jsx.snap
```diff
@@ -1,269 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
edx-abolger, line 1: You don't need a new Snap?

src/components/UserActivationPage/index.jsx
binodpant, line 52: to follow the pattern in place already should this be defined as a var too? `isActive !== undefined && isActive` for readability I guess could be called `isUserLoadedAndActive` ?

binodpant:
@albemarle  this worth a ticket? 
> I’m skipping one test that I couldn’t get working; I confirmed manually that what it’s testing for works; ideas on that are welcome.

edx-abolger (commented):
A few questions before I approve, nice lifting here!

src/components/AdminRegisterPage/index.jsx
adamstankiewicz: Using the `apiClient` from frontend-auth previously provided a `logout` function that would redirect the user to logout page, and setting the `?next=` query param to whatever URL is passed in as an argument.

Essentially, the behavior we wanted here is largely when the user logs out, they should be brought to the branded logistration page for their Enterprise and this is done by passing `?next=` query param.

src/components/App/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+              authenticatedAPIClient={apiClient}
+              redirect={process.env.BASE_URL}
+            />
+            <AuthenticatedPageRoute
+              path="/"
+              render={(routerProps) => <EnterpriseIndexPage {...routerProps} />}
+              authenticatedAPIClient={apiClient}
+              redirect={process.env.BASE_URL}
+            />
+            <PageRoute component={NotFoundPage} />
```
adamstankiewicz, line 64: Just to confirm, the NotFoundPage will appear if the route doesn't match any of these other paths? (ie, i'm wondering if this route needs a `path` prop?)

src/components/EnterpriseList/__snapshots__/EnterpriseList.test.jsx.snap
adamstankiewicz: The `value` of the searchbox is now empty (line 399) which means no active search is applied. the button that is no longer rendering is the clear icon which makes sense if no query is applied.

that said, @raq929 this snapshot seems to check that it "renders correctly with search query and empty enterprises data" so I'm confused why the search query is now empty.

src/components/EnterpriseList/index.jsx
adamstankiewicz: This is correct. This is just passing route props into the component, but doesn't change anything else.

src/components/ErrorBoundary/ErrorBoundary.test.jsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
```
adamstankiewicz, line 1: If we're no longer using `withErrorBoundary`, is there a need for `ErrorBoundary` related components/tests? Aka can we kill more things with 🔥

src/components/Header/Header.test.jsx
adamstankiewicz: We'll get there :)

src/components/Header/index.jsx
adamstankiewicz: We don't have a strict standard, usually just a judgement call based on how complex/large the components are. E.g., Paragon has components that utilize sub-components, but not all those sub-components have their own file.

src/components/AdminRegisterPage/index.jsx
adamstankiewicz: @raq929  That said, frontend-platform provides a `getLogoutRedirectUrl` that should likely be used here instead:

```
const proxyLoginUrl = getProxyLoginUrl(enterpriseSlug);
const logoutUrl = getLogoutRedirectUrl(proxyLoginUrl);
```

https://edx.github.io/frontend-platform/module-Auth.html#~getLogoutRedirectUrl

src/components/ErrorBoundary/ErrorBoundary.test.jsx
adamstankiewicz: Correct, frontend-platform "owns" the New Relic logging through the `logInfo` and `logError` functions

src/components/Header/index.jsx
```diff
@@ -1,102 +1,120 @@
...
+        src={avatarImage}
+        id="avatar-dropdown"
+        size="md"
+        alt={avatarScreenReaderText}
+      >
+        {username}
+      </Dropdown.Toggle>
+      <Dropdown.Menu>
+        <Dropdown.Item
+          href={getLogoutRedirectUrl()}
```
adamstankiewicz, line 56: We should pass in the url to the branded logistration page as an argument to `getLogoutRedirectUrl` here so that the user is redirected to the branded logistration page after logout.

Ssee `getProxyLoginUrl` function.

src/components/Header/index.jsx
```diff
@@ -1,102 +1,120 @@
...
-}
+    // rehydrate if the username changes
+  }, [user?.username, user?.id]);
+
+  return (
+    <header className="container-fluid border-bottom">
+      <Navbar aria-label="header" className="px-0 py-1 justify-content-between">
+        <Nav>
+          {hasSidebarToggle && <SidebarToggle />}
+
```
adamstankiewicz, line 92: nit: extra line

src/components/Header/index.jsx
```diff
@@ -1,102 +1,120 @@
...
+          <Nav.Link
+            href="/"
+            className="navbar-brand"
+          >
+            <Logo enterpriseLogo={enterpriseLogo} enterpriseName={enterpriseName || ''} />
```
adamstankiewicz, line 97: nit: Is the ` || ''` needed for the `enterpriseName` prop, or could `Logo.defaultProps` default it to the empty string?

src/components/LmsConfigurations/DegreedIntegrationConfigForm.test.jsx
```diff
@@ -6,6 +6,7 @@ import React from 'react';
 import snakeCase from 'lodash/snakeCase';
 import DegreedIntegrationConfigForm from './DegreedIntegrationConfigForm';
 import LmsApiService from '../../data/services/LmsApiService';
+// eslint-disable-next-line no-unused-vars
```
adamstankiewicz, line 9: nit: Is this line still needed?

src/components/App/index.jsx
raq929: It's not new so much as moved. I can add tests, but in the interests of not ballooning how much time I spent on this, I did not.

src/components/AdminRegisterPage/index.jsx
raq929: using `history.push()` is the most common React pattern for updating a url. I agree it's uh, poorly named.

src/components/AdminRegisterPage/index.jsx
raq929: Yep, that is correct. I'm removing this entirely so it's moot here, though. :)

src/containers/EnterpriseApp/__snapshots__/EnterpriseApp.test.jsx.snap
```diff
@@ -41,159 +41,4 @@ exports[`EnterpriseApp renders error page correctly 1`] = `
...
-          <p>
-            Either something went wrong or the page doesn't exist anymore.
-          </p>
-        </div>
-      </div>
-    </main>
-  </div>
-</div>
-`;
+exports[`<EnterpriseApp /> renders not found page correctly 1`] = `null`;
```
adamstankiewicz, line 44: Is it expected that this snapshot is now resulting in `null` vs. the 404? This snapshot is for the case when the `enterpriseSlug` is missing from the route I believe.

If so, should we remove that test altogether?

src/components/Header/index.jsx
adamstankiewicz, line 82: This is probably a future-us thing, but I'm wondering if we should split out hydration from the `Header` as it feels a bit odd that the header needs to be responsible for hydrating the user.

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
```diff
@@ -118,7 +118,7 @@ describe('EnterpriseApp', () => {
         document.body.removeChild(div);
       }
     });
-    it('handles location change properly', () => {
+    it.skip('handles location change properly', () => {
```
adamstankiewicz, line 121: This test is failing here, yeah?

```
// ensure focus is set on content wrapper
expect(document.activeElement.className).toEqual('content-wrapper');
```

So long as the content wrapper is actually getting focus on route change, skipping this test is fine for now :)

src/components/InviteLearnersModal/index.jsx
raq929: Good question, I think it's something we can talk about with @davidjoy and #fed-ex. I actually just had a conversation with people today about what kinds of utility functions belong there. I think for validators like this, I lean towards not rolling our own and moving to a third-party library. (And for forms in general, because we sure do have a lot of forms, and there are some nice libraries out there to handle stuff for us. Also just had a conversation today about formik...)

src/components/UserActivationPage/index.jsx
raq929: That component cares if the email is there, this one cares about whether the account is active or not. Both are provided by hydration. (Not sure if this explains it...)

src/components/UserActivationPage/index.jsx
raq929: I tried it in the console and I can compare it like this, so, I'm guessing yes? I think you're thinking about `typeof`?

raq929: I don't feel strongly about it. Do you?

src/containers/EnterpriseApp/__snapshots__/EnterpriseApp.test.jsx.snap
raq929, line 44: I need to look into this one...

src/components/App/index.jsx
raq929: Agree, I usually do that too. I think we should make the linter enforce it, though, because otherwise I forget. (And it takes up code review time) :)

src/components/EnterpriseList/__snapshots__/EnterpriseList.test.jsx.snap
raq929: Oooooh, yes I think this is actually broken. Will investigate, thanks.

src/containers/Header/__snapshots__/Header.test.jsx.snap
raq929: I removed all snapshot tests from this component in favor of unit tests. We've kind of agreed as a group that snapshot tests aren't serving our purposes well. I actually added two more unit tests that capture things we weren't getting from the snapshots.

src/components/App/index.jsx
raq929: Yep! Tested it just now to be sure.

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
raq929: Yeah, I can't figure it out. It's working via manual testing. Ideas welcome.

src/components/Header/index.jsx
raq929: YES. I will open a technical-debt ticket.

src/containers/EnterpriseApp/__snapshots__/EnterpriseApp.test.jsx.snap
raq929: I'm in favor of removing snapshot tests in general. I'll look into adding a unit test that tests the right thing.

src/components/App/index.jsx
binodpant: yeah linter is a good idea.. can check

src/components/ErrorBoundary/ErrorBoundary.test.jsx
binodpant: wohoo!

src/components/Header/index.jsx
binodpant: my take: it is okay to have multi component in a single file if they are always expected to be used together and are very close in what they do in the ui. Generally I would go for one file per component, but yeah it's a bit subjective when this kind of grouping is better, in this case it seems like it is a fine choice given how small and closely knit together they are

src/components/EnterpriseList/__snapshots__/EnterpriseList.test.jsx.snap
raq929: Update: Not actually broken when I test manually, but the snapshot sure it. =/

src/components/Header/Header.test.jsx
binodpant: yeah I love this discussion! Sure we should def bring up ideas and chat about this area! Maybe build up/improve testing docs as we work on new areas and doc/recommend where enzyme may be better suited. my default take is always to go RTL since it forces you to think a bit differently focusing only on what users can see or interact with. It also seems to try to prevent you mostly from trying to access internal state etc. Which is great in my opinion. And we have hooks testing lib for most of the rest, plus plain jest tests for utils. That may leave some balance/room though where RTL does not do well, and we can identify and use enzyme for those. I still make it a point to largely stay away from enzyme these days unless there is a compelling reason to use it (I am certain those reasons exist in places!)

src/components/UserActivationPage/index.jsx
binodpant: nah, given only one usage here not really, so yeah for this PR let's let go of it. I had initially thought I saw more than one usage of it, but since no, it does not really matter here...

src/containers/EnterpriseApp/__snapshots__/EnterpriseApp.test.jsx.snap
binodpant: I do not like snapshot tests! I think once we have some cypress tests, and well written hooks and ui tests, we should feel less of a need for these (or no need hopefully)

src/components/Header/Header.test.jsx
adamstankiewicz: Sounds like a good potential discussion topic for a future Enterprise Architecture Working Group office hours? :)

src/components/UserActivationPage/index.jsx
binodpant: yeah I think Adrienne is correct, 

```
typeof x
"undefined"


x === undefined
VM570:1 Uncaught ReferenceError: x is not defined
    at <anonymous>:1:1
```

typeof should be the safest optoin generally 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description

src/components/Header/index.jsx
adamstankiewicz: Ignore this. This will be tackled via follow-up work ENT-4040!

src/components/UserActivationPage/index.jsx
binodpant: having said that, maybe the destructure above actually defined isActive to undefined?

src/components/SamlProviderConfiguration/index.jsx
```diff
@@ -34,7 +35,7 @@ class SamlProviderConfiguration extends React.Component {
       .catch((error) => {
         const errorMsg = error.message || error.response.status === 500
           ? error.message : JSON.stringify(error.response.data);
-        NewRelicService.logAPIErrorResponse(errorMsg);
+        logError.logAPIErrorResponse(errorMsg);
```
adamstankiewicz, line 38: Is this supposed to be just `logError` vs `logError.logAPIErrorResponse`?

src/components/UserActivationPage/index.jsx
binodpant: yup!!

```
const { x } = {}
undefined
if (x) { console.log('x') }
undefined
```

src/components/subscriptions/licenses/TabContentTable.jsx
```diff
@@ -56,8 +56,10 @@ const TabContentTable = ({ enterpriseSlug }) => {
     errors,
     setErrors,
   });
+
   const hasErrors = Object.values(errors).length > 0;
-  const hasLoadedUsers = !!(users?.numPages);
+  const hasLoadedUsers = !!(users?.numPages || users?.count);
```
adamstankiewicz, line 61: Note to self: file ticket for understanding why the endpoint in license-manager isn't returning these counts properly.

src/data/store.js
```diff
@@ -2,55 +2,16 @@ import { applyMiddleware, createStore } from 'redux';
...
-import { PAGINATION_REQUEST, SORT_REQUEST } from './constants/table';
 
-import apiClient from './apiClient';
-import history from './history';
 import reducers from './reducers';
 
 const loggerMiddleware = createLogger();
-const routerHistoryMiddleware = routerMiddleware(history);
 
-const eventsMap = {
```
adamstankiewicz, line 18: Just leaving a note for additional context on the removal of this `eventsMap` and `segmentMiddleware` below. We are intentionally removing tracking of the CSV downloads and table pagination/sorting as they are not actually used downstream for any product analytics afaik.

More importantly, we'd rather do analytics the right way with frontend-platform rather than dispatching these analytics events using `@redux-beacon/segment` as we were doing here.

src/hooks/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as useInterval } from './useInterval';
+export { default as useWindowSize } from './useWindowSize';
```
adamstankiewicz, line 2: Not for this PR, but is `useWindowSize` the same one that Paragon has? Should we remove `useWindowSize` here in favor of the one exported by Paragon now?

src/index.jsx
```diff
@@ -4,32 +4,33 @@ import 'regenerator-runtime/runtime';
...
+initialize({
+  messages: [],
+  // We don't require authenticated users so that we can perform our own auth redirect to a proxy login that depends on
+  // the route, rather than the LMS like frontend-platform does.
+  requireAuthenticatedUser: false,
+  hydrateAuthenticatedUser: true,
+});
+
+// eslint-disable-next-line import/first
 import './index.scss';
```
adamstankiewicz, line 34: Should this import be at the top of the file?

src/index.jsx
```diff
@@ -4,32 +4,33 @@ import 'regenerator-runtime/runtime';
...
+});
+
+subscribe(APP_INIT_ERROR, (error) => {
+  ReactDOM.render(<ErrorPage message={error.message} />, document.getElementById('root'));
+});
+
+initialize({
+  messages: [],
+  // We don't require authenticated users so that we can perform our own auth redirect to a proxy login that depends on
+  // the route, rather than the LMS like frontend-platform does.
```
adamstankiewicz, line 28: nit: I'm guessing this might be copy/paste from the Learner Portal but Admin Portal doesn't have use its own auth redirect to proxy login quite yet (filed ENT-4040 to cover this).

Should we just remove this comment for now?

src/components/Header/index.jsx
raq929: Aaaah it was `null` in the defaultProps and I think that was causing issues. Updating.

src/components/Header/Header.test.jsx
raq929: Yeah, and I'm super on board with the way `react-testing-library` makes you think about things. It's great in that way. 
And yeah, in our repos, I think we should lean heavily towards it. I think enzyme shines more when you're creating a generic reusable component that you need to test is passing props around as it should. Serious unit testing. :)

src/components/UserActivationPage/index.jsx
raq929: Yes, destructuring sets `isActive` to undefined.

src/components/App/index.jsx
raq929: https://openedx.atlassian.net/browse/ENT-4046

src/components/Header/index.jsx
raq929: https://openedx.atlassian.net/browse/ENT-4047

src/components/SamlProviderConfiguration/index.jsx
raq929: We found that `logAPIErrorResponse` does not actually exist in `frontend-platform`, so this is what we get. :)

src/hooks/index.js
raq929: It is, we should. The paragon one has tests ;)

raq929: https://openedx.atlassian.net/browse/ENT-4048

src/index.jsx
raq929: I suspect it needs to be imported after the app is initialized? It's that way in learner portal, so I'm not gonna mess with it for now.

src/components/EnterpriseList/__snapshots__/EnterpriseList.test.jsx.snap
raq929: Went with unit tests instead.

src/components/EnterpriseApp/index.jsx
```diff
@@ -90,18 +90,18 @@ class EnterpriseApp extends React.Component {
...
       url: baseUrl,
       params: { enterpriseSlug },
     } = match;
-    const defaultContentPadding = 10; // 10px for appropriate padding
 
-    const isUserLoadedAndInactive = !!(userAccount?.loaded && !userAccount?.isActive);
-    const isUserMissingJWTRoles = !authentication?.roles?.length;
+    const defaultContentPadding = 10; // 10px for appropriate padding
+    const { isActive, roles, email } = getAuthenticatedUser() || {};
+    // checking for undefined tells if if the user's info is hydrated
```
adamstankiewicz, line 102: tiniest of the tiniest nits: two "ifs" 😄

src/components/EnterpriseList/index.jsx
```diff
@@ -12,6 +12,8 @@ import SearchBar from '../SearchBar';
 import SurveyPage from '../SurveyPage';
 import { updateUrl } from '../../utils';
 
+export const TITLE = 'Enterprise List';
```
adamstankiewicz, line 15: This title has annoyed me for the longest time. Can we change it to "My Enterprises" 😉

src/index.jsx
adamstankiewicz: Weird, sounds good to me haha.

adamstankiewicz (commented):
I think my only remaining question/feedback is:

Were tests that lived in now-deleted containers moved to their respective component directory? e.g. src/containers/UserActivationPage/UserActivationPage.test.jsx

adamstankiewicz (approved):
🎉 🚀

binodpant (approved):

PR #452, opened 2021-02-01
Fix routing after admin-platform merge

raq929:
Two small routing fixes

### DIFF
--- a/src/components/AdminRegisterPage/index.jsx
+++ b/src/components/AdminRegisterPage/index.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { Redirect, useHistory, withRouter } from 'react-router-dom';
+import { Redirect, withRouter } from 'react-router-dom';
 import { Container, Row, Col } from '@edx/paragon';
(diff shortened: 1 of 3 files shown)


edx-abolger (approved):

src/components/AdminRegisterPage/index.jsx
```diff
@@ -28,7 +27,7 @@ const AdminRegisterPage = ({ match }) => {
         // user is authenticated but doesn't have the `enterprise_admin` JWT role; force a log out so their
         // JWT roles gets refreshed. on their next login, the JWT roles will be updated.
         const logoutRedirectUrl = getLogoutRedirectUrl(getProxyLoginUrl(enterpriseSlug));
-        history.push(logoutRedirectUrl);
+        getAuthenticatedHttpClient().get(logoutRedirectUrl);
```
adamstankiewicz, line 30: Are we now using `getAuthenticatedHttpClient.get()` (aka axios instance) to trigger a redirect? Feels a little strange since we're not really making an API request here.

Should we revert back to using `global.location.href = logoutRedirectUrl`?

adamstankiewicz (approved):

PR #459, opened 2021-02-03
fix: attempting to fix build error on deploy

adamstankiewicz:
it seems that terser may be trying to minify some non-es5 code when `npm run build` is run in GoCD for the past 3 build attempts.

im hoping that adding the @babel/preset-env preset option to our webpack config's babel-loader will ensure babel is correctly using that preset to ensure everything is transpiled :crossed_fingers:


chavesj (approved):
🚢

PR #461, opened 2021-02-04
Reverts back to last successful stage deploy and re-installs node_modules / package-lock.json so they are "fresh"; update babel-loader exclude

adamstankiewicz:
This PR attempts to unblock the stage pipeline by reverting all commits up to the commit associated with the last successful stage build.

Everything that is reverted here is fine to revert, as its mostly just related to the maintenance banner.

I also pinned a couple unpinned dependencies, and deleted local `node_modules` and package-lock.json file before re-running `npm i` so package-lock.json will be "fresh"(-er).

Lastly, I updated our webpack configs to slightly to align more towards what edx/frontend-build does with regard to the include/exclude for `babel-loader`.


config/webpack.prod.config.js
```diff
@@ -35,12 +35,7 @@ module.exports = merge(commonConfig, {
       // Babel is configured with the .babelrc file at the root of the project.
       {
         test: /\.(js|jsx)$/,
-        include: [
-          path.resolve(__dirname, '../src'),
-        ],
-        exclude: [
-          path.resolve(__dirname, '../node_modules'),
-        ],
+        exclude: /node_modules\/(?!edx))/,
```
adamstankiewicz, line 38: 🤞

package.json
```diff
@@ -70,19 +70,19 @@
...
-    "html-webpack-new-relic-plugin": "^2.0.0",
+    "dotenv": "8.2.0",
+    "dotenv-webpack": "3.0.0",
+    "webpack": "4.44.2",
+    "webpack-merge": "5.4.0",
+    "mini-css-extract-plugin": "0.11.2",
+    "html-webpack-new-relic-plugin": "2.0.0",
     "html-webpack-plugin": "4.5.0",
-    "clean-webpack-plugin": "^3.0.0"
+    "clean-webpack-plugin": "3.0.0"
```
adamstankiewicz, line 80: Note: these are only peerDeps to avoid linting errors in our custom webpack configs. These dependencies are installed via the edx/frontend-build `devDependency`.

kellmarie (approved):

PR #465, opened 2021-02-06
feat: introduce CardGrid and clean up batch page layout 

adamstankiewicz:
# Before
The grid of cards do not have equal heights when rows have different lengths of content.

The "Need help?" sidebar distracts from the user's main goal on this page (to quickly find and browse the learners associated with a particular batch) as it shares the same styles as the Cards, making is difficult to distinguish between the sidebar and the cards, especially when there is a full grid of cards.

image

# After
I removed the "Need help?" sidebar in favor of a persistent "Help" link in the header. Iterated with Ange to land on "Help" link in header this morning.

Uses the new `CardGrid` component from Paragon for a full-width 3x3 responsive grid! 😄

image


src/components/subscriptions/styles/_SubscriptionManagementPage.scss
```diff
@@ -12,6 +12,10 @@
     }
   }
 
+  .subscription-card {
+    // width: 100%;
+  }
```
adamstankiewicz, line 17: oops, removing this

chavesj (approved):
LGTM 👍

PR #479, opened 2021-02-22
Adding loading to the base Enterprise App

alex-sheehan-edx:
When the admin portal first refreshes/loads, we fetch the configurations for each potential tab the user/enterprise has setup. These configurations are setup so that if none are loaded, we default to the `NotFound` base component, which has created the situation where there is the small amount of time between the initial loading of the page and the fetching of the configurations where nothing is set, so we flash the `Not Found` page. 

The fix here is pretty straight forward- change the default loading state of fetch configurations redux to `true` and render in a loading component to the Enterprise App until that loading prop is `false`. This way we still have a not found page if the user is somewhere weird, but don't flash it to the user if they refresh the page.

**What we don't want:**
image

**What we want:**
image

Pretty sure there's some straightforward testing to do here, but would love some pointers on where to look

### DIFF
--- a/src/components/EnterpriseApp/index.jsx
+++ b/src/components/EnterpriseApp/index.jsx
@@ -90,6 +90,7 @@ class EnterpriseApp extends React.Component {
       enableAnalyticsScreen,
       enableSamlConfigurationScreen,
       enableLmsConfigurationsScreen,
+      loading,
     } = this.props;
     const { sidebarWidth } = this.state;
     const {
@@ -113,6 +114,9 @@ class EnterpriseApp extends React.Component {
       );
     }
 
+    if (loading) {
+      return <LoadingMessage className="overview" />;
+    }
     return (
       <div className="enterprise-app">
         <MediaQuery minWidth={breakpoints.large.minWidth}>
@@ -261,6 +265,7 @@ EnterpriseApp.defaultProps = {
   enableSamlConfigurationScreen: false,
   enableAnalyticsScreen: false,
   enableLmsConfigurationsScreen: false,
+  loading: true,
 };
 
 EnterpriseApp.propTypes = {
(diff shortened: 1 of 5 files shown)


src/components/EnterpriseApp/index.jsx
```diff
@@ -113,6 +114,9 @@ class EnterpriseApp extends React.Component {
       );
     }
 
+    if (loading) {
```
binodpant, line 117: think it should be simple to test that by passing a loading:true to the `<EnterpriseApp>` component that the output contains whatever matches the loading component. you can add a test at https://github.com/edx/frontend-app-admin-portal/blob/363a412774d680d5ee6f48214deb4c8e3cb0b2b4/src/containers/EnterpriseApp/EnterpriseApp.test.jsx#L48

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
```diff
@@ -105,6 +115,20 @@ describe('<EnterpriseApp />', () => {
     expect(wrapper.text()).toContain(404);
   });
 
+  it('renders the load page correctly', () => {
+    const store = mockStore({
+      ...initialState,
+      portalConfiguration: {
+        ...portalConfigurationBeforeLoad,
```
binodpant, line 122: not: since loading is the only variable of interest, is it a little bit tighter of a test to just change `loading` in the overrride, e.g. , 

```
portalConfiguration: { 
   { ...initialState.portalConfiguration, loading: true },
}
```

just to ensure that this test only focuses on loading as a state change.

binodpant (approved):
lgtm, minor suggestion made!

src/containers/EnterpriseApp/EnterpriseApp.test.jsx
alex-sheehan-edx: oooh good idea. 👍

PR #481, opened 2021-02-22
Fixes to frontend-build update

raq929:
Use the prod config for prod
Add back code-js so that things still work in ie11

**To test on stage before pushing to prod**

- [ ] Check all pages esp. Analytics which doesn't work locally because Tableau
- [ ]  Submit all forms (because this affects transpilation, it could potentially affect any code)
- [ ] Check admin registration

**Test on prod immediately after deploy**
- [ ] Analytics
- [ ] Customer support request

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4663,6 +4663,12 @@
         "regenerator-runtime": "^0.10.5"
       },
       "dependencies": {
+        "core-js": {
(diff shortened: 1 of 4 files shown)


adamstankiewicz (approved):
:shipit:

PR #484, opened 2021-02-25
fix: revert to openedx theme in package.json

adamstankiewicz:
Reverts the installation of @edx/brand-edx.org in favor of @edx/brand-openedx. 

@edx/brand-edx.org is installed at build time in GoCD via an npm alias, but is often installed during local development to make sure what we're doing will look good with the new theme. It's too easy to miss reverting the install of @edx/brand-edx.org, though.


binodpant (approved):
lgtm!

wondering how we can address this going forward

adamstankiewicz:
> lgtm!
> 
> wondering how we can address this going forward

Agreed :) Let's continue the conversation in Slack.

PR #485, opened 2021-02-25
feat: fix brand style issue by only importing course-search styles

binodpant:
JIRA: ENT-4205

Tested locally by:

* changing above import
* [1] verified openedx style applies by default (blue buttons, rounded corners etc)
* [2] did a `npm install --save @edx/brand@npm:@edx/brand-edx.org` to get overriden theme and verified new style (black-ish buttons with flat corners) now show
Screen Shot 2021-02-25 at 12 51 43 PM

will file another ticket for a slightly more future-proof frontend-enterprise change, but this change will suffice to address this issue for now


adamstankiewicz (approved):
:shipit: Thanks for jumping on this, @binodpant 🙌

PR #486, opened 2021-02-26
Update paragon version

raq929:
Resolves ENT-4094


binodpant:
does this on its own fix ENT-4094 or just part of the requirements?

raq929:
This fixes it entirely.

binodpant (approved):
if this works by itself, let's ship it!

PR #503, opened 2021-03-16
feat: ENT-4197: improve error messaging formatting, and also report to NewRelic/console on errors

binodpant:
Changes included to improve error messaging:

- [X] lists multiple validation errors from non_field_errors case instead of printing them raw which looks bad
- [X] Change error alert title from `Unable to assign codes` to `Unable to enroll learners`
- [X] justify text of alert, and make alert text-wrap for better viewing of content
- [X] remove the Cross button from the Alert message ( takes care of ENT-4200 )
- [X] some margin changes in reporting failed learners
- [X] logError to report to console and/or new relic when bulk enrollment fails
- [X] enhance reporting of message by reporting on the customAttributes available from frontend-platform instead of the raw `error.message`, it is a bit cleaner
- [X] some misc cleanup and .gitignore changes

### Work not included:
Another round of improvements will come once the new backend is in place since then we can make more 'intelligent' error messages (there is another ticket to decide error messages)

#### removed truncation so ignore that in below screenshots
Screen Shot 2021-03-17 at 10 11 52 AM

### some failure cases 
Screen Shot 2021-03-16 at 5 53 49 PM
Screen Shot 2021-03-16 at 3 32 34 PM
Screen Shot 2021-03-16 at 3 12 17 PM
Screen Shot 2021-03-16 at 10 06 45 PM

#### case of non_field_errors (bad fields, no input provided etc)
Screen Shot 2021-03-16 at 10 33 38 PM

### DIFF
--- a/src/components/BulkEnrollmentModal/index.jsx
+++ b/src/components/BulkEnrollmentModal/index.jsx
@@ -3,11 +3,16 @@ import PropTypes from 'prop-types';
 import {
   Field, reduxForm, SubmissionError, reset,
 } from 'redux-form';
+
+import { logError } from '@edx/frontend-platform/logging';
 import {
   Alert, Button, Icon, Modal,
+
 } from '@edx/paragon';
+
 import TextAreaAutoSize from '../TextAreaAutoSize';
 import CsvUpload from '../CsvUpload';
+import { truncateString } from '../../utils';
 
 class BulkEnrollmentModal extends React.Component {
   constructor(props) {
@@ -36,7 +41,7 @@ class BulkEnrollmentModal extends React.Component {
     }
 
     if (submitFailed && error !== prevProps.error && errorMessageRef) {
-      // When there is an new error, focus on the error message status alert
+      // When there is a new error, focus on the error message status alert
       errorMessageRef.focus();
     }
   }
@@ -71,8 +76,16 @@ class BulkEnrollmentModal extends React.Component {
         this.props.onSuccess();
       })
       .catch((error) => {
+        const { customAttributes } = error;
+        // fall back to the error.message instead of showing no information but prefer the
+        // cleaner info from customAttributes
+        const customErrMessage = truncateString(
+          customAttributes ? (customAttributes.httpErrorResponseData || error.message) : error.message,
+          240,
+        );
+        logError(error, { enterpriseUuid });
         throw new SubmissionError({
-          _error: [error.message],
+          _error: [customErrMessage],
         });
       });
   }
@@ -84,12 +97,9 @@ class BulkEnrollmentModal extends React.Component {
         ref={this.errorMessageRef}
         tabIndex="-1"
       >
-        <Alert
-          variant="danger"
-          dismissible
-        >
-          <Alert.Heading>Unable to assign codes.</Alert.Heading>
-          <p className="alert-body">{error}</p>
+        <Alert variant="danger">
+          <Alert.Heading>Unable to enroll learners</Alert.Heading>
+          <p className="alert-body text-break">{error}</p>
         </Alert>
       </div>
     );
@@ -101,18 +111,17 @@ class BulkEnrollmentModal extends React.Component {
         ref={this.errorMessageRef}
         tabIndex="-1"
       >
-        <Alert
-          variant="danger"
-          dismissible
-        >
+        <Alert variant="danger">
           <Alert.Heading>
             Before you can invite learners to a course they must have a subscription license.
           </Alert.Heading>
-          <p className="alert-body">
+          <p className="alert-body text-justify">
             The following emails addresses are not yet associated with a subscription.
             Please take note of the following learners and invite them from the subscription management page.
           </p>
-          {failedLearners.map((failedLearner) => <p className="alert-email" key={`${failedLearner}-email`}>{failedLearner}</p>)}
+          {failedLearners.map((failedLearner) => (
+            <p className="mb-1 alert-email" key={`${failedLearner}-email`}>{failedLearner}</p>
+          ))}
         </Alert>
       </div>
     );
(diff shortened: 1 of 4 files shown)


raq929 (changes requested):
We should not show the `non_field_errors` text to the user. There should be a section of the form which handles the display of `non_field_errors` specifically, as that is a standard type of error that Django Rest Framework returns.
It should look for `non_field_errors` in the errors returned and expect a list of errors. It should display that list to the user as bullets; similar to what you've done with the emails for the other error.

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -3,11 +3,16 @@ import PropTypes from 'prop-types';
 import {
   Field, reduxForm, SubmissionError, reset,
 } from 'redux-form';
+
+import { logError } from '@edx/frontend-platform/logging';
 import {
   Alert, Button, Icon, Modal,
+
```
raq929, line 10: nittiest of nits, don't bother unless you're making more changes; extra line here.

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -71,8 +76,16 @@ class BulkEnrollmentModal extends React.Component {
         this.props.onSuccess();
       })
       .catch((error) => {
+        const { customAttributes } = error;
+        // fall back to the error.message instead of showing no information but prefer the
+        // cleaner info from customAttributes
+        const customErrMessage = truncateString(
+          customAttributes ? (customAttributes.httpErrorResponseData || error.message) : error.message,
+          240,
```
raq929, line 84: What is this 240 doing?

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -71,8 +76,16 @@ class BulkEnrollmentModal extends React.Component {
         this.props.onSuccess();
       })
       .catch((error) => {
+        const { customAttributes } = error;
```
raq929, line 79: I wonder if we want to do something like the above where we have a props function handle the error state? I can see some reasons why and why not, tbh.

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -101,18 +111,17 @@ class BulkEnrollmentModal extends React.Component {
...
             Before you can invite learners to a course they must have a subscription license.
           </Alert.Heading>
-          <p className="alert-body">
+          <p className="alert-body text-justify">
             The following emails addresses are not yet associated with a subscription.
             Please take note of the following learners and invite them from the subscription management page.
           </p>
-          {failedLearners.map((failedLearner) => <p className="alert-email" key={`${failedLearner}-email`}>{failedLearner}</p>)}
+          {failedLearners.map((failedLearner) => (
+            <p className="mb-1 alert-email" key={`${failedLearner}-email`}>{failedLearner}</p>
```
raq929, line 123: If this is a list, it **must** be a list, not a seres of `p` tags, to be a11y compliant.
Please change to a series of `li` tags wrapped by a `ul`

src/components/CodeReminderModal/index.jsx
```diff
@@ -270,7 +270,7 @@ class CodeReminderModal extends React.Component {
             >
               <>
                 {mode === 'remind' && submitting && <Icon className="fa fa-spinner fa-spin mr-2" />}
-                {'Remind'}
+                Remind
```
raq929, line 273: 🎉

src/utils.js
```diff
@@ -306,6 +306,13 @@ const getSubscriptionContactText = (contactEmail) => {
   return `${contactText}.`;
 };
 
+function truncateString(str, maxStrLength = 10) {
```
raq929, line 309: Not for right now, but this seems like the type of util we might want to add to https://edx.github.io/frontend-platform/module-Utilities.html

src/components/BulkEnrollmentModal/index.jsx
binodpant: it’s the truncation limit, I’m still torn on truncating vs letting  it go as klong as it wants

src/components/BulkEnrollmentModal/index.jsx
binodpant: I see what you mean. The onSuccess callback is a prop input while the on failure behavior is not. So there is that discrepancy. However, thought another way, right now there is no action to be done outside of the dialog for error case so until then we can leave it here? we can always refactor if we need to handle error case outside the dialog as well

src/components/BulkEnrollmentModal/index.jsx
```diff
@@ -71,25 +75,45 @@ class BulkEnrollmentModal extends React.Component {
         this.props.onSuccess();
       })
       .catch((error) => {
+        const { customAttributes } = error;
+        // fall back to the error.message instead of showing no information but prefer the
+        // cleaner info from customAttributes
+        const customErrMessage = truncateString(
+          customAttributes ? (customAttributes.httpErrorResponseData || error.message) : error.message,
```
binodpant, line 82: this fields comes from the interceptors here, https://github.com/edx/frontend-platform/blob/00d109d12ab207adabd8ecb2b4c8883db00d7e17/src/auth/utils.js#L61 but it's not documented! So pushing in fedx to write/review that doc somewhere to make this official

binodpant:
Ran this by Enterprise UX sync today and no changes needed other than removing truncation, at this time

src/components/BulkEnrollmentModal/index.jsx
binodpant: removed truncation per chat today in UX

binodpant: will leave function in there in case it's handy

raq929 (approved):

PR #504, opened 2021-03-16
Add base_enterprise_url to reminder emails

raq929:
Also fixes an error in how assignment emails were getting information from the Redux store.

### DIFF
--- a/src/components/CodeAssignmentModal/index.jsx
+++ b/src/components/CodeAssignmentModal/index.jsx
@@ -327,7 +327,6 @@ class CodeAssignmentModal extends React.Component {
       template_closing: formData['email-template-closing'],
       enable_nudge_emails: formData['enable-nudge-emails'],
     };
-    console.log('ENABLE LEARNER PORTAL', enableLearnerPortal)
     // If the enterprise has a learner portal, we should direct users to it in our assignment email
     if (enableLearnerPortal && configuration.ENTERPRISE_LEARNER_PORTAL_URL) {
       options.base_enterprise_url = `${configuration.ENTERPRISE_LEARNER_PORTAL_URL}/${enterpriseSlug}`;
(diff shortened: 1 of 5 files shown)


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -24,7 +23,7 @@ import { EMAIL_ADDRESS_TEXT_FORM_DATA, EMAIL_ADDRESS_CSV_FORM_DATA } from '../..
 import './CodeAssignmentModal.scss';
 import { configuration } from '../../config';
 
-class BaseCodeAssignmentModal extends React.Component {
+class CodeAssignmentModal extends React.Component {
```
edx-abolger, line 26: +1 for having it match the file name I hate it when it doesnt.

src/components/CodeReminderModal/index.jsx
```diff
@@ -123,6 +126,11 @@ class CodeReminderModal extends React.Component {
       template_closing: formData['email-template-closing'],
     };
 
+    // If the enterprise has a learner portal, we should direct users to it in our assignment email
+    if (enableLearnerPortal && configuration.ENTERPRISE_LEARNER_PORTAL_URL) {
```
edx-abolger, line 130: Nit: Since I think the backend does the default can you add a comment here about what the default is

src/containers/CodeReminderModal/CodeReminderModal.test.jsx
```diff
@@ -153,6 +161,26 @@ describe('CodeReminderModalWrapper', () => {
...
+      }}
+      isBulkRemind
+    />);
+
+    expect(wrapper.find('.bulk-selected-codes').text()).toEqual('Selected Codes: 2');
+    expect(wrapper.find('#email-template')).toBeTruthy();
+    wrapper.find('.modal-footer .code-remind-save-btn').hostNodes().simulate('click');
+    const expectedData = codeReminderRequestData(2);
+    delete expectedData.base_enterprise_url;
+    expect(spy).toHaveBeenCalledWith(couponId, codeReminderRequestData(2));
```
edx-abolger, line 181: This test confuses me. Do we use expectedData?  I would have expected the test to have been called with expectedData, not a fresh call to the function

raq929: It should be called with `expectedData`, and in fact should probably not be passing. Good catch!

edx-abolger (approved):

PR #505, opened 2021-03-18
feat: adds CSV export button to SubscriptionDetailPage

chavesj:
### Description

* refactors invite learners button to be in line with subscription title
* adds DowloadCsvButton in line with LicenseAllocationHeader SearchBar
* adds functionality to export CSV file with license data

### Ticket link

[ENT-3986](https://openedx.atlassian.net/browse/ENT-3986)

### Testing instructions

* Checkout branch, run `npm install`, run `npm start`
* Navigate to the admin portal subscription management detail page
* Click the 'Download CSV' button and observe the contents of the downloaded file

### Screenshots

image

image


src/components/subscriptions/buttons/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const headers = [
+    { label: 'Status', key: 'status' },
+    { label: 'User Email', key: 'userEmail' },
+    { label: 'Activation Date', key: 'activationDate' },
+    { label: 'Last Remind Date', key: 'lastRemindDate' },
+    { label: 'License Activation Link', key: 'activationLink' },
+  ];
+
+  LicenseManagerApiService.fetchSubscriptionUsers(subscription.uuid, options, 5000)
```
adamstankiewicz, line 40: A few thoughts:
1. What happens if the subscriptions data has more than 5000 records? As is, this API call would only return the first 5000 in the downloaded CSV file. We get around this in other places in frontend-app-admin-portal by disabling pagination on the API call to fetch the CSV data (to get all pages) and have the UI download that response as a file. On the API side, this is done using the `rest_framework_csv.renderers.CSVRenderer` extension for django-rest-framework, which allows you to add `.csv` to an endpoint for it to return the data formatted as a CSV.
* https://github.com/mjumbewu/django-rest-framework-csv
* Example of disabling pagination on API request to fetch full CSV report: https://prod-edx-analyticsapi.edx.org/enterprise/api/v0/enterprise/8ea70d02-c210-4009-a86d-e23423dcda7d/enrollments.csv/?no_page=true
2. Do we need to fetch this data every time the component re-renders? Would it make more sense if the API call is made when the user clicks the download button (where the download button is a [StatefulButton](https://paragon-edx.netlify.app/components/statefulbutton/)). Even if we did fetch the data prior to button click, we'd only need to fetch the data once by using an `useEffect` hook.
3. There is an existing `DownloadCsvButton` component in this MFE that accepts a few props like a fetch method; was it possible to use/extend that existing component vs. creating a separate component? Either way, that existing component could be cleaned up / refactored to no longer rely on Redux and use `StatefulButton`, too, so I don't feel strongly that we absolutely _need_ to use the existing `DownloadCsvButton` component here.

src/components/subscriptions/buttons/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,76 @@
...
+        userEmail: license.user_email,
+        activationDate: license.activation_date,
+        lastRemindDate: license.last_remind_date,
+        activationLink: getLicenseActivationLink(license.activation_key),
+      }));
+      setCsvData(licenseData);
+    })
+    .catch((err) => {
+      logError(err);
+      // TODO: what should the UX be for error here?
```
adamstankiewicz, line 53: Unfortunately, we don't really have much in place for error handling CSV downloads in other places throughout the app either.

I'd say logging the error will suffice for now, but keep the TODO comment here :)

src/components/subscriptions/buttons/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+  return (
+    <CSVLink
+      className="btn btn-outline-primary"
+      data={csvData}
+      filename={getCsvFileName()}
+      headers={headers}
+    >
+      Download CSV
```
adamstankiewicz, line 63: Should the button have a download icon to the left of the text? E.g.:
image

src/components/subscriptions/buttons/DownloadCsvButton.jsx
chavesj: Button is now implemented using the `StatefulButton` component from Paragon and only fetches data on button click. The new license CSV endpoint in license-manager is now used to fetch the data in CSV format. PR for CSV endpoint: https://github.com/edx/license-manager/pull/200

src/components/subscriptions/buttons/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,57 @@
...
+      variant="outline-primary"
+      labels={{
+        default: 'Download CSV',
+        pending: 'Downloading',
+        complete: 'Downloaded',
+      }}
+      icons={{
+        default: <Icon className="fa fa-download" />,
+        pending: <Icon className="fa fa-spinner fa-spin" />,
+        complete: <Icon className="fa fa-check" />,
```
adamstankiewicz, line 33: nit: It's preferable to use SVG icons vs. icons controlled via classes (i.e., a font-based icon) as they are more scalable and less prone to charset issues (utf-8). Unfortunately, the Paragon `Icon` component doesn't support SVG icons from font-awesome as it is intended to be used with the icons hosted within Paragon itself (`@edx/paragon/icons`).

Example:

```jsx
import { faDownload, faSpinner } from '@fortawesome/free-solid-svg-icons';
import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';

return (
  <StatefulButton
    icons={{
     default: <Icon src={faDownload} />,
     pending: <Icon src={faSpinner} className="fa-spin" />,
     ...
    }}
  />
);
```

src/components/subscriptions/buttons/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,57 @@
...
+            const blob = new Blob([response.data], {
+              type: 'text/csv',
+            });
+            saveAs(blob, getCsvFileName());
+            setButtonState('complete');
+          })
+          .catch(err => {
+            setButtonState('default');
+            logError(err);
+            // TODO: what should the UX be for error here?
```
adamstankiewicz, line 50: We don't really have error handling on other CSV download buttons either, so leaving this as a TODO comment is fine for now IMO.

src/components/subscriptions/data/service.js
```diff
@@ -114,10 +114,10 @@ class LicenseManagerApiService {
     return LicenseManagerApiService.apiClient().get(url);
   }
 
-  static fetchSubscriptionUsers(subscriptionUUID, options) {
+  static fetchSubscriptionUsers(subscriptionUUID, options, pageSize = PAGE_SIZE) {
```
adamstankiewicz, line 117: nit: I think it makes sense to keep all the options for the endpoint in the `options` function argument rather than introducing a separate `pageSize` function argument here, especially if this new argument isn't getting used yet.

By keeping everything together in a single `options` object, you could pass everything straight to query parameters while still allowing an overridable default page size option:

```jsx
const queryParams = {
  page_size: PAGE_SIZE,
  ...options,
};
```

```jsx
fetchSubscriptionUsers(subscriptionUUID, { page: 1, page_size: 50 });
```

src/components/subscriptions/data/service.js
```diff
@@ -133,6 +133,11 @@ class LicenseManagerApiService {
     return LicenseManagerApiService.apiClient().get(url);
   }
 
+  static fetchSubscriptionLicenseDataCsv(subscriptionUUID) {
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}/subscriptions/${subscriptionUUID}/licenses/csv/`;
```
adamstankiewicz, line 137: :noice:

src/components/subscriptions/licenses/LicenseAllocationHeader.jsx
```diff
@@ -24,26 +17,18 @@ const LicenseAllocationHeader = () => {
         {' of '}
         {subscription.licenses?.total} licenses allocated
       </p>
-      <div className="my-3 row">
-        <div className="col-12 col-md-5 mb-3 mb-md-0">
+      <Row className="my-3">
+        <Col className="col-8">
```
adamstankiewicz, line 21: Does the search bar need to take up 8 columns when the user input is only really needs to be the length of a typical email address (I feel this will make the search bar take up more visual real estate than it needs)? Might recommend keeping it to 6 columns max for larger screens. 

Also, how does this look at smaller screen sizes since this will try to keep search bar and download button on the same row? I'm pretty sure we'd get a search bar that is not wide enough and a button than has some strange text wrapping behavior.

Previously, the search bar took up the full container width (12 columns) with the button left aligned below the search bar for small screens with the search bar collapsed to 5 columns for larger screens with the button right aligned in its column. I think we'll still want to keep that same sort of responsive behavior here.

src/components/subscriptions/licenses/LicenseAllocationHeader.jsx
```diff
@@ -24,26 +17,18 @@ const LicenseAllocationHeader = () => {
         {' of '}
         {subscription.licenses?.total} licenses allocated
       </p>
-      <div className="my-3 row">
-        <div className="col-12 col-md-5 mb-3 mb-md-0">
+      <Row className="justify-content-between">
+        <Col lg={6} xs={12} className="mt-2">
```
adamstankiewicz, line 21: nit: if both columns require a top margin, might consider putting `mt-2` on the parent `Row` instead.

adamstankiewicz (approved):
2 non-blocking nits.

src/components/subscriptions/licenses/LicenseAllocationHeader.jsx
chavesj: this top margin was included to handle the scenario where the columns stack (e.g. mobile view) - it adds margin in between the search bar and download button. As I'm writing this I now realize I could use a bottom margin on the search bar and that should accomplish the same thing 👍

PR #506, opened 2021-03-18
Create linked learners when assigning codes

raq929:
Makes a request to create pending users before sending them codes.


alex-sheehan-edx (approved):
This looks good to me but I also lack some of the frontend context that other quokkas may be able to provide! 😬

src/components/CodeAssignmentModal/index.jsx
```diff
@@ -351,7 +354,21 @@ class BaseCodeAssignmentModal extends React.Component {
...
-    return sendCodeAssignment(couponId, options)
+    let pendingEnterpriseUserData;
+    if (hasTextAreaEmails) {
+      pendingEnterpriseUserData = validEmails.map((email) => ({
+        user_email: email,
+        enterprise_customer: enterpriseUuid,
+      }));
+    } else {
+      pendingEnterpriseUserData = {
+        user_email: formData['email-address'],
```
taliaedX, line 365: Is this else needed? Where is this from if not the text area or csv (as we put either in the validEmails var)?
Actually I can't even find the email-address field in this file, so I'm extra confused.

src/data/services/LmsApiService.js
```diff
@@ -196,6 +198,10 @@ class LmsApiService {
     const url = `${LmsApiService.enterpriseCustomerUrl}${enterpriseId}/enterprise_learners/`;
     return LmsApiService.apiClient().post(url, options);
   }
+
+  static createPendingEnterpriseUsers(formData) {
+    return LmsApiService.apiClient().post(this.createPendingUsersUrl, formData);
```
taliaedX, line 203: Does `this` actually work with the way the class is setup?
Every other reference is `LmsApiService.function/varname`

taliaedX (approved):
2 quick questions; otherwise lgtm

src/components/CodeAssignmentModal/index.jsx
raq929: So, there are two different forms (but for reasons they are the same ReactForm? this is an architectural problem, but not one I'm dealing with now). One has a single form field for an email address; you get to it by clicking the assign link in the table, and the other is for bulk assignment, and that's the text area. So if we're bulk assigning, we need one option, if not, we use this. It confused me, too.

src/data/services/LmsApiService.js
raq929: IT SURE SEEMS TO BUT YOU ARE RIGHT and there was some weird error elsewhere from using this and I will change it, thanks!

src/components/CodeAssignmentModal/index.jsx
raq929: The `formData` is from redux-form, I used to know more about how it worked, but basically it creates the formData object when you hit submit and passes that data to this function.

PR #509, opened 2021-03-22
ENT-4197 : ux feedback changes 

binodpant:
ultra small change : fixes the sentence case issue listed in the ticket

verified a couple items already work (see ticket)

a few items were ticketed separately leaving just this piece (sentence case button) to be fixed in this PR


binodpant:
Sentence case is the correct one confirmed with @Jgregory closing

PR #511, opened 2021-03-23
Unreverts previously reverted code & updates to new endpoint

raq929:
Allows admins to create new pending users for their enterprises


taliaedX (approved):
LGTM

PR #519, opened 2021-03-30
ENT-4310: Allow Enterprise Customers to Select Catalogs to receive content metadata reports

saleem-latif:
__Description:__
As an enterprise customer, I may receive multiple reports and consume different types of data, including course metadata and course run metadata. 
In the admin dashboard today, it is possible for customers to configure catalog reports to be sent to them. However, they do not have the ability to select specific catalogs they would like to receive. See the pied piper admin dashboard for an example: https://portal.edx.org/pied-piper-non-integrated-test/admin/reporting
It would be helpful if there was a way for customers to select which catalog's metadata they would like to receive when configuring a report.

### DIFF
--- a/src/components/ReportingConfig/ReportingConfigForm.jsx
+++ b/src/components/ReportingConfig/ReportingConfigForm.jsx
@@ -131,7 +131,7 @@ class ReportingConfigForm extends React.Component {
   }
 
   render() {
-    const { config } = this.props;
+    const { config, availableCatalogs } = this.props;
     const {
       frequency,
       invalidFields,
@@ -139,6 +139,7 @@ class ReportingConfigForm extends React.Component {
       active,
       submitState,
     } = this.state;
+    const selectedCatalogs = (config?.enterpriseCustomerCatalogs || []).map(item => item.uuid);
 
     return (
       <form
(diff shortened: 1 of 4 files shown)


src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -327,6 +328,27 @@ class ReportingConfigForm extends React.Component {
...
+            <Input
+              type="select"
+              id="enterpriseCustomerCatalogs"
+              name="enterpriseCustomerCatalogs"
+              multiple
+              defaultValue={selectedEnterpriseCustomerCatalogs}
+              options={
+                availableEnterpriseCustomerCatalogs.map(item => ({
+                  value: item.uuid,
+                  label: `Enterprise Customer Catalog "${item.title}" with UUID "${item.uuid}"`,
```
muhammad-ammar, line 346: Isn't the label is too long? Wha about just `catalog title` and `uuid`?

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -131,14 +131,15 @@ class ReportingConfigForm extends React.Component {
...
-    const { config } = this.props;
+    const { config, availableEnterpriseCustomerCatalogs } = this.props;
     const {
       frequency,
       invalidFields,
       deliveryMethod,
       active,
       submitState,
     } = this.state;
+    const selectedEnterpriseCustomerCatalogs = (config?.enterpriseCustomerCatalogs || []).map(item => item.uuid);
```
sameenfatima78, line 142: Same as above, maybe we can change this variable name to `selectedCatalogs`

muhammad-ammar (approved):

sameenfatima78 (approved):
Some minor suggestions, otherwise LGTM

PR #520, opened 2021-03-30
refactor: remove unused endpoint.

taliaedX:
I should have removed this when swapping our endpoints to use the bulk enrollment endpoint via License Manager, but I forgot!
(See this pr: https://github.com/edx/frontend-app-admin-portal/pull/510)

PS: If you CTRL+F (or CMD+F) through the codebase you may think we're still using this API call but I assure you we aren't. We just named everything exactly the same! 😑


alex-sheehan-edx (approved):
🚢 👍 ✅

PR #521, opened 2021-03-30
fix: license reminder email doesn't need subject

taliaedX:
When consolidating validation (done as a side effect of this pr: https://github.com/edx/frontend-app-admin-portal/pull/510) I missed a call site in the LicenseRemindModal that actually passed the `isSubjectRequired=false` flag.

As this validation function is used in 2 other Code related modals that *do* require subject, I just created a passthrough parameter to the field validation function.

### DIFF
--- a/src/data/validation/email.js
+++ b/src/data/validation/email.js
@@ -111,10 +111,10 @@ is invalid. Please try again.`;
   return errorsDict;
 };
 
-const validateEmailTemplateForm = (formData, templateKey) => {
+const validateEmailTemplateForm = (formData, templateKey, isSubjectRequired = true) => {
   // Takes redux form data and a string template key
   // Throws an error or otherwise returns nothing.
-  const errorsDict = validateEmailTemplateFields(formData, templateKey);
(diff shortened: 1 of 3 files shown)


src/data/validation/email.js
```diff
@@ -111,10 +111,10 @@ is invalid. Please try again.`;
   return errorsDict;
 };
 
-const validateEmailTemplateForm = (formData, templateKey) => {
+const validateEmailTemplateForm = (formData, templateKey, subjectRequired = true) => {
```
georgebabey, line 114: nit: Consider keeping naming consistent with the above `validateEmailTemplateFields()` method by using `isSubjectRequired`?

georgebabey (approved):

src/data/validation/email.js
taliaedX: Fair point

taliaedX:
Retroactive ticket creation:  https://openedx.atlassian.net/browse/ENT-4395

PR #522, opened 2021-03-31
Adding notify learners checkbox to the bulk enrollment modal

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4340)

Default state of modal upon opening:
image

Goals: Add a checkbox to the bulk enrollment modal that allows the user to enable or disable the option to notify the learners who are being enrolled. Functionally, dictate the `notify` parameter sent to the License manager's validation endpoint from the value by the new checkbox.

### DIFF
--- a/src/components/ReduxFormCheckbox/ReduxFormCheckbox.test.jsx
+++ b/src/components/ReduxFormCheckbox/ReduxFormCheckbox.test.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import ReduxFormCheckbox from './index';
+
+describe('<ReduxFormCheckbox />', () => {
+  it('renders checked correctly', () => {
+    const inputProp = { checked: true };
+    const component = renderer
+      .create((
+        <ReduxFormCheckbox input={inputProp} />
+      ))
+      .toJSON();
+    expect(component).toMatchSnapshot();
+  });
+  it('renders unchecked correctly', () => {
+    const inputProp = { checked: false };
+    const component = renderer
+      .create((
+        <ReduxFormCheckbox input={inputProp} />
+      ))
+      .toJSON();
+    expect(component).toMatchSnapshot();
+  });
+});
--- a/src/components/ReduxFormCheckbox/index.jsx
+++ b/src/components/ReduxFormCheckbox/index.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form, ValidationFormGroup } from '@edx/paragon';
+
+const ReduxFormCheckbox = ({
+  input,
+  label,
+}) => (
+  <ValidationFormGroup
+    for="active"
+  >
+    <Form.Check
+      {...input}
+      type="checkbox"
+      id="active"
+      name="active"
+      checked={input.checked}
+      label={label}
+    />
+  </ValidationFormGroup>
+);
+
+ReduxFormCheckbox.propTypes = {
+  label: PropTypes.string.isRequired,
+  input: PropTypes.shape({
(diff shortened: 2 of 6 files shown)


src/components/ReduxFormCheckbox/ReduxFormCheckbox.test.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import ReduxFormCheckbox from './index';
+
+describe('<ReduxFormCheckbox />', () => {
```
binodpant, line 6: Also run this by Lael but I question the value of this snapshot test that only checks that checking/unchecking changes the reduxformcheckbox markup. Isn't that just part of the component's features do we need to test it?

I think perhaps it's more usefl to test that checking/unchecking it changes the state of the params that are sent to the server? Did we already test that separately?

taliaedX: There may be additional context or nuance I'm missing, but from what I experienced trying to test Bulk Enrollment, Redux doesn't store things as accessible state, it passes them around as props. Thus, you can't really do an integration test of the "pretend I'm a user and click things" variety. Everything would be passed around as props, so a snapshot test is a clean version of verifying the checkbox renders when passing in relevant props.

raq929: Agree.
I'm gonna work on better testing of these forms when I refactor. In their current state, integration testing is well nigh impossible.

raq929: Basically, this tests: if we pass the right props to this component, does it do the right thing? 
We're input.checked here, that's what needs to be tested, everything else is a responsibility either of Paragon or Redux-Forms

src/components/ReduxFormCheckbox/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+    for="active"
+  >
+    <Form.Check
+      {...input}
+      type="checkbox"
+      id="active"
+      name="active"
+      className="ml-3"
+      checked={input.checked}
+      label={DEFAULT_CHECKBOX_LABEL}
```
raq929, line 20: This checkbox should be reuseable; the label should be passed in from the Field component (as it is for the similar Textarea component, etc.)
The constant should live where the other constants for this form live, and be called something more specific, like `NOTIFY_LEARNERS_LABEL`

raq929: (I believe if you just set the label on the Field component it will do the right thing and pass it to this checkbox.)

raq929 (changes requested):
Looking good! One quick change to make the checkbox component re-useable, and please get rid of the left margin on the checkbox, it looks weird to me. 
🎉

src/components/ReduxFormCheckbox/index.jsx
alex-sheehan-edx: FWIW the labels for these Fields are hard coded within the Bulk enrollment modal index. I'll move all of them to the constants file.

src/components/ReduxFormCheckbox/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+  <ValidationFormGroup
+    for="active"
+  >
+    <Form.Check
+      {...input}
+      type="checkbox"
+      id="active"
+      name="active"
+      checked={input.checked}
+      label={label}
```
raq929, line 18: Yay! Looks super-reuseable now!

raq929 (approved):

PR #526, opened 2021-04-14
Add description test id to TextAreaAutoSize

raq929:
Required for cypress tests


taliaedX (approved):
weird. I feel like Cypress should be able to handle dynamic ids rather than having to use a static one, but for now I guess this resolves the issue at least.

PR #533, opened 2021-04-23
feat: add loading state and redirect to pages that depend on the subscription picker

raq929:
Instead of flashing a 404 while we are loading data, this adds a loading state and loading skeleton to the CourseSearch and SubscriptionDetailsPage

In addition, instead of showing a 404 when the subscription cannot be found, the user is redirected back to the subscription picker (or to their one subscription if they only have one.)

Also changes the route names for these two routes to constants because my constant misspelling of `managment` made my tests fail until I got frustrated enough to make it never happen again.

Skeleton catalog management mobile:
Screen Shot 2021-04-23 at 10 39 51 AM

Skeleton subs management mobile:
Screen Shot 2021-04-23 at 10 39 39 AM

Desktop catalog management:
Screen Shot 2021-04-23 at 10 38 36 AM

Desktop subs management:
Screen Shot 2021-04-23 at 10 38 24 AM

### DIFF
--- a/src/components/BulkEnrollmentPage/CourseSearch.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearch.jsx
@@ -12,6 +12,7 @@ import Skeleton from 'react-loading-skeleton';
 import CourseSearchResults from './CourseSearchResults';
 import { configuration } from '../../config';
 import { useSubscriptionFromParams } from '../subscriptions/data/contextHooks';
+import { ROUTE_NAMES } from '../EnterpriseApp/constants';
 
 const searchClient = algoliasearch(
   configuration.ALGOLIA.APP_ID,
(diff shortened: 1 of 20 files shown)


src/components/BulkEnrollmentPage/CourseSearch.jsx
```diff
@@ -1,15 +1,18 @@
...
+
 import { InstantSearch, Configure } from 'react-instantsearch-dom';
 import { SearchHeader, SearchData } from '@edx/frontend-enterprise';
+import Skeleton from 'react-loading-skeleton';
+
 import CourseSearchResults from './CourseSearchResults';
 import { configuration } from '../../config';
 import { useSubscriptionFromParams } from '../subscriptions/data/contextHooks';
-import { NotFound } from '../NotFoundPage';
+import { ROUTE_NAMES } from '../EnterpriseApp/constants';
```
binodpant, line 15: 🎉

binodpant (approved):
no flags from me! this looks good to avoid the 404 for the no subs case, and I like the constant for the url route names

edx-abolger (commented):
LGTM with the tests and new Route pattern, I like the collection of user facing stuff in one place. I will need tutoring on useEffect because I believe this code works based on tests and screenshots but I am not an expert on that and useEffect looks just a little magical to me right now.

PR #534, opened 2021-04-23
feat: Add admin page and table skeletons

raq929:
Skeletons for all!!

Learner Progress report desktop:
Screen Shot 2021-04-23 at 2 52 13 PM

Learner progress report mobile:
Screen Shot 2021-04-23 at 2 52 25 PM
Screen Shot 2021-04-23 at 2 52 33 PM

Whole app loading (before nav is available):
Screen Shot 2021-04-23 at 3 33 32 PM

Whole app mobile:
Screen Shot 2021-04-23 at 3 33 50 PM


src/components/TableComponent/index.jsx
```diff
@@ -70,7 +68,6 @@ class TableComponent extends React.Component {
       <div className={className}>
         <div className="row">
           <div className="col">
-            {loading && <TableLoadingOverlay />}
```
raq929, line 73: Can y'all double check my logic here? Based on lines 136 and 139, I don't think we can ever get to this state, but I want to be sure I haven't missed anything.

src/components/Admin/AdminCardsSkeleton.jsx
```diff
@@ -0,0 +1,15 @@
...
+import Skeleton from 'react-loading-skeleton';
+
+const AdminCardsSkeleton = () => (
+  <div
+    className="admin-cards-skeleton mb-3 d-md-flex w-100"
+  >
+    <Skeleton />
+    <Skeleton />
+    <Skeleton />
+    <Skeleton />
```
adamstankiewicz, line 11: Might be good to rely on Bootstrap's existing grid here vs. customizing the width/margin of each skeleton with CSS, e.g.:

```jsx
<Row>
  <Col><Skeleton /></Col>
  <Col><Skeleton /></Col>
  <Col><Skeleton /></Col>
  <Col><Skeleton /></Col>
</Row>
```

src/components/Admin/Admin.scss
```diff
@@ -10,3 +15,23 @@
 .search-label {
   line-height: 1.5;
 }
+
+.admin-cards-skeleton {
```
edx-abolger, line 19: It's fine that these use pixels and not classes?

src/components/TableComponent/index.jsx
adamstankiewicz: The purpose of this component is to display a small loading message overlaying an existing table. The `LoadingMessage` component is rendered first when no data exists at all for the table. When page 1 of a table is rendered, and then the user paginates, we had the `TableLoadingOverlay` component overlay a small loading message on top of the table to give the affordance that the table is reloading, which helped avoid constantly swapping in and out the `LoadingMessage` - or a skeleton, now - when the user paginates which could get annoying for the user:

image

So, this line _is_ reachable when the table is `loading` and we already have existing data from a previous render of the table.

src/components/Admin/Admin.scss
edx-abolger: Edit: nvm it's a min-height

src/components/TableComponent/index.jsx
edx-abolger: What's there is clunky. Frankly I'd remove the `&& !data` part of the loading condition at line 133 because to me it looks like it is masking some other error: unless your loading state is just wrong, why would you care about whether data was non-null? That would only happen in a weird race condition.

edx-abolger (approved):

edx-abolger (changes requested):
Based on Adam's review, revert line 73 needed.

src/components/TableLoadingOverlay/index.jsx
```diff
@@ -4,7 +4,7 @@ import LoadingMessage from '../LoadingMessage';
 const TableLoadingOverlay = () => (
   <>
     <div className="table-loading-overlay" />
-    <div className="table-loading-message d-flex align-items-center justify-content-center ">
+    <div className="table-loading-message d-lex align-items-center justify-content-center ">
```
edx-abolger, line 7: @raq929 is this a bug? or on purpose?

src/components/Admin/AdminCardsSkeleton.jsx
raq929: Oooooh, yes I like that plan.

src/components/TableLoadingOverlay/index.jsx
raq929: that is a buggy bug good catch.

src/components/Admin/AdminCardsSkeleton.jsx
raq929: In the interest of time, I decided against this approach, but I would support us doing it in the future.

edx-abolger (approved):

PR #536, opened 2021-04-28
feat: bulk enrollment stepper

raq929:
This adds a bulk enrollment stepper that allows users to choose the emails of users with subscription licenses and review their selections. 

**THINGS IT CURRENTLY DOES NOT DO**
1. Get the correct learner emails. (Maybe it does but I am *definitely* not sure). We need to either modify the licence-manager endpoint or otherwise figure out how to get only active and pending license holders.  
2. Actually enroll the learners 

I will make/modify tickets to for adding that functionality, but I wanted to have some review on this first because it's a *lot*.

**WHAT IT DOES DO**
- Add selected rows to a BulkEnrollmentContext
- Instruct the DataTables in question to display selected rows based on that context, such that rows can be deleted on the Review page
- Review step: Show/hide emails & courses and emails over 25
- Stepper sends users to the correct page

I'm working on adding tests now. Please ignore console.logs and such, I will clean those up as I test the files.

Step 1: 
You get to it by clicking "Enroll Learners" from the main bulk enrollment page
Screen Shot 2021-04-28 at 10 12 41 AM

Step 2: 
Screen Shot 2021-04-28 at 10 13 39 AM

Show/Hide
Screen Shot 2021-04-28 at 10 04 08 AM


src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -39,16 +44,17 @@ export const EnrollButton = ({ row, setSelectedCourseRuns, setModalOpen }) => {
   );
 };
 
+const ExportDataTableContext = () => {
+  const { courses: [, coursesDispatch] } = useContext(BulkEnrollContext);
```
taliaedX, line 48: My brain is breaking from this definition.
I know the context has `{courses: [courseTableState, coursesDispatch]...}` and based on how we're using `coursesDispatch` later, I am guessing this is saying "get coursesDispatch which is the second item in this list of the courses object in this context"? I just didn't know you could do object retrieval like that (pretty cool if that is the case. Otherwise, I clearly have no idea what's going on here)

taliaedX (commented):
First pass looks pretty good. Gonna want to do a second pass once it seems ready to go.

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
raq929: It is doing exactly that! It's destructuring syntax. It is, for me, **the most hated** part of destructuring syntax, but the linter won't let me have a variable that I don't use, so.
It won't let me link to directly to it, but it's under "Ignoring some return values" 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

src/components/BulkEnrollmentPage/BulkEnrollment.scss
```diff
@@ -10,3 +10,14 @@
     padding-left: 0;
   }
 }
+
```
edx-abolger, line 13: For my own learning: I've previously followed conventions where things like .list-item would be placed as close to the relevant component as possible instead of on the parent.  Does doing it this way just help to make the component itself more reusable, as the styling would change when the component is used elsewhere?

raq929: Oh, yeah, it should be moved. I threw the styles here out of laziness and planned to clean them up, but then completely forgot about it.

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
taliaedX: That's a handy link! The other confusing bit was also the use of the courses key, since normally that'd be a definition (ie `const something = { x: y }`), but here we are just specifying which object to retrieve certain data out of.

edx-abolger:
QA results for things that worked properly:
- can't click enroll without selected data
- delete buttons working
- single and multiple email selections working
- enroll learners link and button both working
- multi subscription enterprise working

Bugs: 
- "Return to Course selection" button (after deleting course) closes modal but holds on to selected learners state and modal state: When you then click 'Enroll Learners' again you get taken straight to the final page... without your newly selected course. You get stuck in a loop where you can't select a course.
- selecting email, clicking 'next' to review screen and then clicking 'previous' back to email list clears your learner selections

Potential future tickets:
- Once you select a subscription plan the only way to get back to that screen is re-clicking Catalog Management in sidebar. Is that fine?
- Really small screens do weird stuff with 2 column layout (see screenshot).  Should we stack the 2 columns?

edx-abolger:
Screen Shot 2021-04-30 at 7 52 22 AM

binodpant:
@raq929 looked at the first notes in desc:

> Get the correct learner emails. (Maybe it does but I am definitely not sure). We need to either modify the licence-manager endpoint or otherwise figure out how to get only active and pending license holders.

I thought that was done as part of https://openedx.atlassian.net/browse/ENT-4417?

Also, a request to include me in all Bulk enrollment PRs! I may not end up reviewing all of them just time wise but hopefully I can review most of them :)

src/components/BulkEnrollmentPage/data/reducer.js
```diff
@@ -0,0 +1,14 @@
+import { SET_SELECTED_ROWS, DELETE_ROW } from './actions';
+
+const selectedRowsReducer = (state = [], action) => {
```
binodpant, line 3: STUDY/LOOK INTO: obviously let's not worry about it for this PR but if we use action and reducers code, we may want to look into using reduxjs/toolkit to reduce some of our actions/reducers boilerplate. e.g., `createSlice` takes care of reducing an amazing amount of code (esp more apparent where there are lots of actions)
https://redux-toolkit.js.org/api/createSlice while keeping all actions and reducers usage intact.

src/components/BulkEnrollmentPage/hooks.jsx
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable import/prefer-default-export */
+import { DataTableContext } from '@edx/paragon';
+import { useContext, useEffect } from 'react';
+import { setSelectedRowsAction } from './data/actions';
+
+export const setExportedTableInstance = ({ dispatch }) => {
```
binodpant, line 6: is this function better named something like, `setSelectedRowsFromTable` or however else it makes sense to call it based on what it's doing?

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -164,13 +167,13 @@ export const BaseCourseSearchResults = ({
...
                 row={row}
-                setSelectedCourseRuns={setSelectedCourseRuns}
-                setModalOpen={setModalOpen}
+                setStepperOpen={setStepperOpen}
               />
             ),
           },
         ]}
       >
+        <ExportDataTableContext />
```
binodpant, line 176: this is an interesting approach to render a null returning component and have it grab table selected rows into the context here, if I understood what it's doing. I presume that is because currently we can't have a callback given to data table? something like one that runs when rows are selected? perhaps that would be a nice addition?

raq929: Oh, we could have, and that would have been better, and I abandoned this approach anyway.

raq929:
@taliaedX 
Once you select a subscription plan the only way to get back to that screen is re-clicking Catalog Management in sidebar. Is that fine?
- I've brought that up with Jenn but it hasn't made it into a ticket yet
Really small screens do weird stuff with 2 column layout (see screenshot). Should we stack the 2 columns?
YES, YES WE SHOULD I will open a ticket.

raq929:
Closing this PR because it is included in the `feat/be-stepper` branch

PR #537, opened 2021-04-28
style: rename Bulk Enrollment titles and endpoints

taliaedX:
- Renames the bulk enrollment endpoint path from `catalog-management` to `enrollment`
- Renames the titles in a few spots (ie wherever it was present) from "Catalog Management" to "Subscription Enrollment"


binodpant (approved):
yeah I think this is it... I think tests if any that relied on the route name must also use the constant already. so looks like don't need test change thanks to Lael's last PR!

PR #538, opened 2021-04-28
chore: clean up config values.

taliaedX:
- Removes webpack config files as we're using slightly different vars that will be coming from config files in edx-internal (that pr MUST BE MERGED FIRST).
- Also puts the vars in the .env.development mostly for convenience.
- There are also 2 removals, because I am 99.99% sure they weren't doing anything:
  - `FEATURE_FLAGS={}` in `.env.test` I cannot find any indication (including local testing) that dotenv supports nesting variables in an object/dict like this. 
  - all the `hasFeatureFlagEnabled` stuff. If I am not wrong this is just to see it in a browser without having to modify webpack right? Which without webpack is pointless? Am I wrong here?
- Adds a test for the index of bulk enrollment 100% because Coveralls was failing for unknown reasons.

related edx-internal PR: https://github.com/edx/edx-internal/pull/5008


binodpant:
on this one, 

>  If I am not wrong this is just to see it in a browser without having to modify webpack right? Which without webpack is pointless? Am I wrong here?

I figured we added this before the edx-internal setup was available so that we could easily switch features from the browser without updating git (in demos and such)? If we turn a feature on permanently (meaning in edx-internal) we should not need this ability for that flag anymore

but don't we still need that setup in case we want to do that for a future flag?

.env.test
```diff
@@ -2,7 +2,6 @@ LMS_BASE_URL='http://localhost:18000'
 DATA_API_BASE_URL='http://localhost:8000'
 ECOMMERCE_BASE_URL='http://localhost:8000'
 LICENSE_MANAGER_BASE_URL='http://localhost:18170'
-FEATURE_FLAGS={}
```
binodpant, line 5: yeah tests should just mock a feature toggle, if they need it anyway so I think this should be fine. some of it can also go to jest.config.js or setupTests.js if we really need it global

binodpant:
> on this one,
> 
> > If I am not wrong this is just to see it in a browser without having to modify webpack right? Which without webpack is pointless? Am I wrong here?
> 
> I figured we added this before the edx-internal setup was available so that we could easily switch features from the browser without updating git (in demos and such)? If we turn a feature on permanently (meaning in edx-internal) we should not need this ability for that flag anymore
> 
> but don't we still need that setup in case we want to do that for a future flag?

also I think it will be useful if this ability to check process.env but be able to override using browser, can be made available from frontend-platform / frontend-build instead so we don't have to have this in each repo ... just a thought, not for this PR

FWIW I see learner portal also uses a similar function for just one feature right now 
https://github.com/edx/frontend-app-learner-portal-enterprise/blob/f199ec7c0a4550f584c3ea0fa9e24b1a1405a576/src/config/index.js#L4

src/config/index.js
```diff
@@ -1,10 +1,3 @@
-import qs from 'query-string';
-
-const hasFeatureFlagEnabled = (featureFlag) => {
```
alex-sheehan-edx, line 3: I'm mixed on if it's worth keeping `hasFeatureFlagEnabled` around or not. In my head the pro's and con's are
PROS:
- Doesn't impede/break anything to keep it
- if an engineer is testing something, say on stage, this offers an easy way to toggle the features on the page without having to create a PR and add a change to edx-internal
CONS:
- Allows devs to be slightly lazy in that they're not forced to enable features on stage
- Current implementation means that the user has to add the query parameters _every_. _single_. _time_. they open/load the page

Overall I think I lean towards removing but want to put the question out there if there's any merit in keeping it around.

alex-sheehan-edx (approved):
couple of questions/discussions but overall this LGTM 👍

src/config/index.js
binodpant: yeah I think it's useful while you are still not released the feature fully and turned on in edx-internal by default. once it's turned on the feature should not need to be toggled via browser. I think that is useful while we are still off, and maube want to test by turning it on in the browser...

taliaedX: yeah I believe @binodpant also brought up a similarish concern: `but don't we still need that setup in case we want to do that for a future flag?`

And yeah the flag function is still definitely a way to not have to make a separate PR in order to test a change in this repo.
However, to your con about devs not enabling the feature in stage: I see the default enabling of features in stage as an additional layer to catch bugs. You might test the specific feature when enabling the flag, but as others are pushing changes to stage, we can inherently test the feature's interaction with other areas of the app and other users' changes by making it enabled for *everyone*.

The real pro I see to having it there is that if your feature completely breaks something, it is turned off automatically for other devs. That's a pretty big pro, but at the same time, why/how often are we merging and thus shipping completely broken code that doesn't even do what was intended and crashes the service such that everyone is affected? I feel like that is something to alert everyone on because uhhhhhh I don't want that code in prod.

Anyways, I know it doesn't come across here, but I am also not opposed to keeping the feature around. I just also don't see a _necessity_ in it and thus just went about deleting!all!the!things!

.env.test
taliaedX: What works as intended? I am like 99% sure this Feature flags var _doesn't_ work as intended, which is why I just chucked it out.

taliaedX: Also the test still pass with this chucked, so 🤷

src/config/index.js
binodpant: yeah I would say let's leave it in? if we end up not using it at all yeah we can scrap it.. I just see it being useful sometimes... (also it's quite easy to add if/when needed so not worried either way).

binodpant (approved):
lgtm, whatever we decide about that function, is okay with me, for this PR but suggsted leaving it in.

PR #539, opened 2021-05-03
ENT4446- Removed extra params from assignment emails call

zamanafzal:


sameenfatima78 (approved):

src/components/CodeAssignmentModal/index.jsx
```diff
@@ -299,16 +296,10 @@ export class BaseCodeAssignmentModal extends React.Component {
         enterprise_customer: enterpriseUuid,
       };
     }
-
+    const assignmentEmails = isBulkAssign ? validEmails : [formData['email-address']];
```
HammadAhmadWaqas, line 299: can we reuse the if statement we already have?
```
const assignmentEmails;
if (isBulkAssign) {
      assignmentEmails = validEmails;

      // Only includes `codes` in `options` if not all codes are selected.
      if (!hasAllCodesSelected) {
        options.codes = selectedCodes.map(selectedCode => selectedCode.code);
      }
    } else {
      assignmentEmails = [formData['email-address']];
      options.codes = [code.code];
    }

```

HammadAhmadWaqas (approved):
just small suggestion, Otherwise good.

src/components/CodeAssignmentModal/index.jsx
zamanafzal: In that case, code will be moved in else block  if `isBulkAssign`  `hasAllCodesSelected` is true which is not desired

`if (isBulkAssign && !hasAllCodesSelected) {`

PR #544, opened 2021-05-10
maint: adding success and failure handling to the new bulk enrollment

alex-sheehan-edx:
[Jira ticket](https://openedx.atlassian.net/browse/ENT-4423)

Changes:
- new bulk enrollment submit component added to the stepper
   - the stepper now calls the license manager API on final submit  
   - success and error responses from the Licence Manager are handled according to ticket AC 
      - success cases take the user back to the first step and displays a success Toast
      - error cases open up a modal alert with a link to support
         - closing the modal should keep the user in the current final step
   

**Bulk Enrollment Submit component**   
- Desktop
image

- Mobile
image

**Success Toast**
- Desktop
image

- Mobile
image

**Error Alert Modal**
- Desktop
image

- Mobile
image

### DIFF
--- a/src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
@@ -1,0 +1,129 @@
+import React, {
+  useContext,
+  useState,
+} from 'react';
+import {
+  Button, Form, AlertModal, Hyperlink, ActionRow, useToggle,
+} from '@edx/paragon';
+import PropTypes from 'prop-types';
+import { logError } from '@edx/frontend-platform/logging';
+import {
+  FINAL_BUTTON_TEST_ID,
+  FINAL_BUTTON_TEXT,
+  NOTIFY_CHECKBOX_TEST_ID,
+  CUSTOMER_SUPPORT_HYPERLINK_TEST_ID,
+  ALERT_MODAL_TITLE_TEXT,
+  ALERT_MODAL_BODY_TEXT,
+  SUPPORT_HYPERLINK_TEXT,
+} from './constants';
+import LicenseManagerApiService from '../../../data/services/LicenseManagerAPIService';
+import { BulkEnrollContext } from '../BulkEnrollmentContext';
+import { ToastsContext } from '../../Toasts';
+import { clearSelectionAction } from '../data/actions';
+
+export const BulkEnrollmentAlertModal = ({ isOpen, toggleClose, enterpriseSlug }) => (
+  <AlertModal
+    title={ALERT_MODAL_TITLE_TEXT}
+    isOpen={isOpen}
+    onClose={toggleClose}
+    footerNode={(
+      <ActionRow>
+        <Button variant="primary" onClick={toggleClose}>OK</Button>
+      </ActionRow>
+    )}
+  >
+    <p>
+      {ALERT_MODAL_BODY_TEXT}
+      <Hyperlink
+        destination={`/${enterpriseSlug}/admin/support`}
+        target="_blank"
+        rel="noopener noreferrer"
+        data-testid={CUSTOMER_SUPPORT_HYPERLINK_TEST_ID}
+      >
+        {SUPPORT_HYPERLINK_TEXT}
+      </Hyperlink>
+    </p>
+  </AlertModal>
+);
+
+BulkEnrollmentAlertModal.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  toggleClose: PropTypes.func.isRequired,
+  enterpriseSlug: PropTypes.string.isRequired,
+};
+
+export const generateSuccessMessage = numEmails => `${numEmails} learners have been enrolled.`;
+
+const BulkEnrollmentSubmit = ({ enterpriseId, enterpriseSlug, returnToInitialStep }) => {
+  const [loading, setLoading] = useState(false);
+  const [checked, setChecked] = useState(true);
+  const handleChange = e => setChecked(e.target.checked);
+
+  const {
+    emails: [selectedEmails, emailsDispatch],
+    courses: [selectedCourses, coursesDispatch],
+  } = useContext(BulkEnrollContext);
+  const { addToast } = useContext(ToastsContext);
+
+  const courseKeys = selectedCourses.map(
+    ({ original, id }) => original?.advertised_course_run?.key || id,
+  );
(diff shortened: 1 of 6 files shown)


src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -0,0 +1,132 @@
...
+  const [checked, setChecked] = useState(true);
+  const handleChange = e => setChecked(e.target.checked);
+
+  const {
+    emails: [selectedEmails, emailsDispatch],
+    courses: [selectedCourses, coursesDispatch],
+  } = useContext(BulkEnrollContext);
+  const { addToast } = useContext(ToastsContext);
+
+  let courses;
```
raq929, line 60: Nit: I would call this `courseRuns` or `courseKeys` or whatever seems most accurate to what the information stored in the array is.

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -0,0 +1,132 @@
...
+  if (selectedCourses.original?.advertised_course_run) {
+    courses = selectedCourses?.map(
+      ({ selectedRow }) => selectedRow.original?.advertised_course_run?.key,
```
raq929, line 63: I think instead of an `if/else` here, you can just do 
```
({ selectedRow }) => selectedRow.original?.advertised_course_run?.key || id
```
Because if any of those `?` keys doesn't exist, it will be `undefined`

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -0,0 +1,132 @@
...
+    );
+  } else {
+    courses = selectedCourses.map(({ id }) => id);
+  }
+  const emails = selectedEmails.map(({ values }) => values.userEmail);
+  const [isOpen, toggleErrorModalOpen, toggleErrorModalClose] = useToggle();
```
raq929, line 69: Nit: I would probably go with `isErrorModalOpen` for consistency.

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -0,0 +1,132 @@
...
+  };
+
+  return (
+    <>
+      <BulkEnrollmentAlertModal
+        enterpriseSlug={enterpriseSlug}
+        toggleClose={toggleErrorModalClose}
+        isOpen={isOpen}
+      />
+      <Form.Checkbox
```
raq929, line 102: We should discuss, probably with @wittjeff if this should be wrapped in a `<Form>` component. I suspect yes.

raq929 (commented):
Yay for testing! This is looking great; I have some nitty nit picks. 😄

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.test.jsx
```diff
@@ -0,0 +1,343 @@
...
+    await userEvent.click(button);
+
+    expect(addToast).toBeCalledTimes(1);
+    expect(logError).toBeCalledTimes(0);
+    expect(addToast).toHaveBeenCalledWith(
+      `${userEmails.length} learners have been enrolled.`,
+    );
+  });
+
+  // TODO: figure out why the modal isn't shown :(
```
binodpant, line 274: is this a new ticket or just wip in this PR?

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -0,0 +1,132 @@
...
+      notify: checked,
+    };
+
+    return LicenseManagerApiService.licenseBulkEnroll(
+      enterpriseId,
+      options,
+    ).then(() => {
+      coursesDispatch(clearSelectionAction());
+      emailsDispatch(clearSelectionAction());
+      addToast(`${selectedEmails.length} learners have been enrolled.`);
```
binodpant, line 86: nit: one of the tests, and this code could perhaps use the same arrow function or sth like this? Just makes it easier to change in one place if we get any UX updates later and test won't have to update either

```
numEmails => `${numEmails} learners have been enrolled.`
```

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.test.jsx
alex-sheehan-edx: Lael and I hit our collective heads against the proverbial wall and couldn't come up with a sufficient solution. We decided to leave it for a new ticket, unless you can come up with something!

binodpant: yeah cool as long as we ticket it up, thanks! if this works manually then yeah a test only ticket would be great

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
alex-sheehan-edx: Talked with Jeff, I think we're ok for now! 👍

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.test.jsx
alex-sheehan-edx: everything works manually! The problem is that it's difficult to get the test/screen to recognize the new tab.

binodpant (approved):
checked the latest changes looking good once passed! will dig in further once merged when testing it out

src/components/BulkEnrollmentPage/stepper/constants.jsx
```diff
@@ -7,7 +7,13 @@ export const FINAL_BUTTON_TEXT = 'Enroll learners';
 export const NEXT_BUTTON_TEST_ID = 'stepper-next';
 export const PREV_BUTTON_TEST_ID = 'stepper-prev';
 export const FINAL_BUTTON_TEST_ID = 'stepper-final';
+export const NOTIFY_CHECKBOX_TEST_ID = 'notify-checkbox';
```
edx-abolger, line 10: Kudos: LOVE the test id convention we're using here, please call out and show to others.

edx-abolger (approved):

PR #545, opened 2021-05-12
Update Footer and Sidebar Support links and remove support form

taliaedX:
https://github.com/edx/edx-internal/pull/5087 needs to merge first.

ECS would like to move to a more self-service model of support for our enterprise customers and as such would like them to visit the help center before immediately contacting support. As such, we have updated the links such that they direct customers to the help center instead of to a direct contact form.

This PR does the following:
- Removes the header link altogether (agreed to be extremely redundant)
- Updates the sidebar support link to direct to an external link (ie the help center) instead of to a form.
  - Thus, the component and all references to the SupportPage component have been removed.
- Updates the footer link. I also added a test since we don't have one.
- I also fixed the format of the dev `.env` because commas aren't a thing you put in them. 

#### Before:
Screen Shot 2021-05-13 at 1 47 46 PM
Screen Shot 2021-05-13 at 1 47 40 PM

#### After:
Screen Shot 2021-05-13 at 1 51 35 PM
Screen Shot 2021-05-13 at 1 52 52 PM


taliaedX:
>Coverage decreased (-0.06%) to 70.946%

I deleted code and added a test!
(╯°□°）╯︵ ┻━┻

src/components/Sidebar/index.jsx
```diff
@@ -105,9 +105,10 @@ class Sidebar extends React.Component {
       // NOTE: keep "Support" link the last nav item
       {
         title: 'Support',
-        to: `${baseUrl}/admin/support`,
+        to: configuration.ENTERPRISE_SUPPORT_URL,
         icon: faLifeRing,
         hidden: !features.SUPPORT,
+        external: true,
```
binodpant, line 111: hmm in the test it says external is default false. by default are we going the 'external' route

binodpant (approved):
made one minor question about default of external (could just be a comment thing)

also would be nice to have a screenshot attached

but lgtm otherwise!

src/components/Sidebar/index.jsx
taliaedX: I'm confused here.

Where external is default:
```
IconLink.defaultProps = {
  icon: null,
  isExpanded: false,
  external: false,
```

I'm explicitly only passing it in for the Support item right above your comment. It's otherwise not being appended to any of the sidebar links.

So yes, the default is false as we're not passing it in so it uses the defaultProp value.

binodpant: sorry I think I confused this, default us false. we pass true so looks good!

PR #546, opened 2021-05-13
fix: get correct learners for bulk enrollment

raq929:
Add query param to only get active and pending users


src/components/subscriptions/data/hooks.js
```diff
@@ -117,7 +120,7 @@ export const useAllSubscriptionUsers = ({
     setLoading(true);
     LicenseManagerApiService.fetchSubscriptionUsers(
       subscriptionUUID,
-      { status: licenseStatusByTab[TAB_ALL_USERS], page_size: 500 },
+      { active_only: 1, page_size: 10000 },
```
taliaedX, line 123: I was losing my mind because I couldn't find any of the bit you're removing on master. You're likely going to hit a hell of a lot of merge conflicts here when the base branch diffs against master.

I'm also unclear if we want to get rid of this tabbing experience. Unclear on that part exactly.

taliaedX (commented):
I'm not entirely familiar with the `licenseStatusByTab` usage to grasp if we're affecting something we don't want to so may want Adam to review? But otherwise, the changes seem fine, I just think you're going to hit a bunch of merge conflicts as none of this exists on master. :/

src/components/subscriptions/data/hooks.js
raq929: I mean, we're going to merge the whole feature; no one working on bulk enrollment is commiting to master right now.

So, the reason this is a different hook from the one below it, is because the original was designed for use by the license management page, which has tabs by user. While I was developing the stepper, I just grabbed all the users. But I don't actually want that, I want just active users, and I want to put them all in the same table, not have them separated into tabs. Hence the separate hook.

taliaedX: ah ah my bad. I was thinking the `useAllSubscriptionUsers` method existed at some point and your other branch was behind. I didn't realize it was added to that branch and that is what you are now changing.

taliaedX: With that, and thus looking at this more clearly, the only issue I see here is similar to the comment Alex D had on your backend pr: is this page size an appropriate jump? Do we even currently have any cases where we have 500 records even to determine if there is a need for a larger result set?

raq929: Switching to backend pagination, it turns out we do have customers with huge numbers of learners.

src/components/BulkEnrollmentPage/data/reducer.js
```diff
@@ -1,11 +1,13 @@
+import { uniqBy } from 'lodash';
 import {
   SET_SELECTED_ROWS, DELETE_ROW, ADD_ROW, CLEAR_SELECTION,
 } from './actions';
 
 const selectedRowsReducer = (state = [], action) => {
   switch (action.type) {
     case SET_SELECTED_ROWS:
-      return action.rows;
+      return uniqBy([...state, ...action.rows], (row) => row.id);
```
taliaedX, line 9: TIL

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
```diff
@@ -43,14 +41,68 @@ const selectColumn = {
...
 const AddLearnersStep = ({
   subscriptionUUID,
   enterpriseSlug,
 }) => {
-  const [errors, setErrors] = useState([]);
+  const [errors, setErrors] = useState('');
+  const [loading, setLoading] = useState(true);
+  const [data, setData] = useState({ results: [], count: 0, numPages: 1 });
   const { emails: [selectedEmails] } = useContext(BulkEnrollContext);
+  const { results, count, numPages } = data;
```
taliaedX, line 70: Surprised there isn't a way to just combo set this at the time you set data. Something like `const [data{results, count, numPages}, setData] = useState...`

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
```diff
@@ -60,27 +112,30 @@ const AddLearnersStep = ({
...
-            pageIndex: 0,
-            selectedRowIds: convertToSelectedRowsObject(selectedEmails),
-          }}
-          initialTableOptions={{
-            getRowId: (row, relativeIndex, parent) => row?.uuid || (parent ? [parent.id, relativeIndex].join('.') : relativeIndex),
-          }}
-        />
-      )}
+      {errors && <Alert variant="danger">There was an error retrieving email data. Please try again later.</Alert>}
+      <DataTable
```
taliaedX, line 116: Is there any visual change here? We added the table child components to appear only on load vs the original, parent table component only being shown on load.

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
```diff
@@ -33,31 +40,40 @@ const StepperWrapper = (props) => (
   </BulkEnrollContextProvider>
 );
 
+const flushPromises = () => new Promise(setImmediate);
```
taliaedX, line 43: also TIL `setImmediate`

src/components/subscriptions/data/hooks.js
```diff
@@ -120,7 +120,7 @@ export const useActiveSubscriptionUsers = ({
     setLoading(true);
     LicenseManagerApiService.fetchSubscriptionUsers(
       subscriptionUUID,
-      { active_only: 1, page_size: 10000 },
+      { active_only: 1, page_size: 3 },
```
taliaedX, line 123: you mentioned doing pagination on the backend, so I'm just wondering why are we passing in a page size here? So we still determine what the size should be on the frontend, pass it to the backend, and thus have the backend pre-limit its results to the amount we specify on the frontend, but is there a reason to specify it on the frontend?

Also, there's several other spots I see where we pass page_size to this method, but with the `PAGE_SIZE` constant which is a different amount (5) and I'm wondering if 1. if these are supposed to be the same number and 2. if so, we should define PAGE_SIZE in an external constants file so that this hook and the `AddLearnersStep.jsx` file where it's defined can both use the same constant/value.

taliaedX (commented):
I have one concern about page_size usage(s), but other than that my comments are more for my understanding and with that clarification, can approve.

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
raq929: Yeah, because of the change in how we call the API, the parent table component has to be present for the API to be called. So I moved some things around so it's not always displaying the table.

src/components/subscriptions/data/hooks.js
raq929: Oh, hahaha yeah, I'm actually not using this hook at all anymore; I can remove it.

raq929: And the PAGE_SIZE constant should be 25. I don't actually have 25 users so I set it lower for testing and forgot to set it back. Thanks!

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
raq929: Oh, yeah, it can definitely be done this way. I probably split it up like this for debugging purposes.

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
raq929: But no, there's no visual change.

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
raq929: Yeah, I *still* don't fully understand *why* this trick works, but it sure does work, so I've decided I don't care.

src/components/BulkEnrollmentPage/data/reducer.js
raq929: Lodash is my hero

src/components/subscriptions/data/hooks.js
taliaedX: Ah. Okay so that explains the hook and I guess the variable only needs to be in that one file, but still why are we defining page size on the frontend? Do we expect to not always use a static size (I haven't looked at the backend PR in a bit though so apologies if this is redundant)?

taliaedX (approved):
discussed page size and thus LGTM!

PR #551, opened 2021-05-17
fix: update sidebar toggle to use Paragon icons

raq929:
I was investigating a Paragon PropTypes bug, and in my debugging efforts, updated this button to use the new icons.


adamstankiewicz (approved):

raq929:
I fixed these tests for similar reasons on the `feat/be-stepper` branch, gonna rebase after that merge.

PR #552, opened 2021-05-18
fix: swap support form link to mailTo

taliaedX:
### Background
Due to completion of ENT-4451, the support form no longer exists, thus we needed to update this.

We decided on a mailTo link for now (eventually support's help center will have a new form and we may want to link directly to that, but that is completely out of scope for anything Bulk Enrollment related)

*NOTE:* I am waiting on confirmation of the language used in the support email to ensure it is useful to the support team.

### What this changes
- Swaps HyperLink to MailtoLink component in BulkEnrollmentAlertModal.
- As part of this we now pass in the actual error message so that support gets context in the email automatically.
- Fixes an issue with the `ALERT_MODAL_BODY_TEXT` constant as the multiline format was not actually compatible with web browsers so it was displaying as just `NaN`

### Screenshot
Screen Shot 2021-05-18 at 12 06 14 PM

### DIFF
--- a/src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
@@ -3,7 +3,7 @@ import React, {
   useState,
 } from 'react';
 import {
-  Button, Form, AlertModal, Hyperlink, ActionRow, useToggle,
+  Button, Form, AlertModal, ActionRow, useToggle, MailtoLink,
 } from '@edx/paragon';
 import PropTypes from 'prop-types';
 import { logError } from '@edx/frontend-platform/logging';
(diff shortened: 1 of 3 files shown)


src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -34,14 +39,16 @@ export const BulkEnrollmentAlertModal = ({ isOpen, toggleClose, enterpriseSlug }
...
       {ALERT_MODAL_BODY_TEXT}
-      <Hyperlink
-        destination={`/${enterpriseSlug}/admin/support`}
+      <MailtoLink
+        to={configuration.CUSTOMER_SUPPORT_EMAIL}
         target="_blank"
         rel="noopener noreferrer"
         data-testid={CUSTOMER_SUPPORT_HYPERLINK_TEST_ID}
+        subject={SUPPORT_EMAIL_SUBJECT + enterpriseSlug}
+        body={`${SUPPORT_EMAIL_BODY} ${error}`}
```
taliaedX, line 48: why is my formatting inconsistent? Idk. 🤷

alex-sheehan-edx (approved):
LGTM 👍  Thanks for updating the test

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
```diff
@@ -50,13 +57,15 @@ BulkEnrollmentAlertModal.propTypes = {
...
   enterpriseSlug: PropTypes.string.isRequired,
+  error: PropTypes.string.isRequired,
 };
 
 export const generateSuccessMessage = numEmails => `${numEmails} learners have been enrolled.`;
 
 const BulkEnrollmentSubmit = ({ enterpriseId, enterpriseSlug, returnToInitialStep }) => {
   const [loading, setLoading] = useState(false);
   const [checked, setChecked] = useState(true);
+  const [error, setError] = useState(true);
```
binodpant, line 68: should this be `useState('')` instead for initial state of 'error'? not sure if it matters, just noting that 'error' is a propTypes.string.required so worth a check ...

binodpant (approved):
minus small comment, I am good with this! looks like when @raq929 's feat/be-stepper is merged, it will adopt this change in the alert modal with minor merging, so we will be golden!

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.jsx
taliaedX: Yep that makes sense. Or a null value. This is some copypasta error.

PR #553, opened 2021-05-19
feat: allowing users to disable coupon code assignment emails

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4476)

**Notify learners checked:**
**Desktop**
image

**Mobile**
image

**Notify learners unchecked**
**Desktop**
image

**Mobile**
image

This PR:
- Changes wording of modal from `Users` to `Learners`
- Adds a checkbox that toggles the visibility of the email template part of the component
- Checkbox also controls a new boolean `option` parameter fed into the API call to ecommerce on the modal submit
- Unit test to check that the checkbox is rendering

NOTE: discussed with the team and due to the current setup, the modal's submitHandler is untestable so it's currently impossible to test the checkbox's functionality of altering the options used within the handler function. There is currently a jira ticket that is going to deal with this issue and the need for this test will be added to it.

### DIFF
--- a/src/components/CodeAssignmentModal/BulkAssignFields.jsx
+++ b/src/components/CodeAssignmentModal/BulkAssignFields.jsx
@@ -7,7 +7,7 @@ import { normalizeFileUpload } from '../../utils';
 
 const BulkAssignFields = () => (
   <>
-    <h3 className="mb-2">Add Users</h3>
+    <h3 className="mb-2">Add learners</h3>
     <div className="pl-4 field-group">
       <Field
         name="email-addresses"
--- a/src/components/CodeAssignmentModal/IndividualAssignFields.jsx
+++ b/src/components/CodeAssignmentModal/IndividualAssignFields.jsx
@@ -5,7 +5,7 @@ import RenderField from '../RenderField';
 
 const IndividualAssignFields = () => (
   <>
-    <h3>Add User</h3>
+    <h3>Add learner</h3>
     <Field
       name="email-address"
       type="email"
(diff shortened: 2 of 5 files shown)


src/components/CodeAssignmentModal/BulkAssignFields.jsx
```diff
@@ -7,7 +7,7 @@ import { normalizeFileUpload } from '../../utils';
 
 const BulkAssignFields = () => (
   <>
-    <h3 className="mb-2">Add Users</h3>
+    <h3 className="mb-2">Add Learners</h3>
```
raq929, line 10: Sentence case! (I know it wasn't originally, but that's the standard now, so we should update when we touch things.

src/components/CodeAssignmentModal/IndividualAssignFields.jsx
```diff
@@ -5,7 +5,7 @@ import RenderField from '../RenderField';
 
 const IndividualAssignFields = () => (
   <>
-    <h3>Add User</h3>
+    <h3>Add Learner</h3>
```
raq929, line 8: Sentence case. 😄

src/components/CodeAssignmentModal/BulkAssignFields.jsx
edx-abolger, line 10: +1 I hate the word User it's so vague!

raq929 (approved):
Small nits, LGTM

PR #554, opened 2021-05-19
fix: change rows to column view for mobile on Review step of stepper

taliaedX:
### Background
Using 2+ columns on mobile is not best for the mobile user experience.
This PR swaps to a single column and displays the boxes/cards on top of each other.

#### Before:
Screen Shot 2021-05-19 at 4 43 18 PM

#### After:
Screen Shot 2021-05-19 at 4 34 57 PM
Screen Shot 2021-05-19 at 4 35 25 PM

### DIFF
--- a/src/components/BulkEnrollmentPage/BulkEnrollment.scss
+++ b/src/components/BulkEnrollmentPage/BulkEnrollment.scss
@@ -28,6 +28,16 @@
   }
 }
 
+@include media-breakpoint-down(md) {
+  .be-review-list {
+    display: flex;
+  }
+  .row {
+    display: flex;
+    flex-direction: column;
+  }
+}
+
 .sticky-stepper-footer {
   display: flex;
   flex-direction: column;
--- a/src/components/BulkEnrollmentPage/stepper/ReviewStep.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/ReviewStep.jsx
@@ -34,20 +34,20 @@ const ReviewStep = ({ setCurrentStep }) => {
       <h2 className="mb-5">{REVIEW_TITLE}</h2>
       <Row>
         <ReviewList
-          key="emails"
-          rows={selectedEmails}
-          accessor="userEmail"
-          dispatch={emailsDispatch}
-          subject={EMAILS}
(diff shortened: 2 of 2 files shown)


edx-abolger (approved):

src/components/BulkEnrollmentPage/BulkEnrollment.scss
```diff
@@ -28,6 +28,16 @@
   }
 }
 
+@media (max-width: 767px) {
```
binodpant, line 31: unclear if this what we want but I see a lot of usage of this max-container-width defined in paragon.. in case it feels relevant..

https://github.com/search?p=3&q=org%3Aedx+container-max-widths&type=Code

it seems to be defined here:

https://github.com/edx/paragon/blob/c70b44b9ea45ce3a201f1f83918f28e9e6527273/scss/core/_variables.scss#L452

binodpant: also some width defs here: https://github.com/edx/paragon/blob/c70b44b9ea45ce3a201f1f83918f28e9e6527273/scss/core/_variables.scss#L470

taliaedX: Yeah I had seen the container variable, but assumed it was typically used for a smaller portion of the interface especially since the sizes are vastly different from what we've designated as "small/xs" and also what Bootstrap has noted as small/xs.

I can try the md width as the closest to the usual width, but I think it then becomes less obvious we're isolating for mobile.

binodpant: yeah I think this is a good #fedx ask/discussion if you are up for that, but in the context of this change I have no flag is the bootstrap one does its job!

taliaedX: Yeah I may honestly look at asking there. Don't want to end up with a huge discussion, but right now it doesn't look like we import the bootstrap variables directly anywhere so they aren't recognized so I'd have to import it probably in the same way ecommerce does and that was implemented over 4 years ago.

Even in Paragon I see some mixed usages (for instance the Toast component just uses a static max width definition), so it might be worth seeing if there is a current consensus...at all.

PR #555, opened 2021-05-20
feat: add PII suppression for hotjar

raq929:
Supresses hotjar tracking for sensitive form data, as well as table cells that contain user emails and/or codes.

### DIFF
--- a/src/components/Admin/AdminSearchForm.jsx
+++ b/src/components/Admin/AdminSearchForm.jsx
@@ -119,6 +119,7 @@ class AdminSearchForm extends React.Component {
                 value={searchQuery}
                 aria-labelledby="search-email-label"
                 className="py-0"
+                inputProps={{ 'data-hj-suppress': true }}
               />
             </div>
           </div>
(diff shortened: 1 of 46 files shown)


binodpant:
in general this is looking fine to me, just wondering how we will verify this, by checking staging hotjar? if we need to swarm on this at some point to validate everything PII is covered lmk!

These are just curiosity questions:
Also just a general q I have is, hotjar doc says it suppresses certain fields / values by default, such as email addresses, https://help.hotjar.com/hc/en-us/articles/115012439167-How-to-Suppress-Text-Images-and-User-Input-from-Collected-Data
I guess these changes are still needed though since the data we are suppressing in ending up in different fields and not covered by the default suppression?

binodpant (approved):
mainly just wondering if we need a post merge verify step of some kind for completeness, but the changes themselves look fine to me.

raq929:
We should do a post-merge verify once we get hotjar turned on on staging. Honestly, from that link it looks like some of what I did might be overkill, but, better safe than sorry?

binodpant:
yeah agreed, safer is better. I was just wondering as we add more code, we will also have to keep track of adding these. I wish they made a whitelisting instead so we could choose what to allow in.. but that also would require work .. just that it would be safe by default... still all that default suppression stuff they mention hopefully combined with your work is going to make this great!

binodpant (approved):
ok looks like the only analytics rendering piece in admin portal inside analyticspage so lgtm!

PR #563, opened 2021-05-25
feat: making bulk enrollment stepper footer sticky

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4548)

Notes for reviewers:
I've tried to include as many screenshots as possible but best way to fully test is fetching the branch and messing with the screen/window to ensure behaviors are as expected

**Screenshots:**

Desktop scrollable table filling screen, footer at bottom of page
image

Desktop table filling partial screen, footer at bottom of page
image

Desktop screen window resized smaller, footer at bottom of page
image

Mobile step one scrollable table top of page, footer at bottom of page
image

Mobile step one scrollable table bottom of page, footer at bottom of page
image

Mobile step two scrollable table top of page, foot at bottom of page
image

Note:
Tests altered to work for both PST and EST timezones

### DIFF
--- a/src/components/BulkEnrollmentPage/BulkEnrollment.scss
+++ b/src/components/BulkEnrollmentPage/BulkEnrollment.scss
@@ -1,4 +1,8 @@
 .bulk-enrollment {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+
   .alert {
     margin-top: 2rem;
   }
@@ -38,8 +42,13 @@
   }
 }
 
-.sticky-stepper-footer {
+.sticky-footer-wrapper {
+  height: 70vh;
   display: flex;
   flex-direction: column;
-  justify-content: space-between;
+
+  .stepper-step {
+    flex-grow: 1;
+    overflow: scroll;
+  }
 }
--- a/src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
@@ -23,7 +23,7 @@ const mockStore = configureMockStore([thunk]);
 
 const testCourseName = 'TestCourseName';
 const testCourseRunKey = 'TestCourseRun';
-const testStartDate = '2020-09-10T04:00:00Z';
+const testStartDate = '2020-09-10T10:00:00Z';
 
 const searchResults = {
   nbHits: 1,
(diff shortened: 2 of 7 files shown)


src/components/BulkEnrollmentPage/BulkEnrollment.scss
```diff
@@ -39,7 +43,12 @@
 }
 
 .sticky-stepper-footer {
+  height: 70vh;
```
taliaedX, line 46: this is still a more dynamic unit of measure, but I do wonder about putting a specific size in here.

taliaedX (approved):
The changes look fine to me. I'm not sure if this much CSS modification would better live somewhere centralized or if it's fine here. (aka I don't know if there is a best practice involving using other css classes or amending ones elsewhere), but I mean it's just css and as long as it works in mobile and desktop, I'm certainly fine with it.

src/components/BulkEnrollmentPage/BulkEnrollment.scss
alex-sheehan-edx: this was out of necessity, here 😞  I was pair programming with Lael and we were totally unable to get the footer at the bottom of the page while having a scrollable table and this seemed to be the only solution we could find. If there is a better answer I'm more than happy to change.

raq929: `vh` is somewhat dynamic, as it changes based on the height of the window. It's not 100% ideal, but I think it's OK?

src/components/BulkEnrollmentPage/BulkEnrollment.scss
```diff
@@ -39,7 +43,12 @@
 }
 
 .sticky-stepper-footer {
```
raq929, line 45: nit: I feel like this class is now not particularly well named? Maybe `sticky-footer-wrapper` would be better? Up to you.

raq929 (approved):

jennxgregory:
It it looks like both the stepper buttons as well as the universal footer are sticky in third from bottom mobile screenshot above. My concern is the valuable vertical real estate the universal footer is occupying on mobile devices, making the rest of the content on the page more difficult to consume. Ideally the universal footer would only display (on mobile or desktop) after the user scrolls past the stepper buttons. I don't see any issue with this on the desktop screenshots. 

That said, as we are not currently prioritizing optimizing the admin portal experience for mobile, if this change requires significant effort I can live with it. :)

PR #564, opened 2021-05-26
Update & Add Bulk Enrollment Page/Sidebar Tests

taliaedX:
Fixes some skipped tests as well as updates the Sidebar to ensure we can correctly test when bulk enrollment _isn't_ present.

### DIFF
--- a/src/components/BulkEnrollmentPage/index.test.jsx
+++ b/src/components/BulkEnrollmentPage/index.test.jsx
@@ -1,94 +1,140 @@
 import React from 'react';
 import { act, screen } from '@testing-library/react';
 import configureMockStore from 'redux-mock-store';
-import { MemoryRouter } from 'react-router-dom';
 import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
+import userEvent from '@testing-library/user-event';
 import { renderWithRouter } from '../test/testUtils';
 import BulkEnrollmentPage from './index';
 import '../../../__mocks__/react-instantsearch-dom';
 import LicenseManagerApiService from '../../data/services/LicenseManagerAPIService';
 import { ROUTE_NAMES } from '../EnterpriseApp/constants';
 import { ADD_COURSES_TITLE } from './stepper/constants';
+import '@testing-library/jest-dom';
 
 jest.mock('../../data/services/LicenseManagerAPIService', () => ({
   __esModule: true,
   default: {
     fetchSubscriptions: jest.fn(),
+    fetchSubscriptionUsersOverview: jest.fn(),
   },
 }));
+
+const testSlug = 'horsesRus';
 
 const sub1 = {
   uuid: 'foo',
   title: 'horse subscription',
-  enterpriseSlug: 'horsesRus',
+  enterpriseSlug: testSlug,
   startDate: 1621347682,
   expirationDate: 1721547682,
   licenses: {
     allocated: 5,
     total: 20,
   },
+  enterpriseCatalogUuid: 'blarghl2',
 };
 
 const sub2 = {
   uuid: 'bearz',
   title: 'bear subscription',
-  enterpriseSlug: 'horsesRus',
+  enterpriseSlug: testSlug,
   startDate: 1621347682,
   expirationDate: 1721547682,
   licenses: {
     allocated: 15,
     total: 40,
   },
+  enterpriseCatalogUuid: 'blarghl',
 };
 
+const mockSubscriptionCount = Promise.resolve({
+  data: [
+    { status: 'unassigned', count: 0 },
+    { status: 'assigned', count: 0 },
+    { status: 'activated', count: 2 }],
+});
 const subscriptions = Promise.resolve({
   data: {
-    results: {
-      subscriptions: [sub1, sub2],
-    },
+    results: [sub1, sub2],
     errors: null,
     setErrors: jest.fn(),
     forceRefresh: jest.fn(),
     loading: false,
+    count: 2,
   },
 });
-
-LicenseManagerApiService.fetchSubscriptions.mockReturnValue(subscriptions);
+const singleSubscription = Promise.resolve({
+  data: {
+    results: [sub1],
+    errors: null,
+    setErrors: jest.fn(),
+    forceRefresh: jest.fn(),
+    loading: false,
+    count: 1,
+  },
+});
+const mockSingleSubscriptionCount = Promise.resolve({
+  data: [
+    { status: 'unassigned', count: 0 },
+    { status: 'assigned', count: 0 },
+    { status: 'activated', count: 1 }],
+});
 
 const fakeStore = {
   portalConfiguration: {
     enterpriseId: 'foo',
+    enableCodeManagementScreen: false,
+    enableSubscriptionManagementScreen: true,
   },
 };
 const mockStore = configureMockStore([thunk]);
 const BulkEnrollmentWrapper = () => (
-  <MemoryRouter>
-    <Provider store={mockStore(fakeStore)}>
-      <BulkEnrollmentPage />
-    </Provider>
-  </MemoryRouter>
+  <Provider store={mockStore(fakeStore)}>
+    <BulkEnrollmentPage />
+  </Provider>
 );
 
+LicenseManagerApiService.fetchSubscriptions.mockImplementation(() => subscriptions);
+LicenseManagerApiService.fetchSubscriptionUsersOverview.mockImplementation(() => mockSubscriptionCount);
+
 describe('<BulkEnrollmentPage />', () => {
-  it('renders the bulk enrollment component', async () => {
-    renderWithRouter(<BulkEnrollmentWrapper />);
-    screen.getByText('Subscription Enrollment');
-    await act(() => subscriptions);
-  });
-  it.skip('renders the subscription picker', async () => {
-    renderWithRouter(<BulkEnrollmentWrapper />, { route: `/sluggo/admin/${ROUTE_NAMES.bulkEnrollment}` });
-    expect(await screen.findByText('Cohorts')).toBeInTheDocument();
-    expect(await screen.findByText(sub1.title)).toBeInTheDocument();
-    expect(await screen.findByText(sub2.title)).toBeInTheDocument();
+  describe('multiple subscriptions', () => {
+    it('renders the bulk enrollment component', async () => {
+      renderWithRouter(<BulkEnrollmentWrapper />);
+      screen.getByText('Subscription Enrollment');
+      await act(() => subscriptions);
+    });
+    it('renders the subscription picker', async () => {
+      renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}` });
+      expect(await screen.findByText('Cohorts')).toBeInTheDocument();
+      expect(await screen.findByText(sub1.title)).toBeInTheDocument();
+      expect(await screen.findByText(sub2.title)).toBeInTheDocument();
+    });
+    it('renders the course search page', async () => {
+      renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}/${sub1.uuid}` });
+      await act(() => subscriptions);
+      expect(screen.getByRole('heading', { level: 2 })).toHaveTextContent(ADD_COURSES_TITLE);
+    });
+    it('renders course search page when navigating via subscription picker', async () => {
+      renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}` });
+      await act(() => subscriptions);
+      // The expiration modal makes the enroll learners link inaccessible, so we need to dismiss it first.
+      const modalButton = await screen.getByRole('button', { name: 'OK' });
+      userEvent.click(modalButton);
 
-    await act(() => subscriptions);
+      const sub1Button = screen.getAllByRole('link', { name: 'Enroll learners' })[0];
+      userEvent.click(sub1Button);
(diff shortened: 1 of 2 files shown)


src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -15,13 +15,16 @@ jest.mock('../../data/services/LicenseManagerAPIService', () => ({
   __esModule: true,
   default: {
     fetchSubscriptions: jest.fn(),
+    fetchSubscriptionUsersOverview: jest.fn(),
   },
 }));
 
+const testSlug = 'horsesRus';
```
raq929, line 22: ❤️

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -76,19 +86,20 @@ describe('<BulkEnrollmentPage />', () => {
...
+    screen.findByText(sub1.title);
+    screen.findByText(sub2.title);
   });
-  it.skip('renders the course search page', async () => {
-    renderWithRouter(<BulkEnrollmentWrapper />, { route: `/sluggo/admin/${ROUTE_NAMES.bulkEnrollment}/${sub1.uuid}` });
-    expect(await screen.findByText('Cohorts')).toBeInTheDocument();
+  it('renders the course search page', async () => {
+    // await act(() => subscriptions);
+    act(() => {
+      flushPromises();
```
raq929, line 99: I think you may want this after your render.

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -76,19 +86,20 @@ describe('<BulkEnrollmentPage />', () => {
...
+      renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}/${sub1.uuid}` });
+    });
+    screen.findByText('Cohorts');
```
raq929, line 102: I think you may want something like:
```
act(async () => { await subscriptions; await mockSubscriptionCount; })
```
That *should* force the promises to resolve.

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,138 @@
...
+    { status: 'unassigned', count: 0 },
+    { status: 'assigned', count: 0 },
+    { status: 'activated', count: 1 }],
+});
 
 const fakeStore = {
   portalConfiguration: {
     enterpriseId: 'foo',
+    enableCodeManagementScreen: false,
+    enableSubscriptionManagementScreen: true,
```
taliaedX, line 88: I think I added these to suppress some of the stupid prop warnings.
There are still a bunch of them because they get generated before the promises resolve. 😑

raq929: If the components are attempting to render while promises are resolving, that sounds like a thing we should actually _fix_. Like, with a loading state.

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,138 @@
...
+  <Provider store={mockStore(fakeStore)}>
+    <BulkEnrollmentPage />
+  </Provider>
 );
 
-describe('<BulkEnrollmentPage />', () => {
+LicenseManagerApiService.fetchSubscriptions.mockImplementation(() => subscriptions);
+LicenseManagerApiService.fetchSubscriptionUsersOverview.mockImplementation(() => mockSubscriptionCount);
+
+describe('<BulkEnrollmentPage /> Multiple Subscriptions', () => {
```
raq929, line 101: Nit: I would do
```
describe('<BulkEnrollmentPage />', () => {
  describe('multiple subscriptions', () => {
  })
})
```
So that it's all the same test suite.

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,138 @@
...
-
+  });
+  it('renders the course search page', async () => {
+    renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}/${sub1.uuid}` });
+    expect(await screen.findAllByText(ADD_COURSES_TITLE)).toHaveLength(2);
+  });
+  it('renders course search page when navigating via subscription picker', async () => {
+    renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}` });
     await act(() => subscriptions);
+    // The expiration modal makes the enroll learners link inaccessible, so we need to dismiss it first.
```
raq929, line 120: Should we make our test data so that there is no expiration modal? And make a separate test that tests that the expiration modal appears?

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,138 @@
...
+    const modalButton = await screen.getByRole('button', { name: 'OK' });
+    userEvent.click(modalButton);
+
+    const sub1Button = screen.getAllByRole('link', { name: 'Enroll learners' })[0];
+    userEvent.click(sub1Button);
+    expect(screen.getByRole('heading', { level: 2 })).toHaveTextContent(ADD_COURSES_TITLE);
```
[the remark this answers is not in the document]
raq929: It appears once in the list of steps at the top and once as a header. I think this is better, tbh. It's possible that the title and the name of the step won't stay the same forever. This will break if we change the header level, but I think that's unlikely and kind of merits a test breaking bc accessibility.

src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -88,6 +89,25 @@ describe('<Sidebar />', () => {
...
+    const store = mockStore({
+      sidebar: {
+        ...initialState.sidebar,
+      },
+      portalConfiguration: {
+        enableSubscriptionManagementScreen: false,
+      },
+    });
+
+    const tree = renderer
```
raq929, line 103: I don't think we should add more snapshot tests if we can avoid it. I'd greatly prefer to render the component and verify that the link text isn't there.

src/components/BulkEnrollmentPage/index.test.jsx
raq929: If that doesn't make sense as part of this ticket, let's open a ticket for it? We shouldn't have a bunch of props errors in our test and definitely not in our console when the page is loading.

raq929 (approved):
Looks good to me! A nit and some comments. I would love to have actual loading states working so that you're *not* getting a bunch of PropTypes errors.

src/components/BulkEnrollmentPage/index.test.jsx
taliaedX: So I did some fancy dynamic expiration date generation and then I realized that we just always show the expiration modal. Perhaps we should revisit that.... 😠 

```
return (
    <SubscriptionDetailContextProvider subscription={subscriptionFurthestFromExpiration}>
      {subscriptionFurthestFromExpiration.daysUntilExpiration > 0 && (
        <SubscriptionExpirationBanner />
      )}
```
and then:
```
  const renderMessage = () => (
    <>
      {hasMultipleSubscriptions && daysUntilExpiration <= 0 ? (
        <>
          This batch has expired. You are able to view the statuses of your invited learners.
        </>
      ) : (
        <>
          Your subscription is {daysUntilExpiration} days from expiration.
          Contact the edX Customer Success team at
          {' '}
          <MailtoLink to="customersuccess@edx.org">customersuccess@edx.org</MailtoLink>
          {' '}
          to extend your contract.
        </>
```
>_<

src/components/BulkEnrollmentPage/index.test.jsx
taliaedX: These particular lines are for props that would normally exist at the very beginning of rendering the EnterpriseApp and get carried all the way through, so it seems fine to add them here as they are indeed expected.

For the others I mention, we have various loading states in a lot of spots, but it's likely we just aren't doing anything with them between components. Like I know we have a loading state in the Subscription Context we're wrapping everything with here, however, we may not be using that loading state in the child components. It's probably worth a separate ticket.

src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -207,4 +211,43 @@ describe('<Sidebar />', () => {
...
+        enableSubscriptionManagementScreen: true,
+      },
+    });
+
+    features.BULK_ENROLLMENT = true;
+
+    const { debug } = render(<SidebarWrapper store={store} />);
+    const subscriptionEnrollmentLink = screen.getByRole('link', { name: 'Subscription Enrollment' });
+    const subscriptionManagementLink = screen.getByRole('link', { name: 'Subscription Management' });
+    debug();
```
binodpant, line 249: remove this line?

src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -207,4 +211,43 @@ describe('<Sidebar />', () => {
...
+    const store = mockStore({
+      sidebar: {
+        ...initialState.sidebar,
+      },
+      portalConfiguration: {
+        enableSubscriptionManagementScreen: true,
+      },
+    });
+
+    features.BULK_ENROLLMENT = true;
```
binodpant, line 244: should this be true for all tests or?

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,139 @@
...
+    });
+    it('renders course search page when navigating via subscription picker', async () => {
+      renderWithRouter(<BulkEnrollmentWrapper />, { route: `/${testSlug}/admin/${ROUTE_NAMES.bulkEnrollment}` });
+      await act(() => subscriptions);
+      // The expiration modal makes the enroll learners link inaccessible, so we need to dismiss it first.
+      const modalButton = await screen.getByRole('button', { name: 'OK' });
+      userEvent.click(modalButton);
 
-    await act(() => subscriptions);
+      const sub1Button = screen.getAllByRole('link', { name: 'Enroll learners' })[0];
```
binodpant, line 125: ah so this is because we get multiple 'Enroll learners' buttons right? yeah I had asked Jen about having a button with a single enroll action on top looks like we removed that so this is fine for now

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,139 @@
...
   title: 'horse subscription',
-  enterpriseSlug: 'horsesRus',
+  enterpriseSlug: testSlug,
   startDate: 1621347682,
   expirationDate: 1721547682,
   licenses: {
     allocated: 5,
     total: 20,
   },
+  enterpriseCatalogUuid: 'blarghl2',
```
binodpant, line 35: great catalog names! 🎉

src/containers/Sidebar/Sidebar.test.jsx
taliaedX: I had originally made it true for all tests, but then that just affects snapshot tests, so meh.

src/components/BulkEnrollmentPage/index.test.jsx
taliaedX: For each subscription we have an Enroll learners button yeah. I struggled for a while trying to basically link 2 screen finds together (ie "get the enroll button under this other div/heading/etc.") and there's a "get closest <element> to" function, but there wasn't an obvious or accessible role/name/etc. for the actual names of the subscriptions to narrow this down.

TL;DR yeah I just picked one at random because there's multiples for now.

src/components/BulkEnrollmentPage/index.test.jsx
```diff
@@ -1,94 +1,139 @@
 import React from 'react';
```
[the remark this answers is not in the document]
binodpant: @taliaedX confirmed this test covers this case, so I am all good (and we also have manual testing so scenario coverage is looking quite good now with this PR plus manual testing): https://github.com/edx/frontend-app-admin-portal/blob/master/src/components/BulkEnrollmentPage/stepper/BulkEnrollmentSubmit.test.jsx#L270

binodpant (approved):
lgtm! covers the cases from the test sheet we marked for integ testing

src/components/BulkEnrollmentPage/index.test.jsx
binodpant: yeah I am not too big a fan of not having the top level bulk button.. I liked having it personally.. but this is more a Jen chat than here... so all good

src/containers/Sidebar/Sidebar.test.jsx
binodpant: Yeah I think we should eventually really limit if not remove all snapshot tests, really only useful for the smallest component boundaries I think

src/containers/Sidebar/Sidebar.test.jsx
binodpant: big plus 1

binodpant: but also okay if we want to ticket it separately in case we think we lose coverage by not having this test

src/components/BulkEnrollmentPage/index.test.jsx
binodpant: this sounds like a new test ticket. Yeah ideally we avoid having that dialog altogether in tests, unless we are testing for it

src/components/BulkEnrollmentPage/index.test.jsx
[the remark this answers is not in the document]
taliaedX: There is also a maybe different Enroll learners button test than you were thinking of that *is* present in this exact test file.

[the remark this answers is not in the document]
raq929: There is an existing unit test for that, but we if we want an integration test, this is the place to do it.

PR #565, opened 2021-05-27
feat: enable hotjar in admin portal

taliaedX:
Enables the admin portal to send various metrics and data to Hotjar.

*Notes:*
- I put it in the wrapper and not the main js app init file, because it utilizes React functionalities (hooks) that don't work in pure JS.
- Because the values of the app id and version number should be ints not strings, I don't really want dummy values populating our env files. Thus the app id listed there is for the *STAGE* admin portal app in Hotjar. If I put `HOTJAR_APP_ID=''` someone is going to stick the number in those quotes and it's going to break. Also I don't see a reason why this app id would be private since the other option they recommend is embedding it directly into the HTML.
- I actually don't want to make an accompanying edx-internal PR just yet, because I want to make sure we aren't falling back on development env vars (I don't _think_ we do, but want to double-check in stage)


src/components/App/index.jsx
```diff
@@ -18,6 +20,14 @@ import store from '../../data/store';
 
 const AppWrapper = () => {
   const apiClient = getAuthenticatedHttpClient();
+
+  if (process.env.HOTJAR_APP_ID) {
+    const { initHotjar } = useHotjar();
+    useEffect(() => {
```
alex-sheehan-edx, line 26: not being fully knowledgable on `useEffect`, does this just allow us to async init hot jar as a hook?

taliaedX: That sounds....plausible. 😅 
Legit just took the snippet from the react-use-hotjar readme.

alex-sheehan-edx (approved):
LGTM 🥳

PR #567, opened 2021-05-28
feat: Add partner column, show first partner only, for now

binodpant:
Decision per UX sync meeting:

When more than one partner in algolia data, just display first one, for now

ENT-4608

looks like this with change:

Screen Shot 2021-06-02 at 2 19 41 PM

### DIFF
--- a/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
@@ -27,6 +27,7 @@ export const ENROLL_TEXT = 'Enroll learners';
 export const TABLE_HEADERS = {
   courseName: 'Course name',
   courseStartDate: 'Course start date',
+  partnerName: 'Partner',
   enroll: '',
 };
 
@@ -94,6 +95,10 @@ export const BaseCourseSearchResults = (props) => {
       Header: TABLE_HEADERS.courseStartDate,
       accessor: 'advertised_course_run.start',
       Cell: FormattedDateCell,
+    },
+    {
+      Header: TABLE_HEADERS.partnerName,
+      accessor: 'partners[0].name',
     },
   ], []);
 
--- a/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
@@ -92,13 +92,13 @@ export const BaseCourseSearchResults = (props) => {
       Cell: ({ value, row }) => <CourseNameCell value={value} row={row} enterpriseSlug={enterpriseSlug} />,
     },
     {
+      Header: TABLE_HEADERS.partnerName,
+      accessor: 'partners[0].name',
+    },
+    {
       Header: TABLE_HEADERS.courseStartDate,
       accessor: 'advertised_course_run.start',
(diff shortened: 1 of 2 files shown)


src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -95,6 +96,10 @@ export const BaseCourseSearchResults = (props) => {
       accessor: 'advertised_course_run.start',
       Cell: FormattedDateCell,
     },
+    {
+      Header: TABLE_HEADERS.partnerName,
+      accessor: 'partners[0].name',
```
taliaedX, line 101: I'm struggling to find it in admin portal. Where does the data for this actually come from? I assume an API call to something, but I'm unclear on what. I'd just like to verify the data format here.

taliaedX (commented):
I just want to verify the data format of the partners array and my guess is Jenn should also review this? But otherwise it looks good to me.

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
binodpant: ah it's in the response of algolia. So it looks like this in devtools... Also if you look directly in algolia, this format also exists there in the catalog data

```
partners: 
  0: {name: "edX",…}
```

jennxgregory:
Can we swap the order of the Partner and Course start date columns, so Partner is next to the Course name? Otherwise this looks good! The table contents will be a little tight while the "Enroll learners" links are still in there but that should be resolved for V2.

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -90,6 +91,10 @@ export const BaseCourseSearchResults = (props) => {
       // eslint-disable-next-line react/prop-types
       Cell: ({ value, row }) => <CourseNameCell value={value} row={row} enterpriseSlug={enterpriseSlug} />,
     },
+    {
+      Header: TABLE_HEADERS.partnerName,
+      accessor: 'partners[0].name',
```
alex-sheehan-edx, line 96: What is the cell field here and why are we not using it for this column?

alex-sheehan-edx (approved):
One question about table cells but otherwise LGTM!

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
taliaedX: Awesome. Thank you

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
binodpant: cell field is only to customize how it's rendered. Data table by default renders as a span I think which suffices for this use case

taliaedX (approved):
Thanks for clarifying! LGTM

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
binodpant: it comes out like this by default 

```
<td role="cell"><span class="pgn__data-table-cell-wrap">edx</span></td>
```

PR #571, opened 2021-06-07
add feature flag for analytics api v1

muhammad-ammar:
**Description:** Add a feature flag so that Analytics API V1 can be tested conditionally on production.


adamstankiewicz (approved):
LGTM. Will this approach work well enough for testing this new API endpoint?

muhammad-ammar:
> LGTM. Will this approach work well enough for testing this new API endpoint?

yes. This is working. Thank You!

PR #573, opened 2021-06-10
feat: Filtering by user email in learner stepper screen

binodpant:
ENT-4573

Getting two calls to fetchData each time I type a key, but the filtering is working

Test only covers we are calling backend correctly, plus display of filter options

Screen Shot 2021-06-10 at 3 12 02 PM

Screen Shot 2021-06-10 at 3 11 33 PM

Also added debouncing with a 400ms default debounce interval, see screen record (IGNORE THE KEEPER ICON ON THE TEXT BOX!). Verified it no longer goes crazy requests per keystroke unlike before the debounce.

https://user-images.githubusercontent.com/528166/121919045-a19e4700-cd04-11eb-86a4-9205d6f2ca80.mov


src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
```diff
@@ -72,9 +72,16 @@ const AddLearnersStep = ({
...
   const fetchData = useCallback((tableInstance = {}) => {
     const pageIndex = tableInstance.pageIndex || INITIAL_PAGE_INDEX;
+    let options = { active_only: 1, page_size: LEARNERS_PAGE_SIZE, page: pageIndex + 1 };
+
+    const { filters } = tableInstance;
+    const emailFilter = filters.find(item => item.id === 'userEmail');
+    if (emailFilter) {
+      options = { ...options, search: emailFilter.value };
+    }
     LicenseManagerApiService.fetchSubscriptionUsers(
```
raq929, line 82: What was the decision about debouncing this call? I think it would be good to avoid a call on every keystroke.

raq929 (approved):
Looks good in general, but I think debouncing might be good.

Just to be sure, is this resetting the page to zero when the filter is set? I'm not seeing logic for that (though it might be something that `react-table` does, so.)

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
binodpant: On reset of filtering: I double checked, and yeah the removal of all characters from the filter box does indeed reset the page to show all learners again (in that page), so looks like reset is built in


Debouncing: yeah I have not done any debouncing here yet. Agreed we should have it. Will take a look.

binodpant: added Debouncing! it works great on manually but so far I have not figure out how to fix the test that now fails to find that the 'search' param was inserted in the calls made

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
```diff
@@ -59,12 +64,25 @@ describe('AddLearnersStep', () => {
...
     expect(await screen.findByText('Search Email')).toBeInTheDocument();
     await act(() => mockTableData);
+    userEvent.type(screen.getByPlaceholderText('Search Email'), 'beAR');
+    expect(await screen.findByText(/Filtered by userEmail/)).toBeInTheDocument();
+    expect(await screen.findByText('Clear Filters')).toBeInTheDocument();
+    const { subscriptionUUID } = defaultProps;
+    // multiple calls will occur to this function, we only test for the last one
+    // for correctness, and don't test backend filtering part here (tested in backend).
+    await screen.findByDisplayValue('beAR');
+    expect(LicenseManagerApiService.fetchSubscriptionUsers).toHaveBeenLastCalledWith(
```
binodpant, line 80: trying to find out how to make this expect pass again, after debounce, it now does not verify that the 'search' was passed

binodpant: asking jest to runAllTimers seems to fix this
also removed useCallback since I already use useMemo now

binodpant:
looks like I have run into this issue

https://github.com/facebook/jest/issues/3465

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
binodpant: spoke too early, https://github.com/facebook/jest/issues/3465 blocks this right now...

binodpant: yeah so can't find any way to make this work and it looks like there is some known issue with current jest plus lodash.debounce when using real or fake timers. I do not want to spend too much time here and since the test covers that we do call the underlying api call, and mocking out debounce altogether seems to get us past this so going to go with this

PR #575, opened 2021-06-11
fix: Bulk Enrollment Selection Option Consistency

taliaedX:
Selecting a course on the course selection step of the Bulk Enrollment stepper was not triggering the selection options/buttons/tool tip (what are we calling this?). It only triggered when clicking "Next" and then clicking Back to go back to the Course step. It also then would show "0 of N selected" which is not how the rest of the data tables behave.

We believe it is doing this because the underlying react-tables' selection is not refreshing the data until reload because of the way we are partially using selection props here and partially using our overrides.

This PR fixes the following:
- Fixes the selection behaviour by manually specifying selectedFlatRows in CourseSearchResults (this was already present in the AddLearner data table options).
- Removes unnecessary/duplicated props from the DataTables rendering in both CourseSearchResults and the AddLearners step components.
- Adds test to verify selection behaviour is present and functioning.

### DIFF
--- a/src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Provider } from 'react-redux';
 import { mount } from 'enzyme';
-import { screen } from '@testing-library/react';
+import { screen, within } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import configureMockStore from 'redux-mock-store';
 import userEvent from '@testing-library/user-event';
@@ -25,12 +25,14 @@ const testCourseName = 'TestCourseName';
 const testCourseRunKey = 'TestCourseRun';
 const testStartDate = '2020-09-10T10:00:00Z';
 
+const testCourseName2 = 'TestCourseName 2';
+const testCourseRunKey2 = 'edX+testcourse2';
+const testStartDate2 = '2021-10-10T10:00:00Z';
+
 const searchResults = {
-  nbHits: 1,
-  hitsPerPage: 10,
-  pageIndex: 10,
-  pageCount: 5,
-  nbPages: 6,
+  nbHits: 2,
+  hitsPerPage: 20,
+  nbPages: 1,
   hits: [
     {
       title: testCourseName,
@@ -41,8 +43,17 @@ const searchResults = {
       key: 'foo',
       partners: [{ name: 'edX' }, { name: 'another_unused' }],
     },
+    {
+      title: testCourseName2,
+      advertised_course_run: {
+        key: testCourseRunKey2,
+        start: testStartDate2,
+      },
+      key: 'foo2',
+      partners: [{ name: 'edX' }, { name: 'another_unused' }],
+    },
   ],
-  page: 3,
+  page: 0,
 };
 
 const searchState = {
@@ -107,7 +118,7 @@ describe('<CourseSearchResults />', () => {
   it('renders search results', () => {
     const wrapper = mount(<CourseSearchWrapper />);
 
-    // Four header columns, one for sorting, one for Course Name, one for Course Run, one for the enroll column
+    // 5 header columns: selection, Course name, Partner, Course Date, and enrollment
     const tableHeaderCells = wrapper.find('TableHeaderCell');
     expect(tableHeaderCells.length).toBe(5);
     expect(tableHeaderCells.at(1).prop('Header')).toBe(TABLE_HEADERS.courseName);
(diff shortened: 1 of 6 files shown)


src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
```diff
@@ -136,6 +147,28 @@ describe('<CourseSearchResults />', () => {
     const wrapper = mount(<CourseSearchWrapper props={{ ...defaultProps, isSearchStalled: true }} />);
     expect(wrapper.find(Skeleton)).toHaveLength(1);
   });
+  it('shows selection options when at least one course is selected', () => {
+    renderWithRouter(<CourseSearchWrapper {...defaultProps} />);
+    const rowToSelect = screen.getByText(testCourseName2).closest('tr');
+    userEvent.click(within(rowToSelect).getByTestId('selectOne'));
```
taliaedX, line 153: you could definitely just do:
```
userEvent.click(screen.getAllByTestId('selectOne')[0]);
```

I honestly just wanted an excuse to use `within()` since I just learned about it (but it's more useful where we actually have unique identifiers up one level, whereas here they both are under generic table row and the same rowGroup with no differentiating ids or elements)

src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
```diff
@@ -107,17 +118,17 @@ describe('<CourseSearchResults />', () => {
   it('renders search results', () => {
     const wrapper = mount(<CourseSearchWrapper />);
 
-    // Four header columns, one for sorting, one for Course Name, one for Course Run, one for the enroll column
+    // 5 header columns: selection, Course name, Partner, Course Date, and enrollment
```
taliaedX, line 121: This comment was confusing the hell out of me since it didn't match the header names nor count. If there is a reason it was labeled like this I can put it back, but it was just really, REALLY confusing.

src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
```diff
@@ -41,8 +43,17 @@ const searchResults = {
...
+      advertised_course_run: {
+        key: testCourseRunKey2,
+        start: testStartDate2,
+      },
+      key: 'foo2',
+      partners: [{ name: 'edX' }, { name: 'another_unused' }],
+    },
   ],
-  page: 3,
+  page: 0,
```
taliaedX, line 56: The Algolia results made no sense to me and were also making it impossible to test multiple results on the same page.

src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
raq929: I love it.

src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
raq929: Do we need it at all?

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
```diff
@@ -11,7 +11,6 @@ import { logError } from '@edx/frontend-platform/logging';
 import { Link } from 'react-router-dom';
 import { BulkEnrollContext } from '../BulkEnrollmentContext';
 import { ADD_LEARNERS_TITLE } from './constants';
-import { convertToSelectedRowsObject } from '../helpers';
```
raq929, line 14: Go ahead and delete `convertToSelectedRowsObject` entirely, if you don't mind. These were the only to components using it.

raq929 (approved):
This looks great, thanks for fixing my mess here. 😅 

One comment to remove a function that we're no longer using.
If you wanted to add a test to pass CI, I'd love to see an integration test where we click forward to the Add Learners page and back to bulk enrollment and verify that the correct courses are selected...

src/components/BulkEnrollmentPage/CourseSearchResults.test.jsx
taliaedX: Not particularly. 🤷  I just figured if it was going to exist it should at least make sense lmao

src/components/BulkEnrollmentPage/stepper/BulkEnrollmentStepper.test.jsx
```diff
@@ -104,12 +104,13 @@ describe('BulkEnrollmentStepper', () => {
...
     await act(() => mockEmailResponse);
   });
-  it('clicking previous from Add learners brings you back to Add courses', async () => {
+  it('clicking previous from Add learners brings you back to Add courses and retains selection', async () => {
     renderWithRouter(<StepperWrapper {...defaultProps} />);
     navigateToAddLearners();
     const prevButton = screen.getByTestId(PREV_BUTTON_TEST_ID);
     userEvent.click(prevButton);
     expect(screen.getAllByText(ADD_COURSES_TITLE)).toHaveLength(2);
+    expect(screen.getByText('All 2 selected')).toBeInTheDocument();
```
taliaedX, line 113: I still cannot for the life of me figure out what the 2 courses are. The debug printout doesn't contain the 2 levels deep where this stuff is generated, but it is definitely looking for 2 courses and if I swap above to selecting one I do need to swap this to `1 selected.` so it is working as intended...I just don't see courses being mocked here, so I'm confuuuuuused. 😩

raq929: The courses are mocked in the `src/__mocks__` folder! I'm mocking all the Algolia components.

taliaedX:
Hmm even rerunning the build, coveralls isn't actually rebuilding. 😑 
Well I know it's a lie anyways, so going to squash and merge

taliaedX:
okay it did rerun now and it still thinks there is a decrease. 😠

PR #576, opened 2021-06-14
ENT-4507: feat: Front end work for showing Notification banner in Admin Portal

saleem-latif:
**JIRA Ticket:** [ENT-4507](https://openedx.atlassian.net/browse/ENT-4507)

__Description:__
As an enterprise admin, I want to see relevant product updates pop up in an informational message when I log into my Admin Portal in order that I may receive 'in-application notification' of new developments on edX.

__Acceptance Criteria:__
1. Front end: the ability to see that banner pop up when the admin logs in on the home page of the Admin Portal.

__Testing Instructions:__
1. Install the latest version of  edx-enterprise (note: backend was added in  `edx-enterprise==3.26.0` you need this version to test this PR)
2. Go to the django admin page for admin notification manager `/admin/enterprise/adminnotification/`
2. Add a new notification.
3. Now open admin portal and select any enterprise customer if not already selected. 
4. You should see a banner at the top with the message specified in step 2.
5. Click on the cross button at the top right corner of the banner.
6. Notification should go away, refresh the page and make sure banner does not re-appear.

### DIFF
--- a/src/components/FeatureAnnouncementBanner/index.jsx
+++ b/src/components/FeatureAnnouncementBanner/index.jsx
@@ -1,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from '@edx/paragon';
+import { logError } from '@edx/frontend-platform/logging';
+
+import LmsApiService from '../../data/services/LmsApiService';
+
+const FeatureAnnouncementBanner = ({ enterpriseSlug }) => {
+  const [enterpriseNotificationBanner, setEnterpriseNotificationBanner] = useState({
+    text: null,
+    notificationId: null,
+  });
+  const [isRead, setIsRead] = useState(false);
+
+  const markAsRead = () => {
+    setIsRead(true);
+    LmsApiService.markBannerNotificationAsRead({
+      enterprise_slug: enterpriseSlug,
+      notification_id: enterpriseNotificationBanner.notificationId,
+    });
+  };
+
+  useEffect(() => {
+    LmsApiService.fetchEnterpriseBySlug(enterpriseSlug)
+      .then((response) => {
+        if (response.data && response.data.enterprise_notification_banner) {
+          setEnterpriseNotificationBanner({
+            text: response.data.enterprise_notification_banner.text,
+            notificationId: response.data.enterprise_notification_banner.id,
+          });
+        }
+      })
+      .catch((error) => {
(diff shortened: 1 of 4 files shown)


src/data/services/LmsApiService.js
```diff
@@ -203,6 +207,14 @@ class LmsApiService {
   static fetchEnterpriseCustomerCatalogs(enterpriseId) {
     return LmsApiService.apiClient().get(`${LmsApiService.enterpriseCustomerCatalogsUrl}?enterprise_customer=${enterpriseId}`);
   }
+
+  static fetchUserDetailsFromEmail(emails) {
+    return LmsApiService.apiClient().post(`${LmsApiService.accountsSearchEmailsUrl}`, emails);
+  }
```
muhammad-ammar, line 213: I am not seeing usage of this?

saleem-latif: I think it got added during rebase, I will remove it now.

src/components/FeatureAnnouncementBanner/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const markAsRead = () => {
+    setIsRead(true);
+    LmsApiService.markBannerNotificationAsRead({
+      enterprise_slug: enterpriseSlug,
+      notification_id: enterpriseNotificationBanner.notificationId,
+    });
+  };
+
+  useEffect(() => {
+    LmsApiService.fetchEnterpriseBySlug(enterpriseSlug)
```
muhammad-ammar, line 25: We already fetch enterprise customer information when an admin selects an enterprise, can not we use that information instead of fetch it again?

saleem-latif: I tried but I do not think we can.

saleem-latif: The main  reason is that the component is rendered as a result of page routing and url only  contains enterprise slug and we must fetch all data from that slug and not depend on anything else.

sameenfatima78 (approved):
Test cases are missing for this new functionality, other than that it looks good

src/components/FeatureAnnouncementBanner/index.jsx
muhammad-ammar: isn' t there any redux store or anything like that?

zamanafzal (approved):

muhammad-ammar (approved):
LGTM

PR #578, opened 2021-06-15
Coupon details refactor and testing

raq929:
- Moves logic from CouponDetails component into helper functions and test it
- Adds and updates CouponDetails tests (mostly on the container)
- Removes CouponDetails snapshots in favor of unit tests
- Updates Paragon and updates snapshots that changed with the Paragon update


package.json
```diff
@@ -21,20 +21,19 @@
...
   "dependencies": {
     "@babel/plugin-transform-runtime": "7.12.1",
-    "@edx/brand": "npm:@edx/brand-openedx@^1.1.0",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-enterprise-catalog-search": "0.1.10",
     "@edx/frontend-enterprise-logistration": "0.1.11",
     "@edx/frontend-enterprise-utils": "0.1.7",
-    "@edx/frontend-platform": "1.11.0",
-    "@edx/paragon": "14.12.4",
+    "@edx/frontend-platform":"1.11.0",
```
taliaedX, line 28: Nit: what happened to the space? 👾

src/components/CouponDetails/ActionButton.jsx
```diff
@@ -0,0 +1,124 @@
...
+      used: PropTypes.number.isRequired,
+      num_assignments: PropTypes.number,
+    }),
+    code: PropTypes.string.isRequired,
+  }).isRequired,
+  handleCodeActionSuccess: PropTypes.func.isRequired,
+  setModalState: PropTypes.func.isRequired,
+};
+
+export default ActionButton;
```
taliaedX, line 124: I may be losing my mind, but I don't see anywhere where we actually use ActionButton outside of the test files for it?
Everywhere else explicitly uses `ActionButtonWithModal` and that component *doesn't* call/inherit `ActionButton`
Is there some sort of implicit functionality I'm missing? (I even pulled the branch locally and searched through VSCode since it's more reliable than Github search and I'm still coming up with nothing outside of the base definition and test files.

src/components/CouponDetails/constants.js
```diff
@@ -0,0 +1,127 @@
...
+  redeemed: 'Redeemed',
+};
+
+export const ACTION_LABELS = {
+  remind: 'Remind',
+  assign: 'Assign',
+  revoke: 'Revoke',
+};
+
+export const ACTION_TYPES = {
```
taliaedX, line 23: Not sure if this would be more annoying or not, but why not do something like:
```
export const ACTIONS = {
  remind: {'label': 'Remind', 'value': 'remind'},
  revoke:
  [...]
}
```
Then they'd be referenced by ACTIONS.revoke.label and ACTIONS.revoke.action
The plus side: only have to remember to add the action to one spot.
Downside: Requires one further level of accessor?
(This goes for `COUPON_FILTER_*` too)

src/components/CouponDetails/helpers.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (shouldHidePartialRedeem) {
+    options = FILTER_OPTIONS.filter(option => option.value !== COUPON_FILTER_TYPES.partiallyRedeemed);
+  }
+
+  return options;
+};
+
+export const getFirstNonDisabledOption = (options) => {
+  const firstNonDisabledOption = options.find(option => !option.disabled);
```
taliaedX, line 19: Do we care about order of operations here? Ie will the "first" option always be the same and does that matter?

src/components/CouponDetails/helpers.js
```diff
@@ -0,0 +1,64 @@
...
+  isAssignView,
+  isRedeemedView,
+  hasTableData,
+  couponAvailable,
+  numUnassignedCodes,
+  numSelectedCodes,
+}) => ([{
+  label: ACTION_LABELS.assign,
+  value: ACTION_TYPES.assign,
+  disabled: !isAssignView || isRedeemedView || !hasTableData || !couponAvailable || numUnassignedCodes === 0, // eslint-disable-line max-len
```
taliaedX, line 38: so many logic gates! 😮

src/components/CouponDetails/helpers.js
```diff
@@ -0,0 +1,64 @@
...
+    return false;
+  }
+
+  if (hasAllCodesSelected) {
+    return true;
+  }
+
+  return (
+    selectedCodes.length === tableData.results.length
+    && selectedCodes.length !== tableData.count
```
taliaedX, line 62: I've read this 3 times and I still cannot make sense of it.
As we are returning based on these values being different, why wouldn't the tableData result set length equal the count of itself (tableData)?

src/components/CouponDetails/helpers.test.jsx
```diff
@@ -0,0 +1,162 @@
...
+      value: 'redeemed',
+    }];
+    expect(getFilterOptions(usageLimitation)).toEqual(expected);
+  });
+});
+
+describe('getFirstNonDisabledOption', () => {
+  test.each([
+    [[{ value: 'foo', disabled: true }, { value: 'bar' }], 'bar'],
+    [[{ value: 'foo' }, { value: 'bar', disabled: false }, { value: 'baz', disabled: true }], 'foo'],
```
taliaedX, line 37: do we expect the potential that more than 1 is not disabled? (I think this is basically the same sort of question I asked before about "first" result)

src/components/CouponDetails/helpers.test.jsx
```diff
@@ -0,0 +1,162 @@
...
+    expect(getFirstNonDisabledOption(options)).toEqual(expected);
+  });
+  it('returns the first option if all are disabled', () => {
+    const options = [{ value: 'foo', disabled: true }, { value: 'bar', disabled: true }, { value: 'baz', disabled: true }];
+    expect(getFirstNonDisabledOption(options)).toEqual('foo');
+  });
+});
+
+describe('shouldShowSelectAllStatusAlert', () => {
+  test.each([
```
taliaedX, line 49: also I just noticed this format instead of the `ddt` data parameters and I love it.

taliaedX (commented):
first pass: just some questions. Will probably want to run this locally given how large it is.

package.json
raq929: I DON'T KNOW.

src/components/CouponDetails/ActionButton.jsx
raq929: Let me check, it may have been lost in the rebase, it was a major rebase.

src/components/CouponDetails/constants.js
raq929: Yeah, it got built this way because I did it one by one, but I like that.

src/components/CouponDetails/helpers.js
raq929: This is after I removed one about visibility!

src/components/CouponDetails/helpers.js
raq929: Yeah, the first option will always be the same, we never re-order the options.

src/components/CouponDetails/helpers.js
raq929: Honestly, I think in the case that `tableData.results.length` is different from `tableData.count` something has gone *seriously* wrong. I didn't change this code, just pulled it out and tested it. I'll to a `git blame` and see if the person who wrote it is still around to see if there's a good reason for it being this way...

src/components/CouponDetails/helpers.test.jsx
raq929: Now that I pulled out the public and private codes, we don't really expect that? But we did at first, and we might again.

src/components/CouponDetails/helpers.test.jsx
raq929: ME TOOOOO

src/components/CouponDetails/helpers.js
taliaedX: Interesting. I figured Javascript might just arbitrarily pull options rather than treating the options like an index in a list (ie immovable)

package.json
```diff
@@ -21,20 +21,19 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@babel/plugin-transform-runtime": "7.12.1",
-    "@edx/brand": "npm:@edx/brand-openedx@^1.1.0",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
```
edx-abolger, line 24: I'm assuming this pin is very deliberate?

src/components/CouponDetails/ActionButton.jsx
```diff
@@ -0,0 +1,124 @@
...
+            assigned_to: code.assigned_to,
+            code: code.code,
+          }}
+          onSuccess={response => handleCodeActionSuccess(ACTION_TYPES.revoke, response)}
+        />
+      </>
+    );
+  }
+
+  if (isPublic) {
```
edx-abolger, line 70: Nit: can we put the null logic for 'isPublic' up at the top with all the other null logics?

src/components/CouponDetails/CouponBulkActions.jsx
```diff
@@ -0,0 +1,56 @@
...
+} from '@edx/paragon';
+
+const CouponBulkActions = ({
+  value, options, disabled, onChange, handleBulkAction,
+}) => (
+  <div className="bulk-actions col-12 col-md-4 text-md-right mt-3 m-md-0 d-flex justify-content-end">
+    <Form.Group className="mt-1 flex-grow-1" controlId="bulkActions" name="bulk-actions">
+      <Form.Control
+        className="float-right w-50"
+        floatingLabel="Bulk action"
```
edx-abolger, line 15: Nit/wish: since this is brand new and a refactor and you already did it for ActionButton can we put all the actual text in a constants file? For my future translation wishes?

src/data/actions/coupons.js
```diff
@@ -63,7 +63,7 @@ const fetchCouponOrders = options => (
         logError(error);
         // This endpoint returns a 404 if no data exists,
         // so we convert it to an empty response here.
-        if (error.response.status === 404) {
+        if (error.response?.status === 404) {
```
edx-abolger, line 66: ? is to handle null, right? What do we do here if we don't get a response at all? Is there a higher level timeout?

src/containers/CouponDetails/CouponDetails.test.jsx
```diff
@@ -155,76 +163,68 @@ describe('CouponDetails container', () => {
...
+      renderWithRouter(<CouponDetailsWrapper store={store} isExpanded />);
+      userEvent.selectOptions(screen.getByLabelText('Filter by code status'), filterType);
+
+      DEFAULT_TABLE_COLUMNS[filterType].forEach(({ label }) => {
+        expect(screen.getByText(label)).toBeInTheDocument();
+      });
     });
 
-    it.skip('with assignment and reminder dates table data', () => {
+    it('shows Assign button for an available coupon', () => {
```
edx-abolger, line 209: Woohoo for no skipping tests!

src/components/NumberCard/index.jsx
```diff
@@ -212,7 +212,7 @@ class NumberCard extends React.Component {
                 aria-controls={`footer-body-${id}`}
               >
                 <div className="d-flex justify-content-between align-items-center">
-                  <div className="details-btn-text">
+                  <div className="details-btn-text mr-1">
                     {detailsExpanded ? 'Detailed breakdown' : 'Details'}
```
edx-abolger, line 216: Nit/wish: again is it worth it to move all these to a constants file for anticipated i18n?

src/components/CouponDetails/index.jsx
```diff
@@ -634,8 +408,13 @@ class CouponDetails extends React.Component {
       assignments_remaining: `${code.redemptions.total - code.redemptions.used - code.redemptions.num_assignments}`,
       assignment_date: `${code.assignment_date}`,
       last_reminder_date: `${code.last_reminder_date}`,
-      is_public: code.is_public ? 'Public' : 'Private',
-      actions: this.getActionButton(code),
+      actions: <ActionButton
```
edx-abolger, line 411: @taliaedX I found the ActionButton usage!

src/components/CouponDetails/index.jsx
```diff
@@ -658,7 +437,7 @@ class CouponDetails extends React.Component {
 
   isTableLoading() {
     const { couponDetailsTable } = this.props;
-    return couponDetailsTable && couponDetailsTable.loading;
+    return !!(couponDetailsTable && couponDetailsTable.loading);
```
edx-abolger, line 440: Ha, I thought only old school embedded C developers like me used this :P

src/components/CouponDetails/index.jsx
taliaedX: Ah hah! I don't know how I missed that one.

src/components/CouponDetails/ActionButton.jsx
raq929: Nope! It's there on line 411 of `src/components/CouponDetails/index.jsx`.

package.json
raq929: We pin everything now, I just saved the bot a PR, that's all.

src/components/CouponDetails/ActionButton.jsx
raq929: Possibly. I feel like there might have been a reason for this here, but I will double check it.

src/data/actions/coupons.js
raq929: This will dispatch the `fetchCouponOrdersFailure` action if there's no response. Mostly this change is for my dev environment which was throwing an uncaught error when the server was down instead of dispatching the failure action. But now that I think of it, I'm not sure what it is that got returned that *had* a response on it but not a status. I'll throw another `?` in there just in case.

raq929: Really this fix is for me in development so that I don't get a blank page when my backend isn't running. We are not doing higher level error handling, afaik. (ie, the class does not do any error handling). I can add an extra question mark after error so that if we actually get *nothing* back, it doesn't throw.

src/components/CouponDetails/index.jsx
raq929: Guaranteed boolean all the time!

src/components/CouponDetails/ActionButton.jsx
```diff
@@ -5,7 +5,7 @@ import {
 } from '@edx/paragon';
 import RemindButton from '../RemindButton';
 import RevokeButton from '../RevokeButton';
-import { ACTION_TYPES, ACTION_LABELS, COUPON_FILTER_TYPES } from './constants';
+import { ACTIONS, COUPON_FILTERS, COUPON_FILTER_TYPES } from './constants';
```
taliaedX, line 8: So the only usage of `COUPON_FILTER_TYPES` I see still in this file is at the very bottom in the prop definition and I'm wondering if it doesn't need to be there either?

`selectedToggle: PropTypes.oneOf(Object.values(COUPON_FILTER_TYPES)).isRequired,`

src/components/CouponDetails/constants.js
```diff
@@ -7,50 +7,78 @@ export const COUPON_FILTER_TYPES = {
...
-};
-
-export const ACTION_LABELS = {
-  remind: 'Remind',
-  assign: 'Assign',
-  revoke: 'Revoke',
+export const COUPON_FILTERS = {
+  unassigned: {
+    label: 'Unassigned',
+    value: COUPON_FILTER_TYPES.unassigned,
```
taliaedX, line 13: Is there a reason to not just define this directly? We aren't using the Types in any other spot anymore and it's just a single word each.

As far as making it easier for internationalization, neither really makes it any more difficult or easier

taliaedX (approved):
I have really minor quibbles around clean up not worth blocking on. This otherwise looks good to me.

src/components/CouponDetails/ActionButton.jsx
raq929: I debated it, but I think it's valuable to have a quick and easy alert. I could define it as an array, but then you'd have to add elements to that array every time you add a type.

src/components/CouponDetails/constants.js
raq929: Yeah, I originally defined them directly, and then when I realized I wasn't going to get rid of `COUPON_FILTER_TYPES` entirely, I changed it to this.

src/components/CouponDetails/ActionButton.jsx
taliaedX: Ah I didn't think of that. Good point!

PR #581, opened 2021-06-23
fix: nudge reminder checkbox is not checked by default

sameenfatima78:
This PR fixes a frontend bug regarding checkbox not being checked by default even though the value being passed on backend is true. More details on the ticket: https://openedx.atlassian.net/browse/ENT-4569

### DIFF
--- a/src/components/ReduxFormCheckbox/index.jsx
+++ b/src/components/ReduxFormCheckbox/index.jsx
@@ -8,6 +8,7 @@ const ReduxFormCheckbox = (props) => {
     label,
     helptext,
     input,
+    defaultChecked,
   } = props;
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/ReduxFormCheckbox/index.jsx
```diff
@@ -28,6 +30,7 @@ const ReduxFormCheckbox = (props) => {
 
 ReduxFormCheckbox.defaultProps = {
   helptext: null,
+  defaultChecked: false,
```
saleem-latif, line 33: shouldn't this be `true` as well?

saleem-latif (approved):
Just on comment, otherwise looks good 👍🏼

src/components/ReduxFormCheckbox/index.jsx
sameenfatima78: No this component is being used in Revoke modal as well which requires the default state to be unchecked.

raq929 (approved):
Looks good to me; just double check that it's not used in other places where it needs to be checked by default. I think you caught the only one but I'm not 100% sure.

zamanafzal (approved):

PR #587, opened 2021-07-01
fix: coupon details styles

raq929:
Fixes styles that were broken by updating Paragon components.
Due to a breaking change in Paragon, I removed the BulkEnrollment enroll button (which was slated for removal anyway)

Before:
Screen Shot 2021-06-30 at 10 05 21 AM

After:
Screen Shot 2021-06-30 at 10 05 01 AM


src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -47,30 +42,6 @@ const selectColumn = {
   disableSortBy: true,
 };
 
-export const EnrollButton = ({ row, goToNextStep, dispatch }) => {
```
binodpant, line 50: thanks for cleaning this out!

binodpant (approved):
styles look good, removal of enroll button too! I culd not verify locally yet due to devstack, we can verify in stage

PR #589, opened 2021-07-02
Move bulk action functionality to bulk action component

raq929:
Component is responsible for it's value and options instead of the parent
Allows for easier testing and debugging

There are no visual changes. This fixes a bug I introduced where when you were clicking back to "Remind" from "Redeemed", the value of the BulkAction select did not update as it ought to.

### DIFF
--- a/src/components/CouponDetails/CouponBulkActions.jsx
+++ b/src/components/CouponDetails/CouponBulkActions.jsx
@@ -1,56 +1,75 @@
-import React from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
 import {
   Button, Form,
 } from '@edx/paragon';
-import { BULK_ACTION } from './constants';
+import { BULK_ACTION, COUPON_FILTERS, COUPON_FILTER_TYPES } from './constants';
+import { getBASelectOptions, getFirstNonDisabledOption } from './helpers';
 
 const CouponBulkActions = ({
-  value, options, disabled, onChange, handleBulkAction,
-}) => (
-  <>
-    <Form.Group className="flex-grow-1" controlId={BULK_ACTION.controlId} name={BULK_ACTION.name}>
-      <Form.Control
-        className="float-right"
-        floatingLabel={BULK_ACTION.label}
-        as="select"
-        value={value}
-        onChange={(e) => onChange(e.target.value)}
-        disabled={disabled}
+  handleBulkAction, selectedToggle, numSelectedCodes, numUnassignedCodes, couponAvailable, isTableLoading, hasTableData,
+}) => {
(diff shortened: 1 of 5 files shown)


src/components/CouponDetails/helpers.test.jsx
```diff
@@ -36,6 +36,7 @@ describe('getFirstNonDisabledOption', () => {
     [[{ value: 'foo', disabled: true }, { value: 'bar' }], 'bar'],
     [[{ value: 'foo' }, { value: 'bar', disabled: false }, { value: 'baz', disabled: true }], 'foo'],
     [[{ value: 'foo', disabled: true }, { value: 'bar', disabled: true }, { value: 'baz', disabled: false }], 'baz'],
+    [[{ value: 'foo', disabled: false }, { value: 'bar', disabled: true }, { value: 'baz', disabled: true }], 'foo'],
```
raq929, line 39: I just wanted to be extra sure that the logic issue wasn't here, so added a test case that was the same as what I was seeing on stage.

src/components/CouponDetails/CouponBulkActions.jsx
```diff
@@ -1,56 +1,75 @@
...
+    isRedeemedView,
+    hasTableData,
+    couponAvailable,
+    numUnassignedCodes,
+    numSelectedCodes,
+  }), [numUnassignedCodes, numSelectedCodes, selectedToggle, hasTableData, couponAvailable]);
+
+  const [value, setValue] = useState(getFirstNonDisabledOption(options));
+  useEffect(() => {
+    setValue(getFirstNonDisabledOption(options));
```
taliaedX, line 27: is there a potential that between the declaration of useState with the default value of `getFirstNonDisabledOption(options)` the value of that resolved method would change by this line? Ie why do we set the value to the same thing the we're defaulting to?

taliaedX (approved):
One question and I'm not entirely sure I get why shifting the logic from parent to action class fixed the bug. I believe that it likely did, but would love to _understand_

src/components/CouponDetails/CouponBulkActions.jsx
raq929: This is actually the part that fixes the problem; it wasn't there in the parent component and would have been difficult to add for various reasons.

Basically this is saying, any time the `BulkActionSelectOptions` change, you should update the value here as well. Gonna make a small change to make that more obvious and keep them in sync.

One of the reasons for moving the code is that it was very hard to reason about in the parent component. It was split between three or four different methods across an 800+ line component. This is work I wanted to do in the refactor anyway but ran out of time for.

raq929: And really, one of the main reasons is that it made it easier to test that my solution worked. I *could* test it solely via the parent component, but I have a lot more control over the inputs this way and I find it much easier to reason about.

PR #590, opened 2021-07-07
feat: Remove the word 'All' from filter status to avoid confusion ENT-4684

binodpant:
Will post screenshot but my local devstack is hosed, again
May have to reboot and/or start over

ENT-4684


binodpant:
To reviewer, I don't think we need to block on a screenshot here I can get also get one from staging, pretty sure this will work, it's just removing the All word which was there conditionally

alex-sheehan-edx (approved):
LGTM

PR #591, opened 2021-07-07
build: upgrade Paragon and other dependencies

adamstankiewicz:
1. Reverts `@edx/brand` back to using `brand-openedx` by default. The `brand-edx.org` theme is applied at build time in GoCD.
2. Upgrades `@edx/paragon` to latest in an attempt to resolve the following style issue with our select inputs:

image

We had previously thought it was something to do with the `@edx/brand-edx.org` theme not properly being applied but since the select icon is correct on the [Open edX theme Paragon documentation site](https://edx.github.io/paragon/components/form/form-control/#input-types), this seems to be a red herring. It should be looking like this:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -189,9 +189,9 @@
       },
       "dependencies": {
         "debug": {
-          "version": "4.3.1",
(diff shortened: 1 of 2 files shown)


adamstankiewicz:
No longer relevant. Closing.

PR #593, opened 2021-07-07
style: Bulk Enrollment: Review Step: Add title & aria label variables

taliaedX:
#### Background
We want to add hover text for the delete icon on the Review Step of the Bulk Enrollment Stepper.
I also wanted to make the Aria label text more descriptive to benefit screenreaders better, so I added prop/parameters to differentiate course and learner ReviewItems at least in the context of Aria & title texts.

#### Specific Changes
- removal, altText props
  - Update existing `alt=` field on IconButton to use the altText prop.
  - Add `title=` attribute to IconButton (it accepts attr passthroughs)
- Add a test for the title (while aria labels show in the dom tree, they aren't actually visible so react testing library's docs state they aren't accessible to screen, but I at least wanted to verify one of these was showing properly and thus the prop passing was working as intended)

**Note:** There are no before/after screenshots, because I cannot pause the hover text on the delete button to showcase this. 😞  But here's a screenshot at least explaining which buttons we're affecting here:
Screen Shot 2021-07-07 at 4 18 34 PM

### DIFF
--- a/src/components/BulkEnrollmentPage/stepper/ReviewList.test.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/ReviewList.test.jsx
@@ -27,6 +27,7 @@ const defaultProps = {
     singular: 'wug',
     plural: 'wugs',
     title: 'Wugs',
+    removal: 'Remove Wugs. Return to egg.',
   },
   returnToSelection: jest.fn(),
 };
@@ -52,6 +53,10 @@ describe('ReviewList', () => {
   it('displays a title', () => {
     render(<ReviewList {...defaultProps} />);
     expect(screen.getByText(defaultProps.subject.title)).toBeInTheDocument();
+  });
+  it('renders delete button hover title', () => {
+    render(<ReviewList {...defaultProps} />);
+    expect(screen.getAllByTitle(defaultProps.subject.removal)[0]).toBeInTheDocument();
   });
   it('displays the number of rows selected', () => {
     render(<ReviewList {...defaultProps} />);
--- a/src/components/BulkEnrollmentPage/stepper/ReviewItem.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/ReviewItem.jsx
@@ -7,7 +7,9 @@ import PropTypes from 'prop-types';
 
 import { deleteSelectedRowAction } from '../data/actions';
 
-const ReviewItem = ({ row, accessor, dispatch }) => {
+const ReviewItem = ({
+  row, accessor, dispatch, altText,
+}) => {
   const onClick = () => {
     dispatch(deleteSelectedRowAction(row.id));
   };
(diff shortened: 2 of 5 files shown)


src/components/BulkEnrollmentPage/stepper/ReviewList.test.jsx
```diff
@@ -53,6 +54,10 @@ describe('ReviewList', () => {
     render(<ReviewList {...defaultProps} />);
     expect(screen.getByText(defaultProps.subject.title)).toBeInTheDocument();
   });
+  it('renders delete button hover title', () => {
+    render(<ReviewList {...defaultProps} />);
+    expect(screen.getAllByTitle(defaultProps.subject.removal)[0]).toBeInTheDocument();
```
binodpant, line 59: as a quick a11y coverage wondering if also getting the button by aria-label if possible also is a good idea? and even by title?

binodpant (approved):
lgtm, made a testing suggestion to cover getting item by title and/or aria-label for better a11y coverage

src/components/BulkEnrollmentPage/stepper/ReviewList.test.jsx
binodpant: scratch the 'title' part I see you already do it

taliaedX: Not sure if I'm misunderstanding, but it sounds like you're asking for a test for the delete button's title and aria text, which:
A. this test you're commenting on is for the title (which is why I'm wondering if I'm misunderstanding)
B. I mention in the PR description that react testing library doesn't really offer a way to actually test that the aria label appears to the user. I can maybe do a unit test, but that seems a bit redundant since 1. the mechanism that is passing in this text is being tested already by getting the title text here and 2. if we're just testing that alt renders as aria-label in the document, we're just testing the IconButton component's functionality, which is already done in the IconButton's own test in Paragon.

binodpant: Yeah I agree. I think testing by title covers our based for a11y for the title you just added. I looked at https://testing-library.com/docs/queries/bylabeltext for clues.. it seems aria-label may be covered by ByLabelText but I would not worry about it in this codebase since it's part of the IconButton features.

I was thinking if there a way for us to capture that we are passing the correct 'alt' value as well

binodpant: Basically a test case for (somehow) `Update existing alt= field on IconButton to use the altText prop.`

taliaedX: I see. This proposed test would cover a case where someone changes the name of the variable and misses updating it on the alt field. Doing so doesn't cause any error, just a prop warning currently for the IconButton component (and we all ignore those). I find the likelihood of such a change very very small, but I can test for it with not too much additional weight besides having to use enzyme again and isn't quite the same as what the IconButton test in Paragon is testing, so it's not actually redundant. I will commit that in a sec.

I just very much want to avoid redundant tests. Eventually all the tests we add will actually impact build performance as well as container size. Each additional test on its own is of course very very minor, but it eventually adds up. I've been in that environment, so I don't want to go back to that state of things lol.

PR #596, opened 2021-07-16
docs: PR template for Paragon work

binodpant:
This PR template will be handy when doing work that involves Paragon since it involves multiple steps, coding and communications wise

it's draft because I have not really through through this much and putting this out there for super early team review if helpful, links if applicable need to be acquired/added to the various checklist steps

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/paragon_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/paragon_template.md
@@ -1,0 +1,24 @@
+### Make Paragon changes and test locally
+
+* [ ] Create Paragon ticket on PAR board, and link it to the ticket in your team board.
+* [ ] Check Paragon development docs on how to add and test a component locally. This step does not need module.config.js file yet.
+* [ ] Share screenshots/UX changes from local Paragon showcase page to UX or Product before committing full changes to ensure Acceptance Criteria met
+* [ ] Mobile friendly? Tested on smaller screens for responsiveness
+* [ ] a11y friendly. Run one of the a11y tools against the UI (axe, ARC toolkit are some options)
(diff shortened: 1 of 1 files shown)


taliaedX (commented):
Wait, what is this PR template for? Is this just for when you have already made Paragon changes, deployed them, etc. and now you're merging in the new version in the admin portal? Feels like it's way too late in the process for pretty much all of these steps, no?

Even if this was in the Paragon repo, I also feel like the creating a ticket and adding component bullet points are unnecessary here. I'm also not a fan of release steps in PRs because I don't really look at PRs for release instructions. It's done and merged. I know we do that for edx-enterprise though so it's not uncommon and maybe most people do actually find it useful (judging by how often any of us check off the boxes in our prs I'm gonna guess no though :p). I don't know that we have a precedent of deployment documentation in Confluence or elsewhere though (we might. I swear I've seen something...) but if we did that'd make more sense to me as a location for release instructions.

I think the bits about screenshots, mobile and ally checks is great and should be checkboxes, but they should be in the Paragon repo (actually I also think we should have them in the admin portal. I often forget all fo those things). If the component isn't mobile or ally friendly by the time it's released in Paragon, that's a huge problem.

Just my thoughts though. I appreciate the effort! 🤷

binodpant:
Thanks @taliaedX . Yeah the location of this doc looks like it can be confusing. It's not specifically admin portal related. And yes arguably this can actually just be a Paragon PR template, but then not all Paragon work involves some of these steps. So it could also be confusing if we put it in Paragon. Also, Paragon already has pretty good docs on dev and release so don't think we need to add more there.

An alternative is: technically we can just have this content be in confluence even, and just link from the PR template to it? And one PR template for each repo can just be identical? Sth like, if you need Paragon work, here is a link with a guide. Plus osme of these checkboxes here. I think even the Paragon steps are relevant when you are doing work in this portal, but they are perhaps also superflous to what is already in Paragon

PR #598, opened 2021-07-19
fix: remove 'Filtered By' message

kiram15:
Set boolean of "showFilteredFields" of FilterStatus to false to remove "Filtered by userEmail" text. Screenshots in [JIRA](https://openedx.atlassian.net/browse/ENT-4677) ticket.


src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
```diff
@@ -82,8 +82,6 @@ describe('AddLearnersStep', () => {
     expect(await screen.findByText('Search Email')).toBeInTheDocument();
     await act(() => mockTableData);
     userEvent.type(screen.getByPlaceholderText('Search Email'), 'beAR');
-    expect(await screen.findByText(/Filtered by userEmail/)).toBeInTheDocument();
```
binodpant, line 85: you can actually enhance this assertoin by checking specifically that the field does not exist in the dom, sth like 

```
expect(await screen.queryByText(/Filtered by userEmail/)).not.toBeInTheDocument();
```

Check the docs for React testing library I think you need to use queryBy instead of findBy for testing absence ... not 100%

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
```diff
@@ -82,8 +82,6 @@ describe('AddLearnersStep', () => {
...
-    expect(await screen.findByText('Clear Filters')).toBeInTheDocument();
```
binodpant, line 86: same here ... try and check for absence ...

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.test.jsx
binodpant: https://testing-library.com/docs/guide-disappearance/#asserting-elements-are-not-present

binodpant:
made a test suggestion

also this will need a paragon upgrade once that is ready but otherwise looking great

package-lock.json
```diff
@@ -4051,9 +4063,9 @@
       }
     },
     "@types/react-transition-group": {
-      "version": "4.4.1",
-      "resolved": "https://registry.npmjs.org/@types/react-transition-group/-/react-transition-group-4.4.1.tgz",
-      "integrity": "sha512-vIo69qKKcYoJ8wKCJjwSgCTM+z3chw3g18dkrDfVX665tMH7tmbDxEAnPdey4gTlwZz5QuHGzd+hul0OVZDqqQ==",
+      "version": "4.4.2",
```
edx-abolger, line 4066: Can you please revert the changes to the package lock that are not nested under "@edx/paragon" ?  In general, these automated packages could affect something else that isn't part of your change here.  Also, are you sure you need the paragon upgrade to submit your changes?

binodpant: the paragon changes (and potentially any needed package-lock changes thus generated by the new dep, will be needed for this PR

binodpant: but yeah package lock changes in the PR currently seem to be causing issues so they need an update

binodpant: typically running locallly `npm ci` and checking in that latest package lock will be best

src/components/BulkEnrollmentPage/stepper/AddLearnersStep.jsx
```diff
@@ -116,6 +116,8 @@ const AddLearnersStep = ({
     getRowId: (row, relativeIndex, parent) => row?.uuid || (parent ? [parent.id, relativeIndex].join('.') : relativeIndex),
   }), []);
 
+  const filterStatusComponent = () => <DataTable.FilterStatus showFilteredFields={false} />;
```
edx-abolger, line 119: In general, I would move this to the top of the file with the other const components and follow the naming convention in the file already: since it returns a component, I think "const filterStatusComponent" should be "const FilterStatus"

binodpant:
As per chat offline, 

> I also think we can completely remove this from the test

```
afterEach(() => {
  // cleanup on exiting
  unmountComponentAtNode(container);
  container.remove();
  container = null;
});
```
and the use of the 'react-dom' utils in the tests, since react testing library takes care of all this .. plus the import of this function at the top ...

src/components/subscriptions/tests/expiration/SubscriptionExpirationBanner.test.jsx
```diff
@@ -51,8 +51,9 @@ describe('<SubscriptionExpirationBanner />', () => {
     };
     render(<ExpirationBannerWithContext detailState={detailStateCopy} />);
     expect(screen.queryByRole('alert')).not.toBeNull();
+    const elem = await screen.findByText('Dismiss');
```
binodpant, line 54: good! I always try to use the findBy instead of getby for these if I suspect there is going to be a timing issue with finding the element

binodpant (approved):
lgtm!

PR #600, opened 2021-07-29
Adds event tracking for licensee expiration actions in modal and alert

manny-m:
#### Context
When a subscription plan nears its expiration date, admin users are presented warnings in the form of modals and alerts, (some of) which have actions to dismiss them or contact customer support. Example:

#### Events
Event names:

- `edx.admin_portal.subscriptions.expiration.modal.dismissed`
- `edx.admin_portal.subscriptions.expiration.modal.support_cta.clicked`
- `edx.admin_portal.subscriptions.expiration.alert.dismissed`
- `edx.admin_portal.subscriptions.expiration.alert.support_cta.clicked`

Event Properties:
- `expiration_threshold`: 30/60/120
- `days_until_expiration`: \<some_integer\>

### DIFF
--- a/src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
+++ b/src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
@@ -1,6 +1,7 @@
 import React, { useContext, useState } from 'react';
 import { Alert } from '@edx/paragon';
 import PropTypes from 'prop-types';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 
 import {
   SUBSCRIPTION_DAYS_REMAINING_MODERATE,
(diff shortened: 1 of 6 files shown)


macdiesel:
FYI you should use a direct ticket link instead of the modal detail view issue link you provided as that link will die after the current sprint is completed. Direct link: https://openedx.atlassian.net/browse/ENT-4633

macdiesel:
I would like to suggest a slight tweak on the event names:

- edx.ui.admin_portal.subscriptions.expiration.modal.dismissed
- edx.ui.admin_portal.subscriptions.expiration.modal.support_cta.clicked
- edx.ui.admin_portal.subscriptions.expiration.alert.dismissed
- edx.ui.admin_portal.subscriptions.expiration.alert.support_cta.clicked

RE: Event properties:
within the context of the UI do you have access to the admin ID, org ID and other information that would allow us to link back to the users who saw this data? If so could you please include that info as well?

manny-m:
@macdiesel 
> RE: Event properties:
> within the context of the UI do you have access to the admin ID, org ID and other information that would allow us to link back to the users who saw this data? If so could you please include that info as well?

Events do seem to include the user ID, enterprise slug, and more. I've included an example below. You can see the userId is included as well as the page path which has the enterprise slug.

Here is what a raw event looks like (redacted some fields <placeholder>):
```
{
  "anonymousId":  <uuid>,
  "context": {
    "ip": <ip>,
    "library": {
      "name": "analytics.js",
      "version": "4.1.8"
    },
    "locale": "en-US",
    "page": {
      "path": "/test-enterprise/admin/subscriptions",
      "referrer": "",
      "search": "",
      "title": "Subscription Management - edX Admin Portal",
      "url": "http://localhost:1991/test-enterprise/admin/subscriptions"
    },
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.164 Safari/537.36"
  },
  "event": "edx.admin_portal.subscriptions.expiration.modal.support_cta.clicked",
  "integrations": {},
  "messageId": "<some_id>",
  "originalTimestamp": "2021-07-30T17:56:47.779Z",
  "properties": {
    "days_until_expiration": -2,
    "expiration_threshold": 30
  },
  "receivedAt": "2021-07-30T17:56:47.742Z",
  "sentAt": "2021-07-30T17:56:47.785Z",
  "timestamp": "2021-07-30T17:56:47.736Z",
  "type": "track",
  "userId": <int>
}
```

src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
```diff
@@ -70,28 +71,44 @@ const SubscriptionExpirationBanner = ({ isSubscriptionPlanDetails }) => {
...
+    subscriptionExpirationThreshold = SUBSCRIPTION_DAYS_REMAINING_SEVERE;
     alertType = 'warning';
   }
   if (daysUntilExpiration <= SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL) {
+    subscriptionExpirationThreshold = SUBSCRIPTION_DAYS_REMAINING_EXCEPTIONAL;
     dismissible = false;
     alertType = 'danger';
   }
 
+  const emitBannerActionEvent = (dismiss) => {
```
adamstankiewicz, line 87: [nit/suggestion] It might be worth splitting this function into two: `emitAlertActionEvent` and `emitAlertDismissedEvent` so it's a bit more readable at a glance without having to decipher the `true/false` flag.

src/components/subscriptions/expiration/SubscriptionExpirationModals.jsx
```diff
@@ -90,17 +91,32 @@ const SubscriptionExpirationModals = ({ enterpriseId }) => {
     }
   }, [isSubscriptionExpired]);
 
+  const emitExpirationModalActionEvent = (dismiss) => {
```
adamstankiewicz, line 94: [nit/suggestion] Similar as above. It might be worth splitting this function into two: `emitModalActionEvent` and `emitModalDismissedEvent` so it's a bit more readable at a glance without having to decipher the `true/false` flag.

adamstankiewicz (approved):
LGTM. Two small non-blocking nits/suggestions.

Per Brian's recommendation, though, we probably want "ui" into the event names. I believe Brian's recommendation is to distinguish between a frontend event vs. backend/server event.

> edx.**_ui_**.admin_portal.*

macdiesel:
Sorry for the late followup on this, was on vacation. I think this is a good start. I've filed another ticket here so we can round out this work: https://openedx.atlassian.net/browse/ENT-4818

PR #601, opened 2021-08-02
ENT4740 Added title for Admin Notification

zamanafzal:
Jira Ticket:
https://openedx.atlassian.net/browse/ENT-4740

Backend PR:
https://github.com/edx/edx-enterprise/pull/1318

Made notification title customizable.

### DIFF
--- a/src/components/FeatureAnnouncementBanner/index.jsx
+++ b/src/components/FeatureAnnouncementBanner/index.jsx
@@ -7,6 +7,7 @@ import LmsApiService from '../../data/services/LmsApiService';
 
 const FeatureAnnouncementBanner = ({ enterpriseSlug }) => {
   const [enterpriseNotificationBanner, setEnterpriseNotificationBanner] = useState({
+    title: null,
     text: null,
     notificationId: null,
   });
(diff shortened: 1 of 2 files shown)


src/components/FeatureAnnouncementBanner/index.jsx
```diff
@@ -37,10 +39,10 @@ const FeatureAnnouncementBanner = ({ enterpriseSlug }) => {
   }, [enterpriseSlug]);
 
   return (
-    isRead || !enterpriseNotificationBanner.text ? null : (
+    isRead || !enterpriseNotificationBanner.text || !enterpriseNotificationBanner.title ? null : (
```
sameenfatima78, line 42: Can you please explain why do we need to add a separate condition for `title` here, wouldn't `!enterpriseNotificationBanner.text` enough?

sameenfatima78 (approved):
Just one minor question, otherwise LGTM

src/components/FeatureAnnouncementBanner/index.jsx
zamanafzal: Just to apply the same validations of the backend. It will work fine without this check but I added it to handle issues if something changed in the future at the backend then we should have the same validation on the frontend.

PR #603, opened 2021-08-11
feat: Add Scrollable component around stepper contents. ENT-4695

binodpant:
Why? To improve the scrollable area UX and make it more apparent the contents are scrollable. Also the shadows look cool!

ENT-4695

DEPENDS ON: https://github.com/edx/paragon/pull/779 ✅ 
ALSO SUBJECT TO: Review from Jen G ✅ 
FIXED! ☑️  Looks like I may need additional margin for the search header 💹 

Tests: The `Scrollable` component will already be tested in Paragon, there is nothing to be done on this side here that adds value.

CURRENT: check sth like https://portal.stage.edx.org/binod-subs/admin/enrollment/a4db1210-93fa-4ffc-af5d-62110484157a

AFTER:

https://user-images.githubusercontent.com/528166/129810541-660d4c96-a6ab-4551-b8b8-4f303aa1131e.mov

### DIFF
--- a/src/components/BulkEnrollmentPage/stepper/AddCoursesStep.jsx
+++ b/src/components/BulkEnrollmentPage/stepper/AddCoursesStep.jsx
@@ -20,23 +20,25 @@ const AddCoursesStep = ({
   <>
     <p>{ADD_COURSE_DESCRIPTION}</p>
     <h2>{ADD_COURSES_TITLE}</h2>
-    <SearchData>
-      <InstantSearch
-        indexName={configuration.ALGOLIA.INDEX_NAME}
-        searchClient={searchClient}
-      >
-        <Configure
-          filters={`enterprise_catalog_uuids:${subscription.enterpriseCatalogUuid}`}
-          hitsPerPage={25}
-        />
-        <SearchHeader variant="default" />
-        <CourseSearchResults
-          enterpriseId={enterpriseId}
-          enterpriseSlug={enterpriseSlug}
-          subscriptionUUID={subscription.uuid}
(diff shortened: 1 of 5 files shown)


src/components/BulkEnrollmentPage/stepper/AddCoursesStep.jsx
```diff
@@ -20,23 +20,25 @@ const AddCoursesStep = ({
...
-        />
-        <SearchHeader variant="default" />
-        <CourseSearchResults
-          enterpriseId={enterpriseId}
-          enterpriseSlug={enterpriseSlug}
-          subscriptionUUID={subscription.uuid}
-        />
-      </InstantSearch>
-    </SearchData>
+    <div className="add-courses-search-data">
```
adamstankiewicz, line 23: nit: Could we use a margin utility class here to accomplish the same thing (e.g., `ml-1.5`)?

adamstankiewicz (approved):
LGTM, with a nit about using a CSS utility class for spacing.

src/components/BulkEnrollmentPage/stepper/AddCoursesStep.jsx
binodpant: yeah given that is the only thing I am doing with the class let me try that !

binodpant: well looks like I don't even need this class or styling anymore since the margins are now auto handled by scrollable (so this looks good even without it!)

PR #604, opened 2021-08-13
feat: the invite learners button should render in the plan detail page

iloveagent57:
...when there are no allocated licenses but there are revoked licenses.
ENT-4804

This also fixes a defect I noticed, where "Invite Learners" wasn't rendering in the All Users tab after inviting a learner - you'd have to do a full page refresh to see it.  I fixed this by adding a `forceRefresh()` function to the `SubscriptionDetailContext` context, which is now called by the `InviteLearnersButton.onSuccess()` handler (in addition to the forceRefresh of the subscription context for the entire agreement).

Testing:
* Create a new subscription plan
* Assign and then revoke a single license.
* On the plan details page, the "Invite Learners" button should still render at the top of the page when you're on the "All Users" tab (this is the fix).


adamstankiewicz (approved):
LGTM 👍

PR #605, opened 2021-08-13
chore(admin_portal): :wastebasket: Remove deprecated Fullstory tooling and library from Admin Portal for ENT-3870

edx-abolger:
Removing unused tool configuration.


binodpant (approved):
scratch prior comment: saw edx-internal changes already. lgtm,

PR #607, opened 2021-08-17
Subscription Management now using customer-agreement api

manny-m:
### [Closes ENT-4631](https://openedx.atlassian.net/browse/ENT-4631)

The purpose of this change is to make our subscription expiration modal/alert reminders to use a new field `netDaysUntilExpiration` coming from `/api/v1/customer-agreement/`, to better alert our users. 

Users will no longer be alerted if they have an upcoming/pending  subscription. They will only be notified about expiration when when the the last expiration date is within the expiration notification threshold (120 days).

Screenshot below show two plans, named `expired sub plan` and `expires later`. 

Screen Shot 2021-08-17 at 4 57 07 PM
 
If we navigate to `expired sub plan` detail view you will see no alert is being showed because we still have an ongoing plan `expires later`.
Screen Shot 2021-08-17 at 4 53 36 PM

If both plans were expired or close to expiration however, a warning would be shown.

### DIFF
--- a/src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
+++ b/src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
@@ -14,12 +14,12 @@ import ContactCustomerSupportButton from '../buttons/ContactCustomerSupportButto
 
 const SubscriptionExpirationBanner = ({ isSubscriptionPlanDetails }) => {
   const {
-    subscription: { daysUntilExpiration, expirationDate, showExpirationNotifications },
+    subscription: { agreementNetDaysUntilExpiration, expirationDate, showExpirationNotifications },
   } = useContext(SubscriptionDetailContext);
   const [showBanner, setShowBanner] = useState(true);
 
   const renderMessage = () => {
-    const subscriptionExpired = daysUntilExpiration <= 0;
+    const subscriptionExpired = agreementNetDaysUntilExpiration <= 0;
     // use subscription detail view messaging
     if (isSubscriptionPlanDetails) {
       return (
@@ -59,7 +59,7 @@ const SubscriptionExpirationBanner = ({ isSubscriptionPlanDetails }) => {
             <Alert.Heading>
               Your subscription contract is expiring soon
             </Alert.Heading>
-            Your current subscription contract will expire in {daysUntilExpiration} days.
+            Your current subscription contract will expire in {agreementNetDaysUntilExpiration} days.
             Renew your subscription today to minimize access disruption for your learners.
           </>
         )}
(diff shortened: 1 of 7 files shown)


src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
```diff
@@ -14,12 +14,12 @@ import ContactCustomerSupportButton from '../buttons/ContactCustomerSupportButto
...
 const SubscriptionExpirationBanner = ({ isSubscriptionPlanDetails }) => {
   const {
-    subscription: { daysUntilExpiration, expirationDate, showExpirationNotifications },
+    subscription: { agreementNetDaysUntilExpiration, expirationDate, showExpirationNotifications },
   } = useContext(SubscriptionDetailContext);
   const [showBanner, setShowBanner] = useState(true);
 
   const renderMessage = () => {
-    const subscriptionExpired = daysUntilExpiration <= 0;
+    const subscriptionExpired = agreementNetDaysUntilExpiration <= 0;
```
iloveagent57, line 22: I think we need to keep using `daysUntilExpiration` in the case where we're looking at the subscription plan details page (when `isSubscriptionPlanDetails` is true).  Otherwise, we won't see the expiration banner on individual plans when the individual plan _has_ expired but there are other plans in the agreement that have _not_ expired.

src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
```diff
@@ -59,26 +59,26 @@ const SubscriptionExpirationBanner = ({ isSubscriptionPlanDetails }) => {
...
+            Your current subscription contract will expire in {agreementNetDaysUntilExpiration} days.
             Renew your subscription today to minimize access disruption for your learners.
           </>
         )}
       </>
     );
   };
 
-  if (daysUntilExpiration > SUBSCRIPTION_DAYS_REMAINING_MODERATE) {
+  if (agreementNetDaysUntilExpiration > SUBSCRIPTION_DAYS_REMAINING_MODERATE) {
```
iloveagent57, line 70: The prop to use down here also depends on whether we're on the plan details page or the agreement "parent" page.  Maybe something like this toward the top of the component would be helpful:
{noformat}
const planDuration = isSubscriptionPlanDetails ? daysUntilExpiration : agreementNetDaysUntilExpiration;
{noformat}
And then you can use `planDuration` in all of the places we were previously using `daysUntilExpiration`.

Let me know if I'm completely off and missing something obvious here.

manny-m: This makes sense to me but maybe I'm unclear on the intended UX. Should we be passing messaging to our users that something is expiring/expired when they still have an on going subscription plan?

Example: 
User has 2 subscription plans, plan A (expired) and plan B (expires past the expiration warning threshold, +120 days)
When in the subscription plan detail view for plan A should we have messaging regarding its expiration?

src/components/subscriptions/expiration/SubscriptionExpirationBanner.jsx
iloveagent57: Actually, it looks from the AC of the ticket like we actually don't want what I'm saying, so ignore.

macdiesel (approved):

PR #618, opened 2021-09-15
fix: revoke emails should also redirect to learner portal

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-4886

Actual: Revoke emails do not redirect to learner portal if learner portal is enabled in enterprise configuration.
Expected: It should redirect to learner portal if learner portal is enabled, otherwise to B2B search page.

### DIFF
--- a/src/components/CodeRevokeModal/index.jsx
+++ b/src/components/CodeRevokeModal/index.jsx
@@ -18,7 +18,7 @@ import {
 } from '../CodeModal';
 import EmailTemplateForm from '../EmailTemplateForm';
 import CheckboxWithTooltip from '../ReduxFormCheckbox/CheckboxWithTooltip';
-import { features } from '../../config';
+import { configuration, features } from '../../config';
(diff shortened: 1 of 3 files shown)


src/components/CodeRevokeModal/index.jsx
```diff
@@ -118,6 +120,11 @@ class CodeRevokeModal extends React.Component {
       do_not_email: doNotEmail,
     };
 
+    // If the enterprise has a learner portal, we should direct users to it in our assignment email
```
muhammad-ammar, line 123: typo: `assignment` -> `revoke`?

muhammad-ammar (approved):
LGTM. Just one minor comment.

src/components/CodeRevokeModal/index.jsx
sameenfatima78: fixed

PR #619, opened 2021-09-17
chore: use bulk-revoke endpoint to revoke licenses

long74100:
Deprecating the /revoke and moving towards /bulk-revoke to remove redundancy. New endpoint will be used for the new table as well.

https://openedx.atlassian.net/browse/ENT-4737


manny-m (commented):
LGTM 👍

manny-m (approved):

PR #623, opened 2021-09-28
fix: change boolean check for conditional enabling file attachment

muneebGH:


.env.development
```diff
@@ -32,4 +32,4 @@ FEATURE_BULK_ENROLLMENT=true
 HOTJAR_APP_ID=''
 HOTJAR_VERSION=6
 HOTJAR_DEBUG=''
-FEATURE_FILE_ATTACHMENT=false
+FEATURE_FILE_ATTACHMENT=true
```
sameenfatima78, line 35: Are we enabling this feature for production too?

sameenfatima78 (approved):
Just one question, otherwise good to go

.env.development
muneebGH: not yet, once i test on stage, i will create edx-internal PR to enable it on production.

sameenfatima78: Sounds good

PR #625, opened 2021-10-04
Integrating DataTable for bulk actions 

manny-m:
#### ⭐ Changes Summary:

- Replaces old table component with [Paragon DataTable](https://paragon-edx.netlify.app/components/datatable/) component 
- Added bulk action functions for revoking and reminding learners
- Rebuilt revoke and remind modals to support multiple users

#### 🚧 WIP & Issues:

- [ ] New remind modal not using `getSubscriptionContactText` in closing portion
- [ ] Modal error state copy
- [ ] SelectAll button on toolbar not working as intended
- [ ] There are no loading states


long74100 (commented):
a few nits, looks good 🔥 !

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementRevokeModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+          {title}
+        </ModalDialog.Title>
+      </ModalDialog.Header>
+      <ModalDialog.Body>
+        {requestState.error
+            && (
+            <Alert variant="danger">
+              <p>There was an error with your request. Please try again.</p>
+              <p>
+                If the error persists,&nbsp;
```
adamstankiewicz, line 104: nit: `{' '}` over `&nbsp;`

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableActionColumn.jsx
```diff
@@ -0,0 +1,83 @@
...
+          <Tooltip id="tooltip-revoke">
+            {revokeText}
+          </Tooltip>
+        )}
+      >
+        <IconButton
+          alt={revokeText}
+          title={revokeText}
+          src={RemoveCircle}
+          style={{ marginLeft: displayRemind ? 0 : 44 }}
```
adamstankiewicz, line 63: nit: Could we use our CSS spacing utility classes here (e.g., `.ml-5`)

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+    selectedUsers.forEach(user => {
+      const userStatus = user.status;
+      if (canRemindLicense(userStatus)) {
+        tempRemind.push(user);
+      }
+      if (canRevokeLicense(userStatus)) {
+        tempRevoke.push(user);
+      }
+    });
```
adamstankiewicz, line 31: nit: Might be a bit cleaner to use `.filter` here? E.g. 

```
const tempRemind = selectedUsers.filter(user => canRemindLicense(user.status));
const tempRevoke = selectedUsers.filter(user => canRevokeLicense(user.status));
```

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementUserBadge.jsx
```diff
@@ -0,0 +1,20 @@
...
+import {
+  Badge,
+} from '@edx/paragon';
+
+import { USER_BADGE_MAP } from '../../data/constants';
+
+const LicenseManagementUserBadge = ({ userStatus }) => {
+  const badgeLabel = USER_BADGE_MAP[userStatus]
+    ? USER_BADGE_MAP[userStatus]
+    : USER_BADGE_MAP.UNDEFINED;
```
adamstankiewicz, line 12: I think I commented this above, but if there isn't a valid `userStatus`, let's not render a badge at all (empty cell or "-") so it's not misleading.

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+      return `Activated: ${formatTimestamp({ timestamp: user.activationDate })}`;
+    }
+    case REVOKED: {
+      return `Revoked: ${formatTimestamp({ timestamp: user.revokedDate })}`;
+    }
+    case ASSIGNED: {
+      return `Invited: ${formatTimestamp({ timestamp: user.lastRemindDate })}`;
+    }
+    default: {
+      return 'No Data';
```
adamstankiewicz, line 42: Let's render an empty cell or "-" when there's nothing to display.

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+  );
+
+  const debouncedSetCurrentPage = debounce(
+    setCurrentPage,
+    DEBOUNCE_TIME_MILLIS,
+  );
+  // Call back function, handles filters and page changes
+  const fetchData = useCallback(
+    (args) => {
+      // pages index from 1 in backend, Datable component index from 0
```
adamstankiewicz, line 127: nit: "DataTable"

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+        isSelectable
+        manualSelectColumn={selectColumn}
+        SelectionStatusComponent={DataTable.ControlledSelectionStatus}
+        defaultColumnValues={{ Filter: TextFilter }}
+        isPaginated
+        manualPagination
+        itemCount={users.count}
+        pageCount={users.numPages || 1}
+        // eslint-disable-next-line no-unused-vars
+        tableActions={(i) => <DownloadCsvButton />}
```
adamstankiewicz, line 209: Do we need to add in the count of rows that the download table action will apply to, as shown in the Figma spec? Also, looks like the button text is changing from "Download CSV" to "Download all" to more closely indicate that clicking it will download the entire table (without any filters applied).

We might need to track the initial number of rows in the table prior to any filtering since the data (table instance) given by the `tableActions` prop would change after applying filters.

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+            },
+            {
+              name: 'Revoked',
+              number: overview.revoked,
+              value: REVOKED,
+            }],
+          },
+          {
+            Header: 'Recent action',
+            accessor: 'recent',
```
adamstankiewicz, line 247: nit: maybe `recentAction` so it's a bit more clear what `recent` is referring when accessed elsewhere?

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+            disableFilters: true,
+          },
+        ]}
+        additionalColumns={[
+          {
+            id: 'action',
+            Header: '',
+            /* eslint-disable react/prop-types */
```
adamstankiewicz, line 255: nit: Does the `eslint-disable-next-line` comment work here to avoid needing to disable/re-enable the eslint error?

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+  const bulkRevokeOnClick = (usersToRevoke) => {
+    setRevokeModal({
+      open: true,
+      users: usersToRevoke,
+    });
+  };
+
+  return (
+    <>
+      {(users.count < 1 && userStatusFilter === null) && <SubscriptionZeroStateMessage /> }
```
adamstankiewicz, line 194: It seems like the `SubscriptionZeroStateMessage` flashes above the table briefly while the initial table data is loading.

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementRemindModal.jsx
```diff
@@ -0,0 +1,199 @@
...
+          <StatefulButton
+            state={getRemindButtonState()}
+            variant="primary"
+            onClick={handleSubmit}
+            {...{
+              labels: {
+                default: `Remind (${numberToRemind})`,
+                pending: `Reminding (${numberToRemind})`,
+                complete: 'Done',
+                error: `Retry Remind (${numberToRemind})`,
```
adamstankiewicz, line 166: nit: "Retry remind" (lowercase "remind") since only the first letter of button CTAs should be capitalized.

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementRevokeModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+          <StatefulButton
+            state={getRevokeButtonState()}
+            variant="danger"
+            onClick={handleSubmit}
+            {...{
+              labels: {
+                default: `Revoke (${numberToRevoke})`,
+                pending: `Revoking (${numberToRevoke})`,
+                complete: 'Done',
+                error: `Retry Revoke (${numberToRevoke})`,
```
adamstankiewicz, line 142: nit: "Retry revoke" (lowercase "revoke") since only the first letter of button CTAs should be capitalized.

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
manny-m: As long as we are ok iterating through the list `2n` instead of `n` times

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
manny-m: Going to defer on this. The ability to select individual rows for CSV download requires a backend change, will add a ticket to add this in the furture.

For now, Changed the button text to `Download all` and updated the Iconography to better resemble Figma spec.

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -0,0 +1,221 @@
...
+            </Alert>
+            )}
+        {showRevocationCapAlert(subscription.isRevocationCapEnabled, subscription.revocations)
+            && (
+            <Alert variant="warning">
+              You have already revoked {subscription.revocations.applied} licenses. You
+              have {subscription.revocations.remaining} revocations left on your plan.
+            </Alert>
+            )}
+        <p>This action cannot be undone. Learners with revoked licenses must be reinvited. </p>
```
long74100, line 159: nit: trailing space here

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableActionColumn.jsx
```diff
@@ -0,0 +1,83 @@
...
+          <Tooltip id="tooltip-revoke">
+            {revokeText}
+          </Tooltip>
+        )}
+      >
+        <IconButton
+          alt={revokeText}
+          title={revokeText}
+          src={RemoveCircle}
+          iconClassNames={displayRemind ? '' : '.ml-5'}
```
long74100, line 63: I think this was meant to be ml-5?

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
adamstankiewicz: SMH. Good call out, let's keep as is :)

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,304 @@
...
+        isPaginated
+        manualPagination
+        itemCount={users.count}
+        pageCount={users.numPages || 1}
+        // eslint-disable-next-line no-unused-vars
+        tableActions={(i) => {
+          if (!showSubscriptionZeroStateMessage) {
+            return <DownloadCsvButton />;
+          }
+          return <></>;
```
adamstankiewicz, line 206: nit: Could we use `return null;` here instead of an empty React fragment? Or does it need to return JSX in both cases?

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -0,0 +1,304 @@
...
+        initialState={{
+          pageSize: PAGE_SIZE,
+          pageIndex: DEFAULT_PAGE - 1,
+        }}
+        initialTableOptions={{
+          getRowId: row => row.id,
+        }}
+        EmptyTableComponent={
+          () => <DataTable.EmptyTable content={loadingUsers ? 'Loading...' : 'No results found'} />
+        }
```
adamstankiewicz, line 217: 🚀

src/containers/SaveTemplateButton/__snapshots__/SaveTemplateButton.test.jsx.snap
```diff
@@ -12,10 +12,10 @@ exports[`<SaveTemplateButton /> renders correctly in disabled state 1`] = `
   <span
     className="d-flex align-items-center justify-content-center"
   >
-    <span
-      className=""
-    >
+    <span>
+       
       Save Template
+
```
adamstankiewicz, line 18: As noted in another comment, if you upgrade to latest Paragon (16.14.7), this snapshot will be a bit different (without the leading/trailing spaces).

PR #628, opened 2021-10-06
docs: Staring up a PR template for frontend changes

binodpant:
Based on actoin item I had from retrospectives that we need to remind ourselves to do screenshot and ux reviews for PRs


johnnagro (approved):
⛵

PR #629, opened 2021-10-07
build: remove es5 compatibility check

muhammad-ammar:
**Description:** The sole purpose of ES5 was to verify that the built code is compatible with older browsers like IE 11. however, since we’re now no longer supporting IE 11, we technically no longer need to transpile code to ES5 hence we can remove this check.

# For all changes


adamstankiewicz (approved):
LGTM!

PR #630, opened 2021-10-08
fix: ENT-4745 | bugfix for csv upload for bulk assign modal

christopappas:
# For all changes

the variable hasTextAreaEmails was used instead of validEmails. so if you used to CSV upload button, hasTextAreaEmails would be blank, resulting in the following code path from being hit
```
pendingEnterpriseUserData = {
        user_email: formData['email-address'],
        enterprise_customer: enterpriseUuid,
      };
```
but formData['email-address'] is nothing, and this would need to be a map anyway

### DIFF
--- a/src/components/CodeAssignmentModal/index.jsx
+++ b/src/components/CodeAssignmentModal/index.jsx
@@ -298,7 +298,7 @@ export class BaseCodeAssignmentModal extends React.Component {
     }
 
     let pendingEnterpriseUserData;
-    if (hasTextAreaEmails) {
+    if (validEmails) {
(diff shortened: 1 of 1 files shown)


src/components/CodeAssignmentModal/index.jsx
```diff
@@ -297,18 +313,8 @@ export class BaseCodeAssignmentModal extends React.Component {
       options.codes = [code.code];
     }
 
-    let pendingEnterpriseUserData;
-    if (hasTextAreaEmails) {
```
christopappas, line 301: note `hasTextAreaEmails` gets changed to `validEmails` when i pulled out the logic. that was the bug

PR #640, opened 2021-10-29
Table Actions Clear Selection

manny-m:
# Context

This change contains no visual changes. 

In the subscription detail page, if admins selected users on the table and clicked reminded or revoked user licenses actions, their selection would be remain. This change:

1. Updates paragon library dependency
2. Rearranges components so that they are under the `DataTable` context umbrella allowing them to use the `clearSelection` function of the `DataTableContext` provider. 
3. Updated some tests and added new tests to cover some of the modal state management that shifted around.

Now revoking and reminding users will clear the table selection the user may have had.

Cloeses [ENT-5060](https://openedx.atlassian.net/browse/ENT-5060)

# For all changes

- [] Ensure to attach screenshots
- [] Ensure to have UX team confirm screenshots

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2513,9 +2513,9 @@
       }
     },
     "@edx/paragon": {
-      "version": "16.14.5",
-      "resolved": "https://registry.npmjs.org/@edx/paragon/-/paragon-16.14.5.tgz",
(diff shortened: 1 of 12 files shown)


long74100 (approved):
this lgtm 🚀 🚀

PR #644, opened 2021-11-03
fix: re-add segment events for legacy table pagination/sort and csv download

adamstankiewicz:
# For all changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,15 +3,19 @@
   "version": "0.1.0",
   "description": "The frontend-app-admin-portal is a frontend that provides branded learning experiences.",
   "repository": "https://github.com/edx/frontend-app-admin-portal",
+  "browserslist": [
+    "extends @edx/browserslist-config"
+  ],
   "scripts": {
     "build": "fedx-scripts webpack",
-    "build:gh-pages": "NODE_ENV=development BABEL_ENV=development webpack --config=config/webpack.dev.config.js",
-    "deploy:gh-pages": "npm run build:gh-pages && gh-pages -d dist",
+    "build:with-theme": "THEME=npm:@edx/brand-edx.org@latest npm run install-theme && fedx-scripts webpack",
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "lint:fix": "fedx-scripts eslint --fix --ext .js --ext .jsx .",
     "precommit": "npm run lint",
     "prepublishOnly": "npm run build",
+    "install-theme": "npm install \"@edx/brand@${THEME}\" --no-save",
     "start": "fedx-scripts webpack-dev-server --progress",
+    "start:with-theme": "THEME=npm:@edx/brand-edx.org@latest npm run install-theme && fedx-scripts webpack-dev-server --progress",
     "debug-test": "node --inspect-brk node_modules/.bin/jest --runInBand --coverage",
     "test": "fedx-scripts jest --coverage --passWithNoTests --maxWorkers=50%",
     "snapshot": "fedx-scripts jest --updateSnapshot"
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -3,34 +3,37 @@
...
-    "build:gh-pages": "NODE_ENV=development BABEL_ENV=development webpack --config=config/webpack.dev.config.js",
-    "deploy:gh-pages": "npm run build:gh-pages && gh-pages -d dist",
+    "build:with-theme": "THEME=npm:@edx/brand-edx.org@latest npm run install-theme && fedx-scripts webpack",
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "lint:fix": "fedx-scripts eslint --fix --ext .js --ext .jsx .",
     "precommit": "npm run lint",
     "prepublishOnly": "npm run build",
+    "install-theme": "npm install \"@edx/brand@${THEME}\" --no-save",
     "start": "fedx-scripts webpack-dev-server --progress",
+    "start:with-theme": "THEME=npm:@edx/brand-edx.org@latest npm run install-theme && fedx-scripts webpack-dev-server --progress",
```
adamstankiewicz, line 18: Helper to run the app with the edx.org theme used on stage/prod.

package.json
```diff
@@ -3,34 +3,37 @@
...
-    "@edx/frontend-enterprise-utils": "1.1.0",
+    "@edx/frontend-enterprise-utils": "1.1.1",
     "@edx/frontend-platform": "1.11.0",
     "@edx/paragon": "^16.14.5",
     "@fortawesome/fontawesome-svg-core": "1.2.35",
     "@fortawesome/free-brands-svg-icons": "5.15.3",
     "@fortawesome/free-regular-svg-icons": "5.15.3",
     "@fortawesome/free-solid-svg-icons": "5.15.3",
     "@fortawesome/react-fontawesome": "0.1.14",
-    "@redux-beacon/segment": "1.1.0",
```
adamstankiewicz, line 33: `@redux-beacon/segment` and `redux-beacon` are no longer used. They were previously used when we had Segment events dispatched via a redux integration, but those were ripped out during the migration to frontend-platform. This PR is re-adding those events that were ripped out.

src/components/BulkEnrollmentPage/stepper/DismissibleCourseWarning.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 
-import { Alert, WarningFilled } from '@edx/paragon';
+import { Alert } from '@edx/paragon';
+import { WarningFilled } from '@edx/paragon/icons';
```
adamstankiewicz, line 4: There was an error in console about `WarningFilled` not existing as an export in `@edx/paragon`, which makes sense because the icons are exported only at `@edx/paragon/icons`.

src/components/Admin/Admin.test.jsx
```diff
@@ -5,10 +5,14 @@ import { mount } from 'enzyme';
 import { MemoryRouter, Link } from 'react-router-dom';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
+// eslint-disable-next-line no-unused-vars
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 
 import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
 import Admin from './index';
 
+jest.mock('@edx/frontend-platform/analytics');
```
christopappas, line 14: do we need to make any assertions related to the event being sent in the tests?

christopappas (approved):

src/components/Admin/Admin.test.jsx
adamstankiewicz: Sure, we haven't asserted on Segment events in the past in our frontend repos, but I added it to these tests.

PR #649, opened 2021-11-30
feat: Bulk Enrollment Asynchronous Notification on Frontend

johnnagro:
This adds a new path to the admin MFE which is for a link emailed to an administrator after their bulk enrollment is complete. This component uses the license-manager api to generate a download link to the results csv and redirects there

# References

- [ENT-5170](https://openedx.atlassian.net/browse/ENT-5170)

# For all changes


src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+  const { addToast } = useContext(ToastsContext);
+  const [isLoading, setLoading] = useState(true);
+  const [redirectUrl, setRedirectUrl] = useState(null);
+  const [error, setError] = useState(null);
+
+  useEffect(() => {
+    if (enterpriseId && isLoading) {
+      LicenseManagerApiService.fetchBulkEnrollmentJob(enterpriseId, bulkEnrollmentJobId)
+        .then((response) => {
+          if (response.data.download_url) {
```
adamstankiewicz, line 20: Might recommend using `camelCaseObject` from `@edx/frontend-platform` to camelCase the response object so we don't need to use snake case in JS 😉 

```js
import { camelCaseObject } from '@edx/frontend-platform/utils';
```

src/data/services/LicenseManagerAPIService.js
```diff
@@ -155,6 +155,11 @@ class LicenseManagerApiService {
     return LicenseManagerApiService.apiClient().post(url, options);
   }
 
+  static fetchBulkEnrollmentJob(enterpriseUuid, bulkEnrollmentJobUuid, options) {
+    const url = `${LicenseManagerApiService.licenseManagerBaseUrl}/bulk-license-enrollment/?enterprise_customer_uuid=${enterpriseUuid}&bulk_enrollment_job_uuid=${bulkEnrollmentJobUuid}`;
```
adamstankiewicz, line 159: nit: Could use `URLSearchParams` here to construct the query params in the URL instead of doing it manually.

```jsx
const queryParams = new URLSearchParams();
queryParams.append('enterprise_customer_uuid', enterpriseUuid);
queryParams.append('bulk_enrollment_job_uuid', bulkEnrollmentJobUuid);
const url = `${LicenseManagerApiService.licenseManagerBaseUrl}/bulk-license-enrollment/?${queryParams.toString()}`;
```

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+        .then((response) => {
+          const result = camelCaseObject(response.data);
+          if (result.downloadUrl) {
+            setRedirectUrl(result.downloadUrl);
+          } else {
+            throw new Error('Your download is not ready yet.');
+          }
+        })
+        .catch((err) => {
+          logError(err);
```
adamstankiewicz, line 29: [clarification/confirmation] If the error on line 25 above is thrown any time a user lands on this page (or refreshes) and their download isn't ready, we do in fact want to report that to New Relic via `logError` and count against any alerts that get routed to on-call?

I'm just imagining someone gets the URL to this page and hammers the refresh button (🔨 😅) until they get the download completed, which could be a big burst of errors in a short period of time. If we want to potentially get alerted on bursts like this, `logError` makes sense with the custom "download isn't ready" error! Always something to come back to if it does become noisy and impacting alerts.

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+  useEffect(() => {
+    if (enterpriseId && isLoading) {
+      LicenseManagerApiService.fetchBulkEnrollmentJob(enterpriseId, bulkEnrollmentJobId)
+        .then((response) => {
+          const result = camelCaseObject(response.data);
+          if (result.downloadUrl) {
+            setRedirectUrl(result.downloadUrl);
+          } else {
+            throw new Error('Your download is not ready yet.');
```
adamstankiewicz, line 25: [clarification/suggestion] How will users navigate to this page? I'm wondering if we could poll the API every so often  until the download is ready, with some messaging along those lines. Not sure if that'd be a good approach without knowing the full context of the page but seems like it might work well?

We do something similar in the `UserActivationPage` component, where we hydrate the authenticated user on an interval (via a handy `useInterval` hook) until a field in the response data comes back as true ([code example](https://github.com/edx/frontend-app-admin-portal/blob/master/src/components/UserActivationPage/index.jsx#L44-L48)). While it's not true, appropriate messaging is shown to inform user to wait and their download will start soon, for example.

Might not be the use case you're looking for here, but polling could be an option!

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+          setError(err);
+        })
+        .finally(() => {
+          setLoading(false);
+        });
+    }
+  });
+
+  if (isLoading) {
+    return <h1>loading...</h1>;
```
adamstankiewicz, line 39: There is an `EnterpriseAppSkeleton` component you might be interested in using here as a loading state.

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+  if (error) {
+    addToast('There was a problem with your request.');
```
adamstankiewicz, line 42: In the current toast message, it seems a problem with the API (e.g., a 500) is synonymous with the download simply not being ready yet? Would we want/need to distinguish between these so the user knows that "oh, i just need to give it more time. i'll check back in 5 minutes" vs. "i need to reach out to support because it says a problem keeps occurring"?

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+    return <Redirect to={`/${enterpriseSlug}/admin/learners`} />;
+  }
+  window.location.href = redirectUrl;
+  return <h1>redirecting...</h1>;
```
adamstankiewicz, line 46: Maybe another spot for `EnterpriseAppSkeleton`? 🤔

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+const BulkEnrollmentResultsDownloadPage = ({ enterpriseId }) => {
+  const { enterpriseSlug, bulkEnrollmentJobId } = useParams();
+  const { addToast } = useContext(ToastsContext);
+  const [isLoading, setLoading] = useState(true);
+  const [redirectUrl, setRedirectUrl] = useState(null);
+  const [error, setError] = useState(null);
+
+  useEffect(() => {
+    if (enterpriseId && isLoading) {
```
adamstankiewicz, line 18: nit: Is `enterpriseId` always truthy since it's a required prop, so doesn't really need to be in the conditional? Or the reverse, is there a prop type warning in the console that says `enterpriseId` isn't provided at some point in the rendering cycle and it shouldn't be a required prop?

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
johnnagro: Thanks. I didn't realize `logError` sent to NewRelic

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
johnnagro: They'll get the URL from a notification email once their download is ready. In theory, a user should never see this page before their download is ready.

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
johnnagro: 👍

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import { Redirect, useParams } from 'react-router-dom';
+import { connect } from 'react-redux';
+import { logError } from '@edx/frontend-platform/logging';
+import { camelCaseObject } from '@edx/frontend-platform';
+import { ToastsContext } from '../Toasts';
+import EnterpriseAppSkeleton from '../EnterpriseApp/EnterpriseAppSkeleton';
+import LicenseManagerApiService from '../../data/services/LicenseManagerAPIService';
+
+const BulkEnrollmentResultsDownloadPage = ({ enterpriseId }) => {
```
binodpant, line 11: it's not specific to this PR but I would love for us to start using a pages/ vs components/ pattern at some point. Seems like we have a lot of components some of them pages, and it's unclear which one plays what role sometimes. However, we can do sth slightly different to organize the pages for the time being, if you are interested, we can actually have a src/components/bulkenrollment/ folder at least in there, and have the two pages be in there?

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+  const { enterpriseSlug, bulkEnrollmentJobId } = useParams();
+  const { addToast } = useContext(ToastsContext);
+  const [isLoading, setLoading] = useState(true);
+  const [redirectUrl, setRedirectUrl] = useState(null);
+  const [error, setError] = useState(null);
+  const [notReady, setNotReady] = useState(false);
+
+  useEffect(() => {
+    if (isLoading) {
```
binodpant, line 20: I think if we only want this download to happen once then instead of using isLoading check here, just pass the empty params to useEffect is more idiomatic? https://css-tricks.com/run-useeffect-only-once/

binodpant:
So, is the plan to add a link to the sidebar or somewhere else to manually go to this page?

binodpant:
Would be great to add a test for this component, looks like it could be made to fake out the api calls and test out the few different cases you will render. I am not 100% sure if you can test that redirect to download link, but you can probably just ensure that redirecturl is called with the right url, maybe is all you need

Also if you use react router <redirect to> vs the window.redirect, will that work? if so you can perhaps use existing testing patterns that already test with the react router (look for renderWithRouter in tests). otherwise your test may need to mock the window.location.href call.. which may also be fine.

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+  if (notReady) {
+    addToast('Your download is not ready yet.');
+    return <Redirect to={`/${enterpriseSlug}/admin/learners`} />;
+  }
+  if (error) {
+    addToast('There was a problem with your request.');
+    return <Redirect to={`/${enterpriseSlug}/admin/learners`} />;
+  }
+  window.location.href = redirectUrl;
```
binodpant, line 51: this this is entirely sufficient if it works great, just curious if using https://v5.reactrouter.com/web/api/Redirect/to-string here will make testing easier (if it even works). Otherwise, your test can just mock out window.location.href call somehow (with may be weird stuff like this: https://remarkablemark.org/blog/2021/04/14/jest-mock-window-location-href/ )

johnnagro: Unfortunately, in my testing/research I found that the react `Redirect` doesn't let you redirect off to a different app/domain - in this case I am sending the user to a pre-signed S3 link.

src/components/BulkEnrollmentResultsDownloadPage/index.jsx
johnnagro: Sounds good - I am still pretty green with React so you'll have to lead the way in defining the conventions.

alex-sheehan-edx (approved):
lgtm!

PR #654, opened 2021-12-01
fix: Making bulk revoke action modal color filled in/ opaque

christopappas:
# For all changes

### DIFF
--- a/src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
+++ b/src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
@@ -193,7 +193,7 @@ const LicenseManagementTableBulkActions = ({
           <Icon src={MoreVert} />
         </Button>
         <ModalPopup positionRef={target} isOpen={isOpen} onClose={close}>
-          <div>
+          <div className="bg-white p-2 rounded shadow">
             <Button
               variant="outline-danger"
               iconBefore={RemoveCircle}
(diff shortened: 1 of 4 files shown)


christopappas:
Screen Shot 2021-12-01 at 1 48 55 PM
Revoke button now filled in

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -195,7 +195,7 @@ const LicenseManagementTableBulkActions = ({
         <ModalPopup positionRef={target} isOpen={isOpen} onClose={close}>
           <div>
             <Button
-              variant="outline-danger"
+              variant="danger"
```
manny-m, line 198: Christ this looks good, the only concern is that this makes the styling on the buttons inconsistent.

Additionally, it actually doesn't address the actual issue that the `ModalPopup` background is transparent. IIRC @adamstankiewicz said the reason it might be transparent is that the development environments weren't properly loading the edx theme. In production, it works as intended.

I also remember this being added as a grooming note on the ticket but maybe I'm remembering incorrectly or it wasn't saved.

christopappas: it does appear it renders correctly on stage. I don't remember the conversation, and it's not in the ticket, so good call out 😅 

i can look into the theme bit

macdiesel (approved):

manny-m (approved):
👍

christopappas:
This now also includes adds a background for the Revoke button. As more buttons are added to that ... menu we'll need something to make the adjacent buttons look grouped up

PR #662, opened 2021-12-17
fix: reporting config page create and delete bugs

alex-sheehan-edx:
# For all changes

### DIFF
--- a/src/components/ReportingConfig/ReportingConfigForm.jsx
+++ b/src/components/ReportingConfig/ReportingConfigForm.jsx
@@ -141,7 +141,9 @@ class ReportingConfigForm extends React.Component {
       submitState,
     } = this.state;
     const selectedCatalogs = (config?.enterpriseCustomerCatalogs || []).map(item => item.uuid);
-    const dataTypesOptions = reportingConfigTypes.dataType.map(item => ({ label: item[1], value: item[0] }));
+    const dataTypesOptions = reportingConfigTypes.dataType.map((item, index) => ({
+      key: index, label: item[1], value: item[0],
+    }));
     const dataTypesOptionsValues = dataTypesOptions.map(item => item.value);
     const selectedDataTypesOption = config ? [{ label: config.dataType, value: config.dataType, hidden: true }] : [];
     return (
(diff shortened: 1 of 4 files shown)


src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -141,7 +141,9 @@ class ReportingConfigForm extends React.Component {
       submitState,
     } = this.state;
     const selectedCatalogs = (config?.enterpriseCustomerCatalogs || []).map(item => item.uuid);
-    const dataTypesOptions = reportingConfigTypes.dataType.map(item => ({ label: item[1], value: item[0] }));
+    const dataTypesOptions = reportingConfigTypes.dataType.map((item, index) => ({
```
binodpant, line 144: think worth a quick comment here on why `item[1]` works?

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -404,7 +406,13 @@ ReportingConfigForm.propTypes = {
     uuid: PropTypes.string,
     title: PropTypes.string,
   })).isRequired,
-  reportingConfigTypes: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)).isRequired,
+  reportingConfigTypes: PropTypes.shape({
+    dataType: PropTypes.array,
```
binodpant, line 410: is there an arrayOf we can use here to be more exact on what type of array?

src/components/ReportingConfig/index.jsx
```diff
@@ -73,7 +73,10 @@ class ReportingConfig extends React.Component {
         .findIndex(config => config.uuid === uuid);
 
       this.setState((state) => {
-        const newReportingConfig = { ...state.reportingConfigs }.splice(deletedIndex, 1);
+        // Copy the existing, needs to be updated, list of reporting configs
+        const newReportingConfig = [...state.reportingConfigs];
+        // Splice out the one that's being deleted
+        newReportingConfig.splice(deletedIndex, 1);
```
binodpant, line 79: maybe worth a quick unit test like directly calling `deleteconfig` or sth like it? I have not checked existing tests and how easy it is to fit this case into those tests but a unit test with mocks should be fine too

src/components/ReportingConfig/ReportingConfigForm.jsx
binodpant: yeah I would rather the type system (TS or language) handle typing but in absence of it, we get to do this fun type stuff lol)

binodpant (approved):
all set

src/components/ReportingConfig/ReportingConfigForm.test.jsx
```diff
@@ -114,6 +114,23 @@ const createConfig = () => { };
...
+      <ReportingConfigForm
+        config={config}
+        createConfig={createConfig}
+        updateConfig={updateConfig}
+        availableCatalogs={availableCatalogs}
+        reportingConfigTypes={reportingConfigTypes}
+        deleteConfig={mock}
+      />
+    ));
+    // It's finding three buttons for some reason??
```
binodpant, line 129: yeah I am coming to prefer react testing library for this reasno, where it lets us focus more on user-visible or user-interactible pieces rather than dom details like this. but yeah not part of scope for this PR!

binodpant: in that setup, this line would ideally be sth like `userEvent.click(screen.findByRole('submit-button'))` which is also more a11y friendly

src/components/ReportingConfig/ReportingConfigForm.test.jsx
```diff
@@ -251,7 +268,7 @@ describe('<ReportingConfigForm />', () => {
       />
     ));
     expect(
-      wrapper.find('select#enterpriseCustomerCatalogs').instance().value = ['test-enterprise-customer-catalog'],
-    );
+      wrapper.find('select#enterpriseCustomerCatalogs').instance().value,
+    ).toEqual('test-enterprise-customer-catalog');
```
binodpant, line 272: ✅

src/components/ReportingConfig/ReportingConfigForm.test.jsx
alex-sheehan-edx: agreed, although it's not the submit button, but the delete one 😅

PR #663, opened 2021-12-22
feat: load invite keys from API

adamstankiewicz:
* Loads invite keys from `/enterprise/api/v1/enterprise-customer-invite-key/basic-list/`
* Wires up invite key deactivation with modal
* Wires up link management disable with modal
* Wires up link deactivation modal
* Handles copying invite URL to clipboard with toast confirmation
* Other miscellaneous page cleanup

## Screenshots

image

image

image

image

image

image

# For all changes


src/components/settings/SettingsAccessTab/DisableLinkManagementAlertModal.jsx
```diff
@@ -0,0 +1,86 @@
...
+      onClose();
+    }
+  };
+
+  const handleDisableLinkManagement = () => {
+    const disableLinkManagement = async () => {
+      setModalDisableButtonState('pending');
+      try {
+        // TODO: make legit API request
+        await new Promise((resolve) => {
```
binodpant, line 29: presuming there is going to be a real api call here soon?

src/components/settings/SettingsAccessTab/SettingsAccessLinkManagement.jsx
```diff
@@ -1,121 +1,145 @@
...
+            {
+              id: 'action',
+              Header: '',
+              Cell: props => <ActionsTableCell {...props} onDeactivateLink={handleDeactivatedLink} />,
+            },
+          ]}
+        >
+          <DataTable.TableControlBar />
+          <DataTable.Table />
+          <DataTable.EmptyTable content={loadingLinks ? 'Loading...' : 'No links found'} />
```
binodpant, line 125: interesting... looks like we want to use this pattern of loading when rendering emptyTable in other places too (I see we use one of these without the 'loading' smarts in explore catalog. Also maybe datatable can auto do 'loading..' by default in some way (just a thought)

src/components/settings/index.jsx
```diff
@@ -30,12 +30,10 @@ const SettingsPage = () => {
       <Switch>
         <Redirect
           exact
-          key="settings"
```
binodpant, line 33: don't quite get what `key` was doing in the first place ❓

src/components/settings/SettingsAccessTab/DisableLinkManagementAlertModal.jsx
adamstankiewicz: That is correct! But, those API endpoints don't yet exist so we're just creating a placeholder for where those API calls might go once they do exist 😄

src/components/settings/SettingsAccessTab/SettingsAccessLinkManagement.jsx
adamstankiewicz: @binodpant There will soon be a `isLoading` prop that can be added to `DataTable` to show a loading spinner overlaying the table data (see [this PR](https://github.com/edx/paragon/pull/991) for its status). Once that it in place, I would imagine this component will use `isLoading` instead of changing up the content of `DataTable.EmptyTable`.

image

src/components/settings/index.jsx
adamstankiewicz: Here, it wasn't doing anything haha. `key` are generally used when rendering a node as part of iterating over an array, but that's not the case here.

Another fun fact about `key`: if `key` changes, the component it's on will be forced to re-render!

binodpant (approved):

src/components/settings/data/service.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/prefer-default-export */
+
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
+import { getConfig } from '@edx/frontend-platform/config';
+
+export const getAccessLinks = (enterpriseUUID) => {
```
manny-m, line 6: I think it might be better to add this to `/src/data/services/LmsApiService.js` for two reasons:
- Consistency, all other parts of the app use it as the source of their API routes
- Single source, keeping the service in one place would make it easier to share and update the service

src/components/settings/SettingsTabs.jsx
```diff
@@ -40,12 +41,12 @@ const SettingsTabs = () => {
   };
 
   return (
-    <div className="m-4">
+    <Container className="py-4" fluid>
```
manny-m, line 44: Not sure why we use `Container` over just a `div` here? Is it for the classes that come with `fluid`?

src/components/settings/tests/SettingsTabs.test.jsx
```diff
@@ -32,7 +32,7 @@ const settingsTabsWithRouter = () => (
   </MemoryRouter>
 );
 
-describe('<SettingsTabs />', () => {
+describe('<SettingsPage />', () => {
```
manny-m, line 35: nit: Why do we describe these tests as `<SettingsPage />` when we are testing `<SettingsTabs />` on this page? Should we just name `SettingsTab` to `SettingsPage`?

manny-m (commented):
This looks great 😲 
Left some questions, especially curious about the one I left on `src/components/settings/data/service.js`

src/components/settings/data/service.js
adamstankiewicz: Ah, yes. Great suggestion. Will update :)

src/components/settings/tests/SettingsTabs.test.jsx
adamstankiewicz: D'oh, good catch. This wasn't meant to be changed! Changing it back to `SettingsTabs`. We already have a `SettingsPage` component (index.jsx).

src/components/settings/SettingsTabs.jsx
adamstankiewicz: Yeah, I noticed the spacing was a bit off with the previous use of `Container`. By moving the `Container` to `SettingsTabs` instead of where it was previously, the spacing is more in line with the containers for other pages with regard to left/right padding.

manny-m (approved):
Awesome!

PR #664, opened 2021-12-27
feat: ENT-4554 Added tableau iframe in react component

irfanuddinahmad:
# For all changes


src/components/analytics/AnalyticsCharts.jsx
```diff
@@ -1,8 +1,7 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { logError } from '@edx/frontend-platform/logging';
-
-import TableauReport from 'tableau-react-embed';
+import url from 'url';
```
zamanafzal, line 4: Is this library built-in available in react or do we need to install it manually? Asking it because if we have to install it manually then we should add it in the Package.json file as well.

src/components/analytics/AnalyticsCharts.jsx
```diff
@@ -12,32 +11,53 @@ import AnalyticsApiService from './data/service';
...
+  };
+  const getUrl = () => {
+    const parsed = url.parse(tableauUrl, true);
+    const { protocol, host, pathname } = parsed;
+    const query = '?:embed=yes&:comments=no&:toolbar=no&:refresh=yes';
+    if (!tokenUsedOnce && token) {
+      setTokenUsedOnce(true);
+      return `${protocol}//${host}/trusted/${token}${pathname}${query}`;
+    }
+    return `${protocol}//${host}${pathname}${query}`;
```
zamanafzal, line 35: What will happen here as we are not passing the token in this URL? Is it to show some kind of error?

package.json
```diff
@@ -65,7 +65,6 @@
     "redux-mock-store": "1.5.4",
     "redux-thunk": "2.3.0",
     "regenerator-runtime": "0.13.7",
-    "tableau-react-embed": "1.1.0",
```
zamanafzal, line 68: I think there are some packages that are being removed from package-lock.json are not related to this like `es6-promise`.

src/components/analytics/AnalyticsCharts.jsx
irfanuddinahmad: This is available as part of node

src/components/analytics/AnalyticsCharts.jsx
irfanuddinahmad: Yes. This will trigger an un-authorized message on the iframe.

package.json
irfanuddinahmad: I think they may be related to the now removed tableau-react-embed

saleem-latif (approved):

.env.development
```diff
@@ -39,3 +39,5 @@ FEATURE_FILE_ATTACHMENT=true
 IS_MAINTENANCE_ALERT_ENABLED=''
 MAINTENANCE_ALERT_MESSAGE=''
 MAINTENANCE_ALERT_START_TIMESTAMP=''
+TABLEAU_URL='https://enterprise-tableau.edx.org'
```
muhammad-ammar, line 42: shouldn't this value come from remote config?

.env.test
```diff
@@ -10,3 +10,4 @@ FAVICON_URL='https://edx-cdn.org/v3/default/favicon.ico'
 FEATURE_FILE_ATTACHMENT=true
 ENTERPRISE_SUPPORT_URL = ''
 ENTERPRISE_SUPPORT_REVOKE_LICENSE_URL = ''
+TABLEAU_URL='https://enterprise-tableau.edx.org'
```
muhammad-ammar, line 13: do we really need to add a real url here? can we use some fake url?

src/components/analytics/AnalyticsCharts.jsx
```diff
@@ -12,32 +11,53 @@ import AnalyticsApiService from './data/service';
 // eslint-disable-next-line no-unused-vars
 export default function AnalyticsCharts(enterpriseId) {
   const [token, setToken] = useState(null);
+  const [tokenUsedOnce, setTokenUsedOnce] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState(null);
-  // Fetch token
+  const ref = useRef(null);
+  let tableauUrl = 'https://enterprise-tableau.edx.org/views/enterpriseadminanalytics/enroll_dash';
```
muhammad-ammar, line 18: shouldn't we use `configuration.TABLEAU_URL` rather than hardcoding the base url?

src/components/analytics/AnalyticsCharts.jsx
```diff
@@ -12,32 +11,53 @@ import AnalyticsApiService from './data/service';
...
+  if (configuration.TABLEAU_URL) {
```
muhammad-ammar, line 19: I think we can remove this `if`? I believe `TABLEAU_URL` should always be there.

zamanafzal (approved):

public/index.html
```diff
@@ -5,6 +5,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="shortcut icon" href="<%=htmlWebpackPlugin.options.FAVICON_URL%>" type="image/x-icon" />
+    <script type="text/javascript" src="https://enterprise-tableau.edx.org/javascripts/api/tableau-2.min.js"></script>
```
muhammad-ammar, line 8: this script will be loaded for all but as per my understanding this should only load for the enterprise customers having tableau analytics enabled?

.env.development
irfanuddinahmad: The one from the remote config will overwrite this value.

.env.test
irfanuddinahmad: The address is already public in DNS records. I donot see any reason to hide it.

src/components/analytics/AnalyticsCharts.jsx
irfanuddinahmad: The one from the remote config will overwrite this value.

public/index.html
irfanuddinahmad: added async

src/components/analytics/AnalyticsCharts.jsx
irfanuddinahmad: lets just keep it for safety

.env.development
irfanuddinahmad: No

PR #668, opened 2022-01-04
Adds generate link button to settings page

manny-m:
### Closes [ENT-5178](https://openedx.atlassian.net/browse/ENT-5178)

This change adds a button that generates an enterprise customer invite key.
This button has 3 states:
- **Default**, the button is ready to use
- **Loading**, disabled while loading required information before it can be used. This should only happen for a few seconds on page load.
- **Pending**, disabled while generating an enterprise customer invite key

**Default**:
Screen Shot 2021-12-10 at 1 05 47 AM

**Loading**
Screen Shot 2022-01-04 at 9 47 41 AM

**Pending**
Screen Shot 2022-01-04 at 9 48 01 AM

# For all changes


src/components/settings/SettingsAccessTab/SettingsAccessLinkManagement.jsx
```diff
@@ -44,31 +44,19 @@ const SettingsAccessLinkManagement = ({ enterpriseUUID }) => {
...
-    setisLinkManagementEnabled(false);
+    setIsLinkManagementEnabled(false);
     setIsLinkManagementAlertModalOpen(false);
   };
 
   const handleLinkManagementFormSwitchChanged = (e) => {
     const isChecked = e.target.checked;
     if (isChecked) {
-      setisLinkManagementEnabled(e.target.checked);
+      setIsLinkManagementEnabled(e.target.checked);
```
adamstankiewicz, line 54: tiniest of nits: maybe replace `e.target.checked` with `isChecked` since we already have its value stored in a variable?

src/components/settings/SettingsAccessTab/SettingsAccessLinkManagement.jsx
```diff
@@ -83,11 +71,8 @@ const SettingsAccessLinkManagement = ({ enterpriseUUID }) => {
           itemCount={links.length}
           // eslint-disable-next-line no-unused-vars
           tableActions={(i) => (
-            // TODO: Consider making this its own component with logic
-            <StatefulButton
-              {...generateLinkButtonProps}
-              onClick={handleGenerateLinkSuccess}
-              disabledStates={!isLinkManagementEnabled ? Object.keys(generateLinkButtonProps.labels) : undefined}
+            <SettingsAccessGenerateLinkButton
+              onSuccess={handleGenerateLinkSuccess}
```
adamstankiewicz, line 75: Currently, when the "Access via Link" section is disabled via the `Form.Switch`, it disables the "Generate link" button as it doesn't make sense to generate a link when "Access via Link" is turned off.

Are we no longer disabling the button here when link management isn't enabled?

image

src/components/settings/SettingsAccessTab/tests/SettingsAccessGenerateLinkButton.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+import {
+  screen,
+  render,
+  cleanup,
+  act,
+} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import moment from 'moment';
+
+import ContextProvider from '../../SettingsContext';
```
adamstankiewicz, line 11: nit: maybe `SettingsContextProvider` to be more explicit 🤷‍♂️

src/components/settings/data/hooks.js
```diff
@@ -60,7 +61,59 @@ export const useLinkManagement = (enterpriseUUID) => {
   };
 };
 
+const initialCustomerAgreementState = {
+  netDaysUntilExpiration: 0,
+};
+/*
+ * @param {Objet {enterpriseId: string}}
```
adamstankiewicz, line 68: nit: "Object"?

src/components/settings/index.jsx
```diff
@@ -14,17 +16,18 @@ import {
...
 /**
  * Behaves as the router for settings page
  * When browsing to {path} (../admin/settings) redirect to default tab
  */
-const SettingsPage = () => {
+const SettingsPage = ({ enterpriseId }) => {
   const { path } = useRouteMatch();
   return (
-    <>
+    <SettingsContext enterpriseId={enterpriseId}>
```
adamstankiewicz, line 30: nit: Curious whether `SettingsPage` should be connected to the Redux store to grab the `enterpriseId` if this component solely uses it to pass it to `SettingsContext`? Or should `SettingsContext` be the component that's connected to the Redux store to grab the `enterpriseId` and avoid prop drilling here?

src/components/settings/data/hooks.js
```diff
@@ -60,7 +61,59 @@ export const useLinkManagement = (enterpriseUUID) => {
...
+ */
+export const useCustomerAgreementData = ({ enterpriseId }) => {
+  const [customerAgreement, setCustomerAgreement] = useState(initialCustomerAgreementState);
+  const [loadingCustomerAgreement, setLoadingCustomerAgreement] = useState(true);
+
+  const loadCustomerAgreementData = (page = 1) => {
+    if (!loadingCustomerAgreement) {
+      setLoadingCustomerAgreement(true);
+    }
+    LicenseManagerApiService.fetchCustomerAgreementData({ enterprise_customer_uuid: enterpriseId, page })
```
adamstankiewicz, line 82: Preferably use `async`/`await`, but it's kinda personal preference, so will leave it up to you if you want to address :)

src/components/settings/data/hooks.js
```diff
@@ -60,7 +61,59 @@ export const useLinkManagement = (enterpriseUUID) => {
...
+ * }
+ */
+export const useCustomerAgreementData = ({ enterpriseId }) => {
+  const [customerAgreement, setCustomerAgreement] = useState(initialCustomerAgreementState);
+  const [loadingCustomerAgreement, setLoadingCustomerAgreement] = useState(true);
+
+  const loadCustomerAgreementData = (page = 1) => {
+    if (!loadingCustomerAgreement) {
+      setLoadingCustomerAgreement(true);
+    }
```
adamstankiewicz, line 81: nit: I think this could probably be simplified to just `setLoadingCustomerAgreement(true)`; not sure the `if` is really adding much here?

src/components/settings/data/hooks.js
```diff
@@ -60,7 +61,59 @@ export const useLinkManagement = (enterpriseUUID) => {
...
+    LicenseManagerApiService.fetchCustomerAgreementData({ enterprise_customer_uuid: enterpriseId, page })
+      .then((response) => {
+        const newCustomerAgreementData = {
+          netDaysUntilExpiration: 0,
+        };
+        const { data: customerAgreementData } = camelCaseObject(response);
+        if (customerAgreementData.results && customerAgreementData.count) {
+          // Only look at customer agreements with subs
+          customerAgreementData.results.filter(result => (result.subscriptions && result.subscriptions.length))
```
adamstankiewicz, line 90: Interesting. I get why you implemented it this way (i.e., the API response is paginated and it seems your logic may have been inspired by [this code](https://github.com/openedx/frontend-app-admin-portal/blob/master/src/components/subscriptions/data/hooks.js#L28)), but the CustomerAgreement model's `enterprise_customer_uuid` field is [unique](https://github.com/openedx/license-manager/blob/master/license_manager/apps/subscriptions/models.py#L76), so the API call should only return at most 1 `CustomerAgreement` at any given time. This means the iteration you're doing over a list of `CustomerAgreement` objects might be unnecessary.

What are your thoughts about baking in this assumption of at most 1 `CustomerAgreement` into this logic rather than doing the iteration? I guess the reason why we might want to keep this iteration logic in place might be so that we don't need to make any changes to the frontend if/when the `CustomerAgreement` model no longer has a unique constraint of the `enterprise_customer_uuid`. What do you think?

src/components/settings/SettingsAccessTab/tests/SettingsAccessGenerateLinkButton.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+    render(renderWithContext());
+    expect(screen.queryByText('Generate link')).toBeTruthy();
+  });
+
+  it('Displays loading state', () => {
+    render(renderWithContext(true));
+    expect(screen.queryByText('Readying link generation')).toBeTruthy();
+  });
+
+  it('Clicking button calls api', async () => {
```
adamstankiewicz, line 55: Should this test also assert that `onSuccess` is called appropriately with `toHaveBeenCalledTimes` and/or `toHaveBeenCalledWith`?

src/components/settings/data/hooks.js
```diff
@@ -60,7 +61,59 @@ export const useLinkManagement = (enterpriseUUID) => {
...
+/*
+ * @param {Objet {enterpriseId: string}}
+ * @returns {customerAgreement: Object, loadingCustomerAgreement: bool}
+ * customerAgreement:{
+ *  netDaysUntilExpiration: number
+ * }
+ */
+export const useCustomerAgreementData = ({ enterpriseId }) => {
+  const [customerAgreement, setCustomerAgreement] = useState(initialCustomerAgreementState);
+  const [loadingCustomerAgreement, setLoadingCustomerAgreement] = useState(true);
```
adamstankiewicz, line 76: Assuming `loadCustomerAgreementData` is going to set `loadingCustomerAgreement` to true below, we can probably have this initial state be defaulted to false?

adamstankiewicz (commented):
This is looking good! I left some comments throughout. Some minor nits, suggestions to use async/await (but not necessary), and a couple other things that we might want to address.

Also, it looks like `npm run test` is failing on one of the test suites related to these changes due to an unknown import:

> Cannot find module '../../../../data/services/LicenseManagerApiService' from 'src/components/settings/data/tests/hooks.test.js'

src/components/settings/data/hooks.js
manny-m: I don't feel strongly about this. I did get most of this code from the code you linked and thought about applying the assumption you described... 

But basically just don't like hard coding a zero like this:
```
customerAcgreement = customerAgreementData.results[0]
```
Also gives us some way to reason in the case there is a freak accident and we get more than one `CustomerAgreement`.

If you feel like making the constraint that we always look at the first thing in `customerAgreementData.results` I'd be ok doing so.

src/components/settings/SettingsAccessTab/SettingsAccessGenerateLinkButton.jsx
```diff
@@ -0,0 +1,70 @@
...
+  } else if (loadingLinkCreation) {
+    buttonState = 'pending';
+  }
+
+  const handleGenerateLink = () => {
+    setLoadingLinkCreation(true);
+    // Generate expiration date of netDaysUntilExpiration
+    const expirationDate = moment().add(netDaysUntilExpiration, 'days').startOf('day').format();
+
+    LmsApiService.createEnterpriseCustomerLink(enterpriseId, expirationDate)
```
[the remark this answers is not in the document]
macdiesel: FWIW I do agree that our primary pattern should be `async` / `await` wherever possible.

src/components/settings/data/hooks.js
manny-m: Yes! 🚀

src/components/settings/SettingsAccessTab/SettingsAccessGenerateLinkButton.jsx
```diff
@@ -0,0 +1,70 @@
...
+import moment from 'moment';
+import { logError } from '@edx/frontend-platform/logging';
+
+import LmsApiService from '../../../data/services/LmsApiService';
+import { SettingsContext } from '../SettingsContext';
+
+const BUTTON_PROPS = {
+  labels: {
+    default: 'Generate link',
+    pending: 'Generating link',
```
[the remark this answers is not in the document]
manny-m: I'm ok with this, assuming we are ok that users will be presented with a disabled button (for not a very long time) with no real feedback as to why it's disabled.

src/components/settings/SettingsAccessTab/SettingsAccessLinkManagement.jsx
manny-m: Complete oversight on my part! Good catch!

src/components/settings/SettingsAccessTab/SettingsAccessGenerateLinkButton.jsx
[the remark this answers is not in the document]
adamstankiewicz: @manny-m Like you mentioned, because it's going to be disabled for such a short period of time (should be hardly noticeable on stage/production), disabling the button briefly feels preferable to showing a flash of long button text that the user likely won't have time to read anyways.

If we absolutely need to optimize for showing any feedback about loading state, we could do it conditionally. Disable the button for the first 200ms and then conditionally show the longer text for any duration of loading time longer than that.

Another reasonable option: don't render the "Generate link" button until the customer agreement data is loaded (avoids the problem of disabled button without feedback).

adamstankiewicz (approved):

PR #670, opened 2022-01-05
fix: prevent allUsersSelected in datatable from being true if filter …

christopappas:
…is applied

# For all changes

### DIFF
--- a/src/components/subscriptions/data/utils.js
+++ b/src/components/subscriptions/data/utils.js
@@ -2,10 +2,10 @@ import moment from 'moment';
 import {
   SEEN_SUBSCRIPTION_EXPIRATION_MODAL_COOKIE_PREFIX,
   ASSIGNED,
-  ACTIVATED,
   ACTIVE,
   SCHEDULED,
   ENDED,
+  REVOCABLE_STATUSES,
 } from './constants';
 
 // eslint-disable-next-line import/prefer-default-export
@@ -16,7 +16,7 @@ export const getSubscriptionExpiringCookieName = ({
   return cookieName;
 };
 
-export const canRevokeLicense = (licenseStatus) => licenseStatus === ASSIGNED || licenseStatus === ACTIVATED;
+export const canRevokeLicense = (licenseStatus) => REVOCABLE_STATUSES.includes(licenseStatus);
 
 export const canRemindLicense = (licenseStatus) => licenseStatus === ASSIGNED;
 
@@ -50,3 +50,16 @@ export const sortSubscriptionsByStatus = (subscriptions) => subscriptions.slice(
     return orderByStatus[sub1Status] - orderByStatus[sub2Status];
   },
 );
+
+export const transformFiltersForRequest = (filters) => {
+  const nameMappings = {
+    emailLabel: 'user_email',
+    statusBadge: 'status_in',
+  };
+  return filters.map(
+    filter => ({
+      name: nameMappings[filter.name],
+      filter_value: filter.filterValue,
+    }),
+  );
+};
--- a/src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
+++ b/src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import {
   Alert,
@@ -18,6 +18,7 @@ import { validateEmailTemplateForm } from '../../../../data/validation/email';
 import LicenseManagerApiService from '../../../../data/services/LicenseManagerAPIService';
 import { configuration } from '../../../../config';
 import { getSubscriptionContactText } from '../../../../utils';
+import { transformFiltersForRequest } from '../../data/utils';
 
 const generateEmailTemplate = (contactEmail) => ({
   greeting: 'We noticed you haven’t had a chance to start learning on edX! It’s easy to get started and browse the course catalog.',
@@ -31,27 +32,19 @@ const generateEmailTemplate = (contactEmail) => ({
 
 /**
  * Returns StatefulButton labels
- * @param {boolean} remindAllUsers
- * @param {number} userCount
  * @param {number} totalToRemind
  * @returns {Object}
  */
-const generateRemindModalSubmitLabel = (remindAllUsers, userCount, totalToRemind) => {
-  // If we are not reminding all users, use users.length
-  // Else if totalToRemind is passed use that
-  // Else use 'All'
-  let buttonNumberLabel = 'All';
-  if (!remindAllUsers) {
-    buttonNumberLabel = userCount;
-  } else if (totalToRemind > 0) {
-    buttonNumberLabel = totalToRemind;
+const generateRemindModalSubmitLabel = (totalToRemind) => {
+  let buttonNumberLabel = 'all';
+  if (Number.isFinite(totalToRemind)) {
+    buttonNumberLabel = `(${totalToRemind})`;
   }
-
   return {
-    default: `Remind (${buttonNumberLabel})`,
-    pending: `Reminding (${buttonNumberLabel})`,
+    default: `Remind ${buttonNumberLabel}`,
+    pending: `Reminding ${buttonNumberLabel}`,
     complete: 'Done',
-    error: `Retry remind (${buttonNumberLabel})`,
+    error: `Retry remind ${buttonNumberLabel}`,
   };
 };
 
@@ -65,17 +58,18 @@ const LicenseManagementRemindModal = ({
   remindAllUsers,
   totalToRemind,
   contactEmail,
+  activeFilters,
 }) => {
   const [requestState, setRequestState, initialRequestState] = useRequestState(isOpen);
 
   const [emailTemplate, setEmailTemplate] = useState(generateEmailTemplate(contactEmail));
   const isExpired = moment().isAfter(subscription.expirationDate);
 
-  const buttonLabels = generateRemindModalSubmitLabel(remindAllUsers, usersToRemind.length, totalToRemind);
+  const buttonLabels = generateRemindModalSubmitLabel(totalToRemind);
 
-  const title = `Remind User${remindAllUsers || usersToRemind.length > 1 ? 's' : ''}`;
+  const title = `Remind User${remindAllUsers || totalToRemind > 1 ? 's' : ''}`;
 
-  const handleSubmit = () => {
+  const handleSubmit = useCallback(async () => {
     if (onSubmit) {
       onSubmit();
     }
@@ -88,30 +82,40 @@ const LicenseManagementRemindModal = ({
       return;
     }
 
-    const userEmailsToRemind = usersToRemind.map((user) => user.email);
+    const makeRequest = async () => {
+      const options = {
+        greeting: emailTemplate.greeting,
+        closing: emailTemplate.closing,
+      };
 
-    const options = {
-      greeting: emailTemplate.greeting,
-      closing: emailTemplate.closing,
+      const filtersPresent = activeFilters.length > 0;
+
+      // If reminding all users and there are no filters, hit remind-all endpoint
+      if (remindAllUsers && !filtersPresent) {
+        return LicenseManagerApiService.licenseRemindAll(subscription.uuid);
+      }
+
+      const userEmailsToRemind = usersToRemind.map((user) => user.email);
+
+      // If emails are passed in, send them in the payload.
+      if (userEmailsToRemind.length > 0) {
+        options.user_emails = userEmailsToRemind;
+      } else {
+        options.filters = transformFiltersForRequest(activeFilters);
+      }
+
+      return LicenseManagerApiService.licenseBulkRemind(subscription.uuid, options);
     };
-    if (!remindAllUsers) {
-      options.user_emails = userEmailsToRemind;
-    }
 
-    LicenseManagerApiService.licenseRemind(
-      options,
-      subscription.uuid,
-      remindAllUsers === undefined ? false : remindAllUsers,
-    )
-      .then((response) => {
-        setRequestState({ ...initialRequestState, success: true });
-        onSuccess(response);
-      })
-      .catch((error) => {
-        logError(error);
-        setRequestState({ ...initialRequestState, error });
-      });
-  };
+    try {
+      const response = await makeRequest();
+      setRequestState({ ...initialRequestState, success: true });
+      onSuccess(response);
+    } catch (error) {
+      logError(error);
+      setRequestState({ ...initialRequestState, error });
+    }
+  }, [onSubmit, activeFilters, emailTemplate, remindAllUsers, usersToRemind, subscription.uuid]);
 
   const handleClose = () => {
     if (!requestState.loading) {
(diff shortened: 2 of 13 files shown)


src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -150,6 +150,31 @@ const LicenseManagementTable = () => {
     [currentPage],
   );
 
+  const filtersAreAppliedToColumns = (
```
christopappas, line 153: can probably do something fancy here to not need these 2 helper functions (when passing stuff into the bulkaction)

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -90,11 +103,26 @@ const LicenseManagementRevokeModal = ({
     }
     setRequestState({ ...initialRequestState, loading: true });
     const makeRequest = () => {
-      if (revokeAllUsers) {
+      const filtersPresent = checkIfFiltersPresent(filtersAndValues)
```
christopappas, line 106: filters or not, if a user had clicked "select all", whenever they unclick a specific user, isEntireTableSelected gets set to false (which flows though to this file as `revokeAllUsers`), so the "unclick" scenario will never hit revoke-all endpoint or the bulk-revoke (with filters), and will only hit bulk-revoke with user_emails

The addition of filtersPresent will protect us again the "has filter, selects all, revokes every single license". Once the backend implements filters as an option, then in the case where the "entire table is selected" with a filter, we simply pass that filter to the backend and perform the action on (hopefully) the same licenses. Deeefinitely going to need to have some thorough testing here

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -62,6 +62,18 @@ const generateRevokeModalSubmitLabel = (revokeAllUsers, userCount, totalToRevoke
   };
 };
 
+const checkIfFiltersPresent = (filtersAndValues) => {
```
adamstankiewicz, line 65: nit: this could likely be simplified as

```jsx
const checkIfFiltersPresent = filtersAndValues => filtersAndValues.some(filter => !!filter.filterValue);
```

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -90,11 +103,26 @@ const LicenseManagementRevokeModal = ({
     }
     setRequestState({ ...initialRequestState, loading: true });
     const makeRequest = () => {
-      if (revokeAllUsers) {
+      const filtersPresent = checkIfFiltersPresent(filtersAndValues)
+
+      // If all users are selected and you have no filters, hit revoke-all endpoint
+      if (revokeAllUsers && filtersPresent === false) {
```
adamstankiewicz, line 109: nit: `filtersPresent === false` ➡️  `!filtersPresent`

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -90,11 +103,26 @@ const LicenseManagementRevokeModal = ({
...
         return LicenseManagerApiService.licenseRevokeAll(subscription.uuid);
       }
+
+      // If all users are selected and you DO have filters, but bulk-revoke endpoint
+      if (revokeAllUsers && filtersPresent === true) {
```
adamstankiewicz, line 114: nit: `filtersPresent === true` ➡️ `filtersPresent`

src/components/subscriptions/licenses/LicenseManagementTable/index.jsx
```diff
@@ -150,6 +150,19 @@ const LicenseManagementTable = () => {
     [currentPage],
   );
 
+  const filtersAndValues = (
```
adamstankiewicz, line 153: nit: might be simpler as

```jsx
const filtersAndValues = columns => columns.map(column => ({
  name: column.id,
  filter: column.filter,
  filterValue: column.filterValue,
});
```

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -287,6 +290,13 @@ LicenseManagementTableBulkActions.propTypes = {
   onRevokeSuccess: PropTypes.func.isRequired,
   onEnrollSuccess: PropTypes.func.isRequired,
   allUsersSelected: PropTypes.bool.isRequired,
+  filtersAndValues: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string,
+      filter: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
adamstankiewicz, line 296: Is this PropType definition accurate? Per the [react-table docs on filtering](https://react-table.tanstack.com/docs/api/useFilters), it looks like acceptable values for `filter` are either a string or a function. Not sure if `PropTypes.array` is applicable here?

adamstankiewicz (commented):
This seems like a sane approach to me. I just have a few nits to simplify some of the code, but it logically makes sense. Curious to hear @manny-m 's thoughts as well.

Also, what's your thinking for the remind case (in this PR, or done separately)?

christopappas:
@adamstankiewicz thanks for the review. these are good nits. figured there would be clever ways to simplify this stuff :)

I'll be doing the remind in this PR as well! stay tuned

manny-m (commented):
These changes make sense to me! Nice work!

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
christopappas: sorry those proptypes are suppose to be for filterType, not filter
Screen Shot 2022-01-07 at 10 00 58 AM

src/components/subscriptions/data/utils.js
```diff
@@ -16,7 +16,7 @@ export const getSubscriptionExpiringCookieName = ({
   return cookieName;
 };
 
-export const canRevokeLicense = (licenseStatus) => licenseStatus === ASSIGNED || licenseStatus === ACTIVATED;
+export const canRevokeLicense = (licenseStatus) => REVOCABLE_STATUSES.includes(licenseStatus);
```
adamstankiewicz, line 19: I like this change :)

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
```diff
@@ -31,27 +32,19 @@ const generateEmailTemplate = (contactEmail) => ({
...
-  // If we are not reminding all users, use users.length
-  // Else if totalToRemind is passed use that
-  // Else use 'All'
-  let buttonNumberLabel = 'All';
-  if (!remindAllUsers) {
-    buttonNumberLabel = userCount;
-  } else if (totalToRemind > 0) {
-    buttonNumberLabel = totalToRemind;
+const generateRemindModalSubmitLabel = (totalToRemind) => {
+  let buttonNumberLabel = '(all)';
```
manny-m, line 39: Small nit, don't feel strongly about it all but this use to be `'(All)'` instead of `'(all)'` this is also the case in the revoke modal. There are no mocks in this state so I think both are fine.

manny-m (approved):
Cool! 🚀  Tried this out on my machine and seems to work as intended! Left small nonblocking nit.

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
long74100: I believe for all our buttons only the first word should to be upper case. @adamstankiewicz

adamstankiewicz: @manny-m @long74100 I made the suggestion to change the casing to Long because typically our buttons only have the first letter capitalized. That said, when it does say "all", it should probably be without parentheses, e.g. "Revoke all" or "Remind all"

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -38,27 +39,21 @@ const showRevocationCapAlert = (revocationCapEnabled, revocations) => {
...
-  // Else use 'All'
-  let buttonNumberLabel = 'All';
-  if (!revokeAllUsers) {
-    buttonNumberLabel = userCount;
-  } else if (totalToRevoke > 0) {
-    buttonNumberLabel = totalToRevoke;
+const generateRevokeModalSubmitLabel = (totalToRevoke) => {
+  let buttonNumberLabel = '(all)';
+
+  if (totalToRevoke !== null && totalToRevoke !== undefined) {
```
adamstankiewicz, line 48: nit: Wonder if we could use `Number.isFinite(...)` here 🤔 

```jsx
Number.isFinite(null) // false
Number.isFinite(undefined) // false
Number.isFinite(0) // true
Number.isFinite(12) // true
```

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
```diff
@@ -31,27 +32,19 @@ const generateEmailTemplate = (contactEmail) => ({
...
+  if (totalToRemind !== null && totalToRemind !== undefined) {
```
adamstankiewicz, line 40: nit: Wonder if we could use `Number.isFinite(...)` here 🤔 

```jsx
Number.isFinite(null) // false
Number.isFinite(undefined) // false
Number.isFinite(0) // true
Number.isFinite(12) // true
```

src/components/subscriptions/licenses/LicenseManagementModals/tests/LicenseManagementRevokeModal.test.jsx
```diff
@@ -158,4 +159,73 @@ describe('<LicenseManagementRevokeModal />', () => {
       expect(screen.getByRole('alert')).toBeTruthy();
     });
   });
+
+  describe('calls the correct revoke endpoint', () => {
+    beforeEach(() => {
+      const mockPromiseResolve = Promise.resolve({ data: {} });
+      LicenseManagerApiService.licenseRevokeAll.mockReturnValue(mockPromiseResolve);
+      LicenseManagerApiService.licenseBulkRevoke.mockReturnValue(mockPromiseResolve);
```
adamstankiewicz, line 167: nit: this works, but could probably use the helper function `mockResolvedValue({ data: {} })` as well

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -181,6 +184,70 @@ const LicenseManagementTableBulkActions = ({
     );
   };
 
+  const getTotalToRevoke = () => {
```
adamstankiewicz, line 187: nit: This could likely utilize `useMemo` to only compute the value when necessary rather the re-calculate it on each render.

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -181,6 +184,70 @@ const LicenseManagementTableBulkActions = ({
...
+    // of revoked users from the table item count
+    const activeFilter = activeFilters[0];
+    if (activeFilter.name === 'statusBadge' && activeFilter.filterValue.includes(REVOKED)) {
+      return tableItemCount - revokedUsersCount;
+    }
+
+    return tableItemCount;
+  };
+
+  const getTotalToRemind = () => {
```
adamstankiewicz, line 213: nit: This could likely utilize `useMemo` to only compute the value when necessary rather the re-calculate it on each render.

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -198,19 +265,19 @@ const LicenseManagementTableBulkActions = ({
               variant="outline-danger"
               iconBefore={RemoveCircle}
               onClick={() => revokeOnClick(usersToRevoke)}
-              disabled={(!usersToRevoke.length && !allUsersSelected) || disabled}
+              disabled={(totalToRevoke === 0) || disabled}
             >
-              Revoke ({allUsersSelected ? activatedUsers + assignedUsers : usersToRevoke.length})
+              Revoke {totalToRevoke !== null ? `(${totalToRevoke})` : 'All'}
```
adamstankiewicz, line 270: [nit/clarification] Should this `All` be `all`?

src/components/subscriptions/licenses/LicenseManagementTable/LicenseManagementTableBulkActions.jsx
```diff
@@ -198,19 +265,19 @@ const LicenseManagementTableBulkActions = ({
...
         </ModalPopup>
         <Button
           variant="outline-primary"
           iconBefore={Email}
           onClick={() => remindOnClick(usersToRemind)}
-          disabled={(!usersToRemind.length && !allUsersSelected) || disabled}
+          disabled={(totalToRemind === 0) || disabled}
         >
-          Remind ({allUsersSelected ? assignedUsers : usersToRemind.length })
+          Remind {totalToRemind !== null ? `(${totalToRemind})` : 'All'}
```
adamstankiewicz, line 280: [nit/clarification] Should this `All` be `all`?

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRevokeModal.jsx
```diff
@@ -71,42 +66,50 @@ const LicenseManagementRevokeModal = ({
...
-  );
+  const buttonLabels = generateRevokeModalSubmitLabel(totalToRevoke);
 
-  const title = `Revoke License${revokeAllUsers || usersToRevoke.length > 1 ? 's' : ''}`;
+  const title = `Revoke License${revokeAllUsers || totalToRevoke > 1 ? 's' : ''}`;
 
   const isExpired = moment().isAfter(subscription.expirationDate);
 
-  const handleSubmit = () => {
+  const handleSubmit = async () => {
```
adamstankiewicz, line 79: nit: I wonder if this should use `useCallback` to only create the function definition when it needs to versus on every component re-render 🤔

src/components/subscriptions/licenses/LicenseManagementModals/LicenseManagementRemindModal.jsx
```diff
@@ -65,17 +58,18 @@ const LicenseManagementRemindModal = ({
...
   const isExpired = moment().isAfter(subscription.expirationDate);
 
-  const buttonLabels = generateRemindModalSubmitLabel(remindAllUsers, usersToRemind.length, totalToRemind);
+  const buttonLabels = generateRemindModalSubmitLabel(totalToRemind);
 
-  const title = `Remind User${remindAllUsers || usersToRemind.length > 1 ? 's' : ''}`;
+  const title = `Remind User${remindAllUsers || totalToRemind > 1 ? 's' : ''}`;
 
-  const handleSubmit = () => {
+  const handleSubmit = async () => {
```
adamstankiewicz, line 72: nit: I wonder if this should use `useCallback` to avoid re-creating the function definition on every component re-render.

src/components/subscriptions/licenses/LicenseManagementModals/tests/LicenseManagementRevokeModal.test.jsx
long74100: agree should be using mockResolvedValue here

adamstankiewicz (approved):
LGTM 🚀

PR #679, opened 2022-01-19
fix: don't need saml- prefix for metadata url

binodpant:
When forming metadata url, the tpa_hint parameter need not contain the saml- prefix, unlike what we do for the SSO link like test link.

This is because [the backend](https://github.com/openedx/edx-platform/blob/db32ff2cdf678fa8edd12c9da76a76eef0478614/common/djangoapps/third_party_auth/views.py#L60) simply use tpa_hint as the saml idp slug, and does not expect a saml- prefix.

## Testing notes

Added automated test for util function that generates the urls: which is used as such

```
const { testLink, spMetadataLink } = createSAMLURLs({
      configuration,
      idpSlug,
      enterpriseSlug,
      learnerPortalEnabled,
    });
```

After this change, verified in admin portal the metadata url works
Screen Shot 2022-01-19 at 12 22 24 PM

# For all changes


mattdrayer:
Does a test also need to be updated?

binodpant:
added tests!

mattdrayer (approved):

PR #688, opened 2022-01-28
feat: base LMS page

kiram15:
# For all changes

Screen Shot 2022-01-28 at 4 10 04 PM


src/components/settings/SettingsLMSTab/constants.js
```diff
@@ -0,0 +1,6 @@
+export const BLACKBOARD_ICON = 'https://uthsc.edu/tlc/images/bblogoc.png';
```
kiram15, line 1: I know this way is probably not advised, but couldn't really find any good examples of best practice of other's icons!

johnnagro (approved):

johnnagro:
looks good. would be great to convert those other PNGs into SVGs but, not a hard requirement imo.

PR #691, opened 2022-02-04
fix: rectify bug of parsing undefined in query string

muneebGH:
# For all changes

https://user-images.githubusercontent.com/49611774/152549289-8d837b52-ef9c-4cd2-b1e4-b0a7cbfab5ea.mov

### DIFF
--- a/src/components/EnterpriseList/index.jsx
+++ b/src/components/EnterpriseList/index.jsx
@@ -53,7 +53,7 @@ class EnterpriseList extends React.Component {
   fetchEnterprisesWithSearch = (options) => {
     const optionsWithSearch = {
       ...options,
-      search: this.state.searchQuery || undefined,
+      ...(this.state.searchQuery && { search: this.state.searchQuery }),
     };
     return LmsApiService.fetchEnterpriseList(optionsWithSearch);
   }
--- a/src/components/EnterpriseList/index.jsx
+++ b/src/components/EnterpriseList/index.jsx
@@ -53,8 +53,10 @@ class EnterpriseList extends React.Component {
   fetchEnterprisesWithSearch = (options) => {
     const optionsWithSearch = {
       ...options,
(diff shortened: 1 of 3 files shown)


johnnagro:
any idea why the tests are failing?

muneebGH:
@binodpant this is the issue
Screenshot 2022-02-04 at 8 20 41 PM

src/components/EnterpriseList/index.jsx
```diff
@@ -53,7 +53,7 @@ class EnterpriseList extends React.Component {
   fetchEnterprisesWithSearch = (options) => {
     const optionsWithSearch = {
       ...options,
-      search: this.state.searchQuery || undefined,
+      ...(this.state.searchQuery && { search: this.state.searchQuery }),
```
adamstankiewicz, line 56: Makes sense to me, though, (optional nit) I wonder if the following is slightly more readable as:

```jsx
const optionsWithSearch = { ...options };
if (this.state.searchQuery) {
  optionsWithSearch.search = this.state.searchQuery;
}
```

adamstankiewicz (approved):

src/components/EnterpriseList/index.jsx
binodpant: ah yeah I see now, this is returning undefined for the search value so simply avoiding it in the object is good.

optional:
if you wanted to harden the function fetchEnterprisesList in case of undefined, you can also do sth like this there (optionally, since this won't happen after your change. I just don't think functions should assume correctness of their inputs generally)

```
const { search, ...restopts} = options
const finalOpts = search ? { ...options } : restopts
```

or sth close to this

binodpant (approved):
lgtm

johnnagro (approved):

src/components/EnterpriseList/index.jsx
muneebGH: done

PR #700, opened 2022-02-16
feat: adding lms config validation

kiram15:
# For all changes

Screen Shot 2022-02-16 at 10 42 30 AM
Screen Shot 2022-02-16 at 10 42 17 AM

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
@@ -6,10 +6,11 @@ import LmsApiService from '../../../../data/services/LmsApiService';
 import { snakeCaseDict, urlValidation } from '../../../../utils';
 import ConfigError from '../ConfigError';
 import ConfigModal from '../ConfigModal';
-
-import { SUCCESS_LABEL } from '../../data/constants';
+import { INVALID_LINK, INVALID_NAME, SUCCESS_LABEL } from '../../data/constants';
 
 const BlackboardConfig = ({ id, onClick }) => {
+  const [displayName, setDisplayName] = React.useState('');
+  const [nameValid, setNameValid] = React.useState(true);
   const [clientId, setClientId] = React.useState('');
   const [clientSecret, setClientSecret] = React.useState('');
   const [blackboardBaseUrl, setBlackboardBaseUrl] = React.useState('');
@@ -18,6 +19,7 @@ const BlackboardConfig = ({ id, onClick }) => {
   const [modalIsOpen, openModal, closeModal] = useToggle(false);
 
   const config = {
+    displayName,
     clientId,
     clientSecret,
     blackboardBaseUrl,
@@ -41,10 +43,19 @@ const BlackboardConfig = ({ id, onClick }) => {
     }
   };
 
-  const validateUrl = (input) => {
-    setBlackboardBaseUrl(input);
-    const urlBool = (urlValidation(input) || input.length === 0);
-    setUrlValid(urlBool);
+  const validateField = (field, input) => {
+    switch (field) {
+      case 'Canvas Base URL':
+        setBlackboardBaseUrl(input);
+        setUrlValid(urlValidation(input) || input.length === 0);
+        break;
+      case 'Display Name':
+        setDisplayName(input);
+        setNameValid(input.length <= 20);
+        break;
+      default:
+        break;
+    }
   };
 
   return (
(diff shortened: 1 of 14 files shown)


johnnagro (approved):

src/utils.js
```diff
@@ -226,6 +226,11 @@ function truncateString(str, maxStrLength = 10) {
   return `${str.slice(0, maxStrLength)}...`;
 }
 
+function urlValidation(url) {
+  const regex = new RegExp('(https?://)?([\\da-z.-]+)\\.([a-z.]{2,6})[/\\w .-]*/?');
```
johnnagro, line 230: its not a deal breaker but sometimes using a system library does a better job than a regex, example: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url

src/utils.js
```diff
@@ -226,6 +226,11 @@ function truncateString(str, maxStrLength = 10) {
   return `${str.slice(0, maxStrLength)}...`;
 }
 
+function urlValidation(url) {
```
binodpant, line 229: will be nice to have tests for various urls here esp. since it looks like we have a slightly custom url validation than what we may get with, say the URL() constructor (like requiring https which is awesome!) ? https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

src/components/settings/SettingsLMSTab/LMSConfigs/DegreedConfig.jsx
```diff
@@ -13,6 +13,7 @@ const DegreedConfig = ({ id, onClick }) => {
   const [secret, setSecret] = React.useState('');
   const [degreedCompanyId, setDegreedCompanyId] = React.useState('');
   const [degreedBaseUrl, setDegreedBaseUrl] = React.useState('');
+  const [urlValid, setUrlValid] = React.useState(true);
```
binodpant, line 16: do we want to default to false, assuming initially url field's value will be empty?

src/components/settings/SettingsLMSTab/LMSConfigs/DegreedConfig.jsx
```diff
@@ -45,6 +46,12 @@ const DegreedConfig = ({ id, onClick }) => {
     }
   };
 
+  const validateUrl = (input) => {
```
binodpant, line 49: since this idea of the urlValid, setUrlValid is common across all the various config JSX files, fun idea if you are interested (optional) is to abstract out the common idea of a valid url, you could have sth like this 

so for ex

```src/components/settings/SettingsLMSTab/hooks.jsx```

could export a function like this that exports this state, 

```
function useValidatedURL(inputURL) {
    const [urlValid, setUrlValid] = React.useState(false);
    const isURLValid = useMemo(() => {
        const urlBool = (urlValidation(input) || input.length === 0);
        setUrlValid(urlBool);
        return isURLValid; // you may not need this return val anywhere so you can skip if you like
    }, inputURL); // only re evals setUrlValid if the inputURL is different from before
    return [urlValid, setUrlVal];
}
```

then in all usage code you could use

```
const [urlValid, setUrlValid] = useValidatedURL(degreedBaseUrl);
const validateUrl = input => { setDegreedBaseUrl(input); /* now urlValid should automatically update when degreedBaseURL changes */ }
```

not much of a refactor but I think it lends itself to a bit cleaner setup if you want to explore the option

the one future advantage is if we change the way we manage that state (such as calling other hooks as side effects), a change only needs to be made inside the use hook

File: src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx [hunk 5]
```diff
@@ -59,26 +66,31 @@ const BlackboardConfig = ({ id, onClick }) => {
...
+            isInvalid={!urlValid}
             onChange={(e) => {
-              setBlackboardBaseUrl(e.target.value);
+              validateUrl(e.target.value);
             }}
             floatingLabel="Blackboard Base URL"
           />
+          {!urlValid && (
+            <Form.Control.Feedback type="invalid">
+              This does not look like a valid url
```
binodpant, line 87: suggested: can we have a slightly more informative message (may need to run by UX), but sth like

'URLs needs start with https, have alphanumeric chars etc etc..' whatever makes most sense.

this will also be similar to the other message we have for validation which looks great as in `Display name cannot be over 30 characters`

src/utils.js
binodpant: could somethine like this work? idk if https is our main requirements, plus being a valid URI then this one should also work https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

```
url = 'http://test'
'http://test'


try {
    new URL(url); if (url.startsWith('https') === false) { throw new TypeError('failed since not https'); }
} catch(e) { console.error('failed with error ' , e) }
VM1144:3 failed with error  TypeError: failed since not https
    at <anonymous>:2:66
```

binodpant: but if there are other requirements beyond met by above, then yeah you may have to check

binodpant: new URL() will catch 'basic' error in urls, and the second check catches http vs https

src/components/settings/SettingsLMSTab/LMSConfigs/DegreedConfig.jsx
kiram15: No, because further on with draft functionality we will allow blank fields, also if its set to false before the user even clicks on it, it will be red so it seemed aggressive

binodpant: I see, ok so we start clean and I did forget about the draft ability, so yeah this is correct!

PR #705, opened 2022-02-18
Updates to error handling

kiram15:
# For all changes

Screen Shot 2022-02-22 at 10 35 16 PM


src/components/settings/SettingsLMSTab/LMSConfigPage.jsx
```diff
@@ -35,7 +35,7 @@ export function handleErrors(error) {
     ? error.message
     : JSON.stringify(error.response.data);
   logError(errorMsg);
-  return errorMsg;
+  return error.response?.status;
```
johnnagro, line 38: there are probably situations where there is no response - perhaps a timeout, dropped connection, browser extension blocking. will consumers of `handleError` be okay if this returns a `null`?

johnnagro (approved):

PR #712, opened 2022-02-27
feat: LMS config settings page checks for IDP configs

alex-sheehan-edx:
image

NEEDED BACKEND PR/BRANCH IN ORDER TO TEST:
https://github.com/openedx/edx-enterprise/pull/1465

This PR adds:
- a new LMS api service fetch method to a new backend endpoint that returns if a customer still needs to set up an IDP config.
- If the api fetch method returns `True`, 
    - renders an alert modal at the top of the LMS config settings page that has a redirect to the SSO settings tab
    - disables creation cards and display creation cards button on the LMS config page

# For all changes

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSCard.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSCard.jsx
@@ -3,11 +3,11 @@ import PropTypes from 'prop-types';
 import { Card, Image, Stack } from '@edx/paragon';
 import { channelMapping } from '../../../utils';
 
-const LMSCard = ({ LMSType, onClick }) => (
+const LMSCard = ({ LMSType, onClick, disabled }) => (
   <Card
-    isClickable
-    className="pb-4"
-    onClick={() => onClick(LMSType)}
+    isClickable={!disabled}
+    className={`pb-4 ${disabled ? 'opacity-50' : ''}`}
+    onClick={() => !disabled && onClick(LMSType)}
   >
     <Card.Header
       title={(
@@ -23,5 +23,6 @@ const LMSCard = ({ LMSType, onClick }) => (
 LMSCard.propTypes = {
   LMSType: PropTypes.string.isRequired,
   onClick: PropTypes.func.isRequired,
+  disabled: PropTypes.bool.isRequired,
 };
 export default LMSCard;
--- a/src/components/settings/SettingsLMSTab/index.jsx
+++ b/src/components/settings/SettingsLMSTab/index.jsx
@@ -1,9 +1,11 @@
 import React, { useEffect, useState } from 'react';
 import {
-  Button, Hyperlink, CardGrid, Toast,
+  Alert, Button, Hyperlink, CardGrid, Toast,
 } from '@edx/paragon';
-import { Add } from '@edx/paragon/icons';
+import { Link } from 'react-router-dom';
+import { Add, Info } from '@edx/paragon/icons';
 import { logError } from '@edx/frontend-platform/logging';
+import PropTypes from 'prop-types';
 
 import { camelCaseDictArray } from '../../../utils';
 import LMSCard from './LMSCard';
@@ -27,7 +29,12 @@ const SUBMIT_TOAST_MESSAGE = 'Configuration was submitted successfully.';
 const TOGGLE_TOAST_MESSAGE = 'Configuration was toggled successfully.';
 const DELETE_TOAST_MESSAGE = 'Configuration was successfully removed.';
 
-export default function SettingsLMSTab(enterpriseId) {
+export default function SettingsLMSTab({
+  enterpriseId,
+  enterpriseSlug,
+  enableSamlConfigurationScreen,
+  identityProvider,
+}) {
   const [config, setConfig] = useState();
   const [showToast, setShowToast] = useState(false);
 
@@ -37,6 +44,7 @@ export default function SettingsLMSTab(enterpriseId) {
 
   const [existingConfigFormData, setExistingConfigFormData] = useState({});
   const [toastMessage, setToastMessage] = useState();
+  const [displayNeedsSSOAlert, setDisplayNeedsSSOAlert] = useState(false);
 
   // onClick function for existing config cards' edit action
   const editExistingConfig = (configData, configType) => {
@@ -51,7 +59,7 @@ export default function SettingsLMSTab(enterpriseId) {
   };
 
   const fetchExistingConfigs = () => {
-    const options = { enterprise_customer: enterpriseId.enterpriseId };
+    const options = { enterprise_customer: enterpriseId };
     LmsApiService.fetchEnterpriseCustomerIntegrationConfigs(options)
       .then((response) => {
         // If the enterprise has existing configs
@@ -113,6 +121,11 @@ export default function SettingsLMSTab(enterpriseId) {
     fetchExistingConfigs();
   }, []);
 
+  useEffect(() => {
+    // Check if the customer needs an idp config
+    setDisplayNeedsSSOAlert(enableSamlConfigurationScreen && !identityProvider);
+  }, [enableSamlConfigurationScreen, identityProvider]);
+
   return (
     <div>
       <div className="d-flex">
@@ -129,13 +142,30 @@ export default function SettingsLMSTab(enterpriseId) {
         Enabling a learning management system for your edX account allows quick
         access to the catalog
       </p>
+      {displayNeedsSSOAlert && (
+        <Alert
+          className="mr-6 sso-alert-modal-margin"
+          variant="danger"
+          icon={Info}
+          actions={[
+            <Link to={`/${enterpriseSlug}/admin/settings/sso`}>
+              <Button>Configure SSO</Button>
+            </Link>,
+          ]}
+        >
+          <Alert.Heading>No SSO configured</Alert.Heading>
+          <p>
+            At least one Single Sign On configuration is needed to create an LMS configuration
+          </p>
+        </Alert>
+      )}
       {configsExist && (
         <span>
           <h4 className="mt-5 mb-4">Existing configurations</h4>
           <ExistingLMSCardDeck
             configData={existingConfigsData}
             editExistingConfig={editExistingConfig}
-            enterpriseCustomerUuid={enterpriseId.enterpriseId}
+            enterpriseCustomerUuid={enterpriseId}
             onClick={onClick}
           />
         </span>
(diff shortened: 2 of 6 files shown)


kiram15 (commented):
Cards are clickable even if SSO is not configured - Alex is fixing now

src/components/settings/SettingsLMSTab/LMSCard.jsx
```diff
@@ -3,11 +3,15 @@ import PropTypes from 'prop-types';
...
 
-const LMSCard = ({ LMSType, onClick }) => (
+const LMSCard = ({ LMSType, onClick, disabled }) => (
   <Card
-    isClickable
-    className="pb-4"
-    onClick={() => onClick(LMSType)}
+    isClickable={!disabled}
+    className={`pb-4 ${disabled ? 'lms-card-disabled-opacity' : ''}`}
+    onClick={() => {
```
binodpant, line 10: optional: if you wish, a slight bit neater would be perhaps sth like 

```
onClick={ () => !disabled && onClick(LMSType)  }
```

src/components/settings/SettingsLMSTab/LMSCard.jsx
```diff
@@ -3,11 +3,15 @@ import PropTypes from 'prop-types';
...
 import { channelMapping } from '../../../utils';
 
-const LMSCard = ({ LMSType, onClick }) => (
+const LMSCard = ({ LMSType, onClick, disabled }) => (
   <Card
-    isClickable
-    className="pb-4"
-    onClick={() => onClick(LMSType)}
+    isClickable={!disabled}
+    className={`pb-4 ${disabled ? 'lms-card-disabled-opacity' : ''}`}
```
binodpant, line 9: for opacity, is it required to have a custom class? in general it's best to use existing util classes if possible, for example, do any of these bootstrap utils work? https://getbootstrap.com/docs/5.1/utilities/opacity/

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -70,6 +73,17 @@ export default function SettingsLMSTab(enterpriseId) {
       });
   };
 
+  const checkIfCustomerNeedsIDPConfig = () => {
```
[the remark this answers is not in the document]
binodpant: if you do use it, you may need to add it explicitly here to the field added to the `state.portalConfiguration`

https://github.com/edx/frontend-app-admin-portal/blob/7b3605d05993842a45cc20059c4fe24ad4872249/src/data/reducers/portalConfiguration.js

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -111,6 +125,7 @@ export default function SettingsLMSTab(enterpriseId) {
   useEffect(() => {
     // On load fetch potential existing configs
     fetchExistingConfigs();
+    checkIfCustomerNeedsIDPConfig();
```
binodpant, line 128: I also think you can directly populate this value as a boolean in the state.portalConfiguration and then just read it in this component just like the enterpriseId, it will also avoid the extra re-render you will have with this check.

The only disadvantage of this method is that you will need to reload page if the enterprise customer state is changed while the UI is being used. I don't know if that is an issue at all honestly but if it is, you can have that portalConfiguration be refetched instead ... to apply the change (if this is even an issue). Arguably that issue exists for all values used in the enterprise configuration anyway

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -129,13 +144,30 @@ export default function SettingsLMSTab(enterpriseId) {
...
+          icon={Info}
+          actions={[
+            <Link to={`/${enterpriseSlug}/admin/settings/sso`}>
+              <Button>Configure SSO</Button>
+            </Link>,
+          ]}
+        >
+          <Alert.Heading>No SSO configured</Alert.Heading>
+          <p>
+            At least one Single Sign On configuration is needed to create an LMS configuration
```
binodpant, line 160: Did we want to block LMS creation on SSO? I think we simply want to warn the user that they need to create an SSO config (we can confirm with Alena but I am pretty sure we wanted to not disallow LMS config creation, just that we want to let users know SSO is also going to be needed?) Sorry if any confusions

src/components/settings/settings.scss
```diff
@@ -48,3 +52,7 @@
 .existing-lms-card-width {
   width: 70rem !important;
 }
+
+.lms-card-disabled-opacity {
```
binodpant, line 56: see above, but potentially use builtin bootstrap opacity classes?

src/components/settings/SettingsLMSTab/tests/LmsConfigPage.test.jsx
```diff
@@ -217,6 +225,33 @@ describe('<SettingsLMSTab />', () => {
...
+    fireEvent.click(configureSSOButton);
+    expect(history.location.pathname).toEqual(`/${enterpriseSlug}/admin/settings/sso`);
+  });
+  test('Expected behavior when customer has IDP configured', async () => {
+    const expectedApiResponse = {
+      data: { needs_idp_config: false },
+    };
+    mockNeedsIDPConfig.mockImplementation(() => (Promise.resolve(expectedApiResponse)));
+    renderWithRouter(<SettingsLMSWrapper />);
+    expect(screen.queryByText('No SSO configured')).toBeFalsy();
```
binodpant, line 253: I would recommend using the jest-dom extended matchers such as `.not.toBeInTheDocument()` for verifying dom state, since there are cases wheren things don't render and yet toBeTruthy and toBeFalsy can be still true/false

src/components/settings/SettingsLMSTab/tests/LmsConfigPage.test.jsx
```diff
@@ -217,6 +225,33 @@ describe('<SettingsLMSTab />', () => {
...
+      <Router history={history}>
+        <Provider store={mockStore({ ...initialState })}>
+          <SettingsLMSTab enterpriseId={enterpriseId} enterpriseSlug={enterpriseSlug} />
+        </Provider>
+      </Router>,
+    );
+    expect(history.location.pathname).toEqual('/');
+    await screen.findByText('Configure SSO');
+    const configureSSOButton = screen.getByText('Configure SSO');
+    fireEvent.click(configureSSOButton);
```
binodpant, line 244: please do ensure you are not getting the act wranings in these tests, I had to use `findByText` and await using act to avoid those warnings, 

here is a PR for your ref, if that is happening at all for you: 

https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/199

src/components/settings/SettingsLMSTab/tests/LmsConfigPage.test.jsx
```diff
@@ -217,6 +225,33 @@ describe('<SettingsLMSTab />', () => {
     expect(screen.queryByText('Exit without saving')).toBeFalsy();
     expect(screen.queryByText('Connect SAP')).toBeFalsy();
   });
+  test('Expected behavior when customer has no IDP configured', async () => {
+    const expectedApiResponse = {
+      data: { needs_idp_config: true },
+    };
+    mockNeedsIDPConfig.mockImplementation(() => (Promise.resolve(expectedApiResponse)));
```
binodpant, line 232: nit: 

```
mockNeedsIDPConfig.mockImplementation(() => Promise.resolve(expectedApiResponse));
```

src/components/settings/SettingsLMSTab/tests/LmsConfigPage.test.jsx
alex-sheehan-edx: I'm finding no act warnings as I run tests!

src/components/settings/SettingsLMSTab/tests/LmsConfigPage.test.jsx
alex-sheehan-edx: I was just following conventions found in the admin portal! I can easily change.

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -111,6 +119,8 @@ export default function SettingsLMSTab(enterpriseId) {
   useEffect(() => {
     // On load fetch potential existing configs
     fetchExistingConfigs();
+    // Check if the customer needs an idp config
+    setDisplayNeedsSSOAlert(enableSamlConfigurationScreen && !identityProvider);
```
binodpant, line 123: hmm.. I wonder how we should use this flag `enableSamlConfigurationScreen` in this section of the UI. looks like prior SAML screen is completely hidden when this flag is false. For purposes of LMS settings UI, does this flag even matter?

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -111,6 +119,8 @@ export default function SettingsLMSTab(enterpriseId) {
   useEffect(() => {
     // On load fetch potential existing configs
     fetchExistingConfigs();
+    // Check if the customer needs an idp config
```
binodpant, line 122: I think it's better to have this set in a separate useEffect of its own. That is because this has a dependency on the values like this, whereas the original useEffect should only ever run once when this component first renders

so I would suggest a separate useEffect like this

```
useEffect(() => {
    // Check if the customer needs an idp config
    setDisplayNeedsSSOAlert(enableSamlConfigurationScreen && !identityProvider);
}, [enableSamlConfigurationScreen, identityProvider] );
```

binodpant (approved):
one small clarification asked, and one change suggested. Otherwise looking great!

src/components/settings/SettingsLMSTab/index.jsx
binodpant: an alternative way to achieve this, perhaps a bit cleaner even, is 

```
const displayNeedsSSOAlert = useMemo( () => enableSamlConfigurationScreen && !identityProvider, [enableSamlConfigurationScreen, identityProvider]);
```

then you won't need the set at all, it's just a regular constant passed down to other components as props. but either option is fine!

src/components/settings/SettingsLMSTab/index.jsx
alex-sheehan-edx: I talked with product as well as ECS and because there are instances where the user/customer doesn't have/need an idp but still wants an LMS config, we only want to force them to make one if they have the screen enabled.

binodpant: cool! ty

binodpant (approved):

PR #717, opened 2022-03-03
Only pass page option if param exists

manny-m:
### [Cloeses ENT-5513](https://openedx.atlassian.net/browse/ENT-5513)

No longer passing `page: null` in the options for `fetchCodeSearchResults`.


long74100:
we should probably do a quick scan to see if we need the same thing somewhere else after.

PR #718, opened 2022-03-04
"Manage Learners/Requests" tabs on Subs Mgmt and Codes Mgmt pages with SubsidyRequestTable

adamstankiewicz:
# Description

### Subscription Management

image

image

image

image

### Code Management

image

image

image

image

# For all changes


src/components/CodeManagement/ManageCodesTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+import { CheckCircle, Info, WarningFilled } from '@edx/paragon/icons';
+
+import SearchBar from '../SearchBar';
+import CodeSearchResults from '../CodeSearchResults';
+import LoadingMessage from '../LoadingMessage';
+import Coupon from '../Coupon';
+import { updateUrl } from '../../utils';
+import { fetchCouponOrders, clearCouponOrders } from '../../data/actions/coupons';
+
+class ManageCodesTab extends React.Component {
```
adamstankiewicz, line 21: Largely a copy/paste from existing `CodeManagement` component.

src/components/CodeManagement/ManageCodesTab.jsx
```diff
@@ -0,0 +1,361 @@
...
+            paginationLabel="coupons pagination"
+          />
+        </div>
+      </>
+    );
+  }
+
+  renderRequestCodesSuccessMessage() {
+    return (
+      <Alert variant="success" icon={CheckCircle} dismissible>
```
adamstankiewicz, line 210: The `StatusAlert` components were swapped for `Alert`, as `StatusAlert` component is deprecated in Paragon.

src/components/BulkEnrollmentPage/stepper/DismissibleCourseWarning.jsx
```diff
@@ -8,16 +8,14 @@ import { WARNING_ALERT_TITLE_TEXT, WARNING_ALERT_BODY_TEXT } from './constants';
  * Displays simple dismissible banner to remind admins.
  */
 const DismissibleCourseWarning = () => {
-  const [closed, setClosed] = useState(false);
-
-  if (closed) { return null; }
-
+  const [isOpen, setIsOpen] = useState(true);
   return (
     <Alert
```
adamstankiewicz, line 13: I realized some of our dismissible `Alert` components are not actually dismissible, such as this one. The `dismissible`  prop only controls whether the "Dismiss" button is shown. The actual dismiss behavior is controlled by the `show` and `onClose` props and some state.

I fixed it here, and added an item to the tech debt tracker.

src/components/CodeManagement/CodeManagementRoutes.jsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+    setTabKey(key);
+  };
+
+  if (isTabsFeatureEnabled) {
+    return (
+      <Switch>
+        <Tabs
+          id="tabs-code-management"
+          activeKey={tabKey}
```
manny-m, line 55: I would consider having a single source of truth for the state of `activeKey`. I would recommend not using `useState` but instead using `useParams`. In [the settings page](https://github.com/openedx/frontend-app-admin-portal/blob/7a8a3174911f6b4b2d0e75e1478c149593cd125e/src/components/settings/SettingsTabs.jsx#L25) we actually have a simple custom hook that does this for us.

I think doing something like that will make this simpler and remove the need to use `useEffect`.

src/components/SubsidyRequestManagementTable/data/hooks.js
```diff
@@ -1,10 +1,125 @@
...
+import debounce from 'lodash.debounce';
 
+import { transformRequestOverview, transformRequests } from './utils';
 import DiscoveryApiService from '../../../data/services/DiscoveryApiService';
 
-// eslint-disable-next-line import/prefer-default-export
+export const DEBOUNCE_TIME_MS = 200;
+export const PAGE_SIZE = 20;
+
+export const useSubsidyRequests = (
```
manny-m, line 12: Nice code reuse here 🏅 

Don't feel strongly about this but wonder if maybe we should pass some sort of enum to map to services? For example: 
```
const SUBSIDY_REQUESTS_TYPES = {
    'CODES': {
        overview: /* service reference here
        requests: /* service reference here
    },
    'LICENSES': {
        overview: /* service reference here
        requests: /* service reference here
    },
}
```
That way we only pass the enums `LICENSES` or `CODES` to `useSubsidyRequests` ensuring we are calling the right combination of services.

src/components/subscriptions/SubscriptionRoutes.jsx
```diff
@@ -0,0 +1,141 @@
...
+  const isTabsFeatureEnabled = !!features.FEATURE_BROWSE_AND_REQUEST;
+
+  const routesByTabKey = {
+    requests: `/${enterpriseSlug}/admin/${ROUTE_NAMES.subscriptionManagement}/requests`,
+    default: `/${enterpriseSlug}/admin/${ROUTE_NAMES.subscriptionManagement}`,
+  };
+  const history = useHistory();
+  const requestsTabMatch = useRouteMatch(routesByTabKey.requests);
+  const initialTabKey = requestsTabMatch ? 'requests' : 'default';
+  const [tabKey, setTabKey] = useState(initialTabKey);
```
manny-m, line 75: See comment on `CodeManagementRoutes` that allows us to remove `useSate` and `useEffect` and simplify this file.

src/components/CodeManagement/CodeManagementRoutes.jsx
```diff
@@ -0,0 +1,101 @@
...
+} from 'react-router-dom';
+import { Tabs, Tab } from '@edx/paragon';
+
+import ManageCodesTab from './ManageCodesTab';
+import ManageRequestsTab from './ManageRequestsTab';
+import { ROUTE_NAMES } from '../EnterpriseApp/constants';
+import { features } from '../../config';
+
+const CodeManagementRoutes = ({ enterpriseSlug }) => {
+  const isTabsFeatureEnabled = !!features.FEATURE_BROWSE_AND_REQUEST;
```
manny-m, line 18: Should we also check if this has been enabled by the user on the settings page?

manny-m (commented):
Awesome!

Left comments, specifically would love `CodeManagementRoutes` / `SubscriptionRoutes`

src/components/CodeManagement/CodeManagementRoutes.jsx
adamstankiewicz: @manny-m Good suggestion, updated. Originally didn't go with this approach because the existing routing structure of subscription management wasn't quite the same as the settings page, and I was originally trying to avoid creating a new URL (e.g., /subscriptions -> /subscriptions/manage-learners).

That said, I agree with your feedback and perhaps this is a decision we should document for this repo. That is, when implementing tabs, each tab should have its own key param for a route so we're consistent with tabs throughout the application. ADR worthy?

src/components/subscriptions/SubscriptionRoutes.jsx
adamstankiewicz: Updated!

PR #720, opened 2022-03-07
feat: add incomplete card hover

kiram15:
# For all changes

Screen Shot 2022-03-07 at 9 07 11 AM

### DIFF
--- a/src/components/settings/SettingsLMSTab/tests/ExistingLMSCardDeck.test.jsx
+++ b/src/components/settings/SettingsLMSTab/tests/ExistingLMSCardDeck.test.jsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import {
+  act,
   fireEvent,
-  render, screen,
+  render,
+  screen,
 } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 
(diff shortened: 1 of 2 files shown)


src/components/settings/SettingsLMSTab/tests/ExistingLMSCardDeck.test.jsx
```diff
@@ -67,6 +69,11 @@ describe('<ExistingLMSCardDeck />', () => {
     );
     expect(screen.getByText('Incomplete')).toBeInTheDocument();
     expect(screen.getByText('barfoo')).toBeInTheDocument();
+    act(() => {
```
binodpant, line 72: I don't think we require act around fireEvent calls, since their api already includes act

https://testing-library.com/docs/preact-testing-library/api/#fireevent

binodpant, line 72: I believe if you use the user-events library instead you will not need to use act() here (that may also be true of the built-in fireEvent but I am not sure). I have preferred using user-events when I can because it simluated the user events a bit more accurately wrt what happens in the browser

https://testing-library.com/docs/ecosystem-user-event#hoverelement-options

binodpant: we use it here, for example: https://github.com/edx/frontend-app-enterprise-public-catalog/blob/803d46471c857587a78eebd13cbb2f295169e22b/src/components/catalogSearchResults/buttons/downloadCsvButton/DownloadCsvButton.test.jsx

alex-sheehan-edx (approved):
one more open ended question about formatting but everything looks good here 👍

PR #721, opened 2022-03-07
feat: adding oauth authorization step to lms configs

alex-sheehan-edx:
# For all changes

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
@@ -1,41 +1,84 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Button, Form, useToggle } from '@edx/paragon';
+import { CheckCircle, Error } from '@edx/paragon/icons';
 import isEmpty from 'lodash/isEmpty';
 import { buttonBool, handleErrors } from '../utils';
 
 import LmsApiService from '../../../../data/services/LmsApiService';
 import { snakeCaseDict, urlValidation } from '../../../../utils';
 import ConfigError from '../ConfigError';
+import { useTimeout, useInterval } from '../../../../data/hooks';
 import ConfigModal from '../ConfigModal';
-import { INVALID_LINK, INVALID_NAME, SUCCESS_LABEL } from '../../data/constants';
+import {
+  BLACKBOARD_OAUTH_REDIRECT_URL,
+  INVALID_LINK,
+  INVALID_NAME,
+  SUCCESS_LABEL,
+  LMS_CONFIG_OAUTH_POLLING_INTERVAL,
+  LMS_CONFIG_OAUTH_POLLING_TIMEOUT,
+} from '../../data/constants';
 
 const BlackboardConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => {
   const [displayName, setDisplayName] = React.useState('');
   const [nameValid, setNameValid] = React.useState(true);
-  const [clientId, setClientId] = React.useState('');
-  const [clientSecret, setClientSecret] = React.useState('');
   const [blackboardBaseUrl, setBlackboardBaseUrl] = React.useState('');
   const [urlValid, setUrlValid] = React.useState(true);
   const [errorIsOpen, openError, closeError] = useToggle(false);
   const [modalIsOpen, openModal, closeModal] = useToggle(false);
   const [errCode, setErrCode] = React.useState();
   const [edited, setEdited] = React.useState(false);
-
+  const [authorized, setAuthorized] = React.useState(false);
+  const [oauthPollingInterval, setOauthPollingInterval] = React.useState(null);
+  const [oauthPollingTimeout, setOauthPollingTimeout] = React.useState(null);
+  const [oauthTimeout, setOauthTimeout] = React.useState(false);
+  const [configId, setConfigId] = React.useState();
   const config = {
     displayName,
-    clientId,
-    clientSecret,
     blackboardBaseUrl,
   };
 
+  // Polling method to determine if the user has authorized their config
+  useInterval(async () => {
+    if (configId) {
+      let err;
+      try {
+        const response = await LmsApiService.fetchSingleBlackboardConfig(configId);
+        if (response.data.refresh_token) {
+          // Config has been authorized
+          setAuthorized(true);
+          // Stop both the backend polling and the timeout timer
+          setOauthPollingInterval(null);
+          setOauthPollingTimeout(null);
+          setOauthTimeout(false);
+        }
+      } catch (error) {
+        err = handleErrors(error);
+      }
+      if (err) {
+        setErrCode(errCode);
+        openError();
+      }
+    }
+  }, oauthPollingInterval);
+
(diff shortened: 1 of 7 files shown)


src/components/settings/data/hooks.js
```diff
@@ -109,3 +109,45 @@ export default {
   useCurrentSettingsTab,
   useLinkManagement,
 };
+
+export function useInterval(callback, delay) {
```
binodpant, line 113: think it's best to place these in a morte general place outside settings

src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
```diff
@@ -44,19 +87,93 @@ const BlackboardConfig = ({ enterpriseCustomerUuid, onClick, existingData }) =>
...
+        // Save the config ID so we know one was created in the authorization flow
+        setConfigId(fetchedConfigId);
+        // Reset config polling timeout flag
+        setOauthTimeout(false);
+        // Start the config polling
+        setOauthPollingInterval(LMS_CONFIG_OAUTH_POLLING_INTERVAL);
+        // Start the polling timeout timer
+        setOauthPollingTimeout(LMS_CONFIG_OAUTH_POLLING_TIMEOUT);
+        // Open the oauth window for the user
+        const oauthUrl = `${blackboardBaseUrl}/learn/api/public/v1/oauth2/authorizationcode?`
```
binodpant, line 151: not for today would be great toi get this url from backend

src/components/settings/SettingsLMSTab/tests/BlackboardConfig.test.jsx
```diff
@@ -76,97 +94,145 @@ describe('<BlackboardConfig />', () => {
...
+      fireEvent.change(screen.getByLabelText('Display Name'), {
+        target: { value: 'displayName' },
+      });
+      fireEvent.change(screen.getByLabelText('Blackboard Base URL'), {
+        target: { value: 'https://www.test.com' },
+      });
+    });
+    expect(screen.getByText('Authorize')).not.toBeDisabled();
+    await act(async () => {
+      fireEvent.click(screen.getByText('Authorize'));
```
binodpant, line 134: I think using user-events here instead is better, also probably don't need act around the click

src/components/settings/SettingsLMSTab/tests/BlackboardConfig.test.jsx
```diff
@@ -3,18 +3,44 @@ import {
   render, fireEvent, screen,
 } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
+import { act } from 'react-dom/test-utils';
```
binodpant, line 6: I think you want to import `act` from testing-library/react instead

src/components/settings/SettingsLMSTab/tests/BlackboardConfig.test.jsx
```diff
@@ -3,18 +3,44 @@ import {
...
 import BlackboardConfig from '../LMSConfigs/BlackboardConfig';
 import { INVALID_LINK, INVALID_NAME } from '../../data/constants';
 import LmsApiService from '../../../../data/services/LmsApiService';
 
-jest.mock('../../../../data/services/LmsApiService');
+jest.mock('../../data/constants', () => ({
+  ...jest.requireActual('../../data/constants'),
+  LMS_CONFIG_OAUTH_POLLING_INTERVAL: 1,
+}));
+window.open = jest.fn();
```
binodpant, line 16: would to good to spy on this or sth so you can verify we do window.open when expected with correct url

src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
alex-sheehan-edx: agreed, probably in a follow up?

src/components/settings/SettingsLMSTab/tests/CanvasConfig.test.jsx
```diff
@@ -148,24 +184,120 @@ describe('<CanvasConfig />', () => {
...
+    act(() => {
+      fireEvent.change(screen.getByLabelText('Display Name'), {
+        target: { value: 'displayName' },
+      });
+      fireEvent.change(screen.getByLabelText('Canvas Base URL'), {
+        target: { value: 'https://www.test.com' },
+      });
+    });
+    expect(screen.getByText('Authorize')).not.toBeDisabled();
+    await act(async () => {
```
binodpant, line 280: yeah I am certain in genearl we don't need act around the event actions, esp if you use user-events library

you want to await act() on whatever needs to be awaited for _after_ you click that always works

src/components/settings/SettingsLMSTab/tests/CanvasConfig.test.jsx
```diff
@@ -3,19 +3,45 @@ import {
   render, fireEvent, screen,
 } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
+import { act } from 'react-dom/test-utils';
```
binodpant, line 6: import from testing-library/react instead (even though it's jus a re-export)

src/components/settings/SettingsLMSTab/tests/CanvasConfig.test.jsx
alex-sheehan-edx: because of the hooks we actually need the acts

alex-sheehan-edx: to be more descriptive, I was getting act errors when not wrapping these, I believe because we have hooks running in the background

binodpant: hmm ok we can leave these be for now then, maybe I missed something. Generally the event triggering functions if you use the user-events library are already wrapped in act. I am guessing you need to do this because you are using the built in fireEvent instead of user-events?

src/components/settings/SettingsLMSTab/tests/CanvasConfig.test.jsx
```diff
@@ -148,24 +181,115 @@ describe('<CanvasConfig />', () => {
...
+    act(() => {
+      fireEvent.change(screen.getByLabelText('Display Name'), {
+        target: { value: 'displayName' },
+      });
+      fireEvent.change(screen.getByLabelText('Canvas Base URL'), {
+        target: { value: 'https://www.test.com' },
+      });
+    });
+    expect(screen.getByText('Authorize')).not.toBeDisabled();
+    await act(async () => { userEvent.click(screen.getByText('Authorize')); });
```
binodpant, line 273: For act warning fixes refer to this PR https://github.com/openedx/frontend-app-enterprise-public-catalog/pull/199/files#diff-fd2c79f0d05a4d6d1caeb7bbe4af38a8f9f39c2885d6e212c499f453cd0c2292R318

generally, the click from userEvent already includes act() so we should not need to wrap those in act. Is my understanding. Usually you want to act() on things you await after clicks have been called, for example

I would not stress much over it if this gets tricky but maybe try it one test and see if that idea works!

binodpant (approved):
made a suggestion around act wranings in tests, but otherwise the changes lgtm!

PR #740, opened 2022-03-25
ENT-5573: Config bug bash fixes

kiram15:
# For all changes

- SAP config’s Company ID field needs to not validate numbers only and instead should allow strings
- All string fields need to have a char limit validation similar to display name’s with a 255 limit
- Moodle’s password field needs to be obscured like other password fields
- Degreed2 config forms need to include an optional token fetch base url with a tool tip stating that if one isn’t supplied then we will use the degreed base url provided
- Degreed2 backend should not require the config to have the optional token fetch base url 
- Toggle and Delete error message should read configuration instead of config, and remove “later”
- “Try Again” button should also be removed from error modals

### DIFF
--- a/src/components/settings/SettingsLMSTab/ExistingLMSCardDeck.jsx
+++ b/src/components/settings/SettingsLMSTab/ExistingLMSCardDeck.jsx
@@ -12,10 +12,8 @@ import { handleErrors } from './utils';
 import { TOGGLE_SUCCESS_LABEL, DELETE_SUCCESS_LABEL } from '../data/constants';
 import ConfigError from './ConfigError';
 
-const errorToggleModalText = 'We were unable to toggle your config. Please try submitting again later or contact support for help.';
-const errorDeleteModalText = 'We were unable to delete your config. Please try removing again later or contact support for help.';
-const TOGGLE_ACTION = 'toggle';
-const DELETE_ACTION = 'delete';
+const errorToggleModalText = 'We were unable to toggle your configuration. Please try submitting again or contact support for help.';
+const errorDeleteModalText = 'We were unable to delete your configuration. Please try removing again or contact support for help.';
 const INCOMPLETE = 'incomplete';
 const ACTIVE = 'active';
 const INACTIVE = 'inactive';
@@ -27,8 +25,6 @@ const ExistingLMSCardDeck = ({
   onClick,
 }) => {
   const [errorIsOpen, openError, closeError] = useToggle(false);
-  const [erroredConfig, setErroredConfig] = useState();
-  const [errorCardActionType, setErrorCardActionType] = useState();
   const [errorModalText, setErrorModalText] = useState();
 
   const toggleConfig = async (id, channelType, toggle) => {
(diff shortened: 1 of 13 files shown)


alex-sheehan-edx:
Ok for now because I worked to triage these bugs but screenshots/jira ticket link would be super helpful 😇

src/components/settings/SettingsLMSTab/ExistingLMSCardDeck.jsx
```diff
@@ -45,8 +41,6 @@ const ExistingLMSCardDeck = ({
     if (err) {
       setErrorModalText(errorToggleModalText);
       openError();
-      setErroredConfig({ id, channelType, toggle });
```
alex-sheehan-edx, line 48: actually much cleaner 🥳

src/components/settings/SettingsLMSTab/LMSConfigs/SAPConfig.jsx
```diff
@@ -138,7 +137,8 @@ const SAPConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => {
         </Form.Group>
         <Form.Group className="my-4">
           <Form.Control
-            type="number"
+            type="text"
```
alex-sheehan-edx, line 140: 👍

src/components/settings/SettingsLMSTab/utils.js
```diff
@@ -2,8 +2,9 @@ import { logError } from '@edx/frontend-platform/logging';
 
 export function buttonBool(config) {
   let returnVal = true;
-  Object.values(config).forEach(value => {
-    if (!value) {
+  Object.entries(config).forEach(entry => {
+    const [key, value] = entry;
+    if ((key !== 'displayName' && key !== 'degreedFetchUrl') && !value) {
```
alex-sheehan-edx, line 7: since this is a generic utils file, do we want a degreed specific handling? I would maybe add a comment as to why that's included or move/make a function specific for the degreed config component

src/components/settings/SettingsLMSTab/LMSConfigs/Degreed2Config.jsx
```diff
@@ -150,6 +164,27 @@ const Degreed2Config = ({ enterpriseCustomerUuid, onClick, existingData }) => {
...
+            isInvalid={!fetchUrlValid}
+            onChange={(e) => {
+              setEdited(true);
+              validateField('Degreed Token Fetch Base Url', e.target.value);
+            }}
+            floatingLabel="Degreed Token Fetch Base Url"
+            defaultValue={existingData.degreedFetchUrl}
+          />
+          <Form.Text>
+            Optional: If provided, will be used as base url instead of Degreed Base URL to fetch tokens.
```
alex-sheehan-edx, line 180: Did this text come from UX? Idk why but reading "will be used as base url..." made me think this field would replace the other base url field. It could say "be used as the url to fetch API tokens..." or something but I could also very well be crazy. What do you think?

kiram15: No it didn't come from UX, just from the model description. I'll change it to what you suggested!

PR #745, opened 2022-03-28
fix: bug fixes

kiram15:
# For all changes

- Error modal’s contact support buttons (for both modals related to existing config kebab menu usage and config form usage) should link to   and open in a new tab
- The url used for deleting Canvas configs needs to be fixed /Canvas/ => /canvas/
- For all channels, config forms should only set the config to inactive when creating a new config    
- currently `active=false` is being appended to all payloads 

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
@@ -160,8 +160,6 @@ const BlackboardConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => 
     event.preventDefault();
     // format config data for the backend
     const transformedConfig = snakeCaseDict(config);
-    // this will need to change based on save draft/submit
-    transformedConfig.active = false;
     transformedConfig.enterprise_customer = enterpriseCustomerUuid;
     let err;
     // If we have a config that already exists, or a config that was created when authorized, post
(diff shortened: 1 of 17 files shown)


src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
```diff
@@ -176,6 +174,7 @@ const BlackboardConfig = ({ enterpriseCustomerUuid, onClick, existingData }) =>
     // Otherwise post a new config
     } else {
       try {
+        transformedConfig.active = false;
```
binodpant, line 177: not a major thing but: do we have to do tis for each LMS implementation, or can it be done in a common place such as whatever creates the `config` before snakeCaseDict is called?  that may help centralize this state value and this code would just simply post the config ... just an idea

binodpant: one reason would be if we add a new config, implementation, we would have to remember to do this there as well, vs having in a central place, is what I thought

kiram15: It doesn't have to be done for each implementation, because the error that we're correcting is that it should only be set to false for new ones. If we attempted to centralize it, it would get reset to false even when an update was being posted.

binodpant: ah I see ok thanks for clarifying

PR #756, opened 2022-04-01
fix: pass enterprise customer uuid to backend when we create a new report configuration

muhammad-ammar:
# For all changes

### DIFF
--- a/src/components/ReportingConfig/ReportingConfigForm.jsx
+++ b/src/components/ReportingConfig/ReportingConfigForm.jsx
@@ -122,6 +122,7 @@ class ReportingConfigForm extends React.Component {
       }
     } else {
       // ...or create a new configuration
+      formData.append('enterprise_customer_id', this.props.enterpriseCustomerUuid);
       const err = await this.props.createConfig(formData);
       if (err) {
         this.setState({ submitState: SUBMIT_STATES.ERROR });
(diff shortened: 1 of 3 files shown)


src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -122,6 +122,7 @@ class ReportingConfigForm extends React.Component {
       }
     } else {
       // ...or create a new configuration
+      formData.append('enterprise_customer_id', this.props.enterpriseId);
```
saleem-latif, line 125: Can/Should we use `config.enterpriseCustomer.uuid` here like on https://github.com/openedx/frontend-app-admin-portal/pull/756/files#diff-d6c597698145cb29bdb34a6ead6524e9808691c30432db9f451bc697e99b9bd4R117

src/components/ReportingConfig/ReportingConfigForm.jsx
```diff
@@ -381,6 +382,7 @@ ReportingConfigForm.defaultProps = {
 };
 
 ReportingConfigForm.propTypes = {
+  enterpriseId: PropTypes.string.isRequired,
```
saleem-latif, line 385: nit: I think we should rename the prop to `enterpriseCustomerId` to make it consistent with other places?

saleem-latif (approved):
Just a couple of suggestions, other than that looks good  👍

src/components/ReportingConfig/ReportingConfigForm.jsx
muhammad-ammar: `config` object is available only when we are updating an existing report config

PR #761, opened 2022-04-12
chore: validating existing LMS configs on load in the settings page

alex-sheehan-edx:
# Description
- LMS config settings page now validates existing configs when editing on load

# For all changes

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/SAPConfig.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/SAPConfig.jsx
@@ -82,16 +82,23 @@ const SAPConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => {
     switch (field) {
       case 'SAP Base URL':
         setSapsfBaseUrl(input);
-        setUrlValid(urlValidation(input) || input.length === 0);
+        setUrlValid(urlValidation(input) || input?.length === 0);
         break;
       case 'Display Name':
         setDisplayName(input);
-        setNameValid(input.length <= 20);
+        setNameValid(input?.length <= 20);
         break;
       default:
         break;
     }
   };
+
+  useEffect(() => {
+    if (!isEmpty(existingData)) {
+      validateField('SAP Base URL', existingData.sapsfBaseUrl);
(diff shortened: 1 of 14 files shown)


src/components/settings/SettingsLMSTab/tests/BlackboardConfig.test.jsx
```diff
@@ -247,4 +252,26 @@ describe('<BlackboardConfig />', () => {
     expect(window.open).toHaveBeenCalled();
     expect(mockFetchSingleConfig).toHaveBeenCalledWith(1);
   });
+  test('validates poorly formatted existing data on load', async () => {
```
binodpant, line 255: do we need async here? I did not see any awaits

kiram15 (approved):
LGTM! :^)

src/components/settings/SettingsLMSTab/LMSConfigs/SAPConfig.jsx
```diff
@@ -82,17 +82,24 @@ const SAPConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => {
...
         setDisplayName(input);
-        setNameValid(input.length <= 20);
+        setNameValid(input?.length <= 20);
         break;
       default:
         break;
     }
   };
 
+  useEffect(() => {
```
binodpant, line 96: Due to the multiple impl we have for all the LMS configs, I see a fewopportunities for DRY; stuff like this could be a common hook for example, all the implementations use it as a validation hook of sorts (can pass in custom validation functions for example), but I don't think it's needed right now, just mentioning as a future idea if the codebase becomes larger

this could e.g., become a nice validation hook all components can use or be required to us-e by default 

the hook would take some inputs like field names, and take care of the actual validating via other shared `validateField` functions, but it would require refactors :)

src/components/settings/SettingsLMSTab/tests/BlackboardConfig.test.jsx
alex-sheehan-edx: I thiiink I just accidentally left the async in haha

PR #768, opened 2022-04-20
fix: Stop returning undefined query parameter values.

macdiesel:
# For all changes


georgebabey (approved):
I can't think of a use case where we would want to pass along undefined values so this makes sense to me.

PR #774, opened 2022-04-23
fix: ENT-4182 use enterprise UUID for verifying enterprise admin role

irfanuddinahmad:
# For all changes


zamanafzal (approved):

saleem-latif (approved):

src/components/AdminRegisterPage/index.jsx
```diff
@@ -19,18 +23,33 @@ const AdminRegisterPage = ({ match }) => {
...
+      } else {
+        // user is authenticated but doesn't have the `enterprise_admin` JWT role; force a log out so their
+        // JWT roles gets refreshed. on their next login, the JWT roles will be updated.
+        const logoutRedirectUrl = getLogoutRedirectUrl(getProxyLoginUrl(enterpriseSlug));
+        global.location.href = logoutRedirectUrl;
+      }
+      return <EnterpriseAppSkeleton />;
+    };
+
+    LmsApiService.fetchEnterpriseBySlug(enterpriseSlug)
```
zamanafzal, line 42: We didn't find any way to get the enterpriseUUID from context or any other place?

zamanafzal (approved):

src/components/AdminRegisterPage/index.jsx
```diff
@@ -19,18 +23,33 @@ const AdminRegisterPage = ({ match }) => {
...
+        // user is authenticated and has the ``enterprise_admin`` JWT role, so redirect user to
+        // account activation page to ensure they verify their email address.
+        history.push(`/${enterpriseSlug}/admin/register/activate`);
+      } else {
+        // user is authenticated but doesn't have the `enterprise_admin` JWT role; force a log out so their
+        // JWT roles gets refreshed. on their next login, the JWT roles will be updated.
+        const logoutRedirectUrl = getLogoutRedirectUrl(getProxyLoginUrl(enterpriseSlug));
+        global.location.href = logoutRedirectUrl;
+      }
+      return <EnterpriseAppSkeleton />;
```
muneebGH, line 39: is it necessary to return JSX element here? i think this should work without this return statement, is it?

sameenfatima78 (approved):

muhammad-ammar (approved):
LGTM

src/components/AdminRegisterPage/index.jsx
muneebGH, line 42: just thinking, how about creating a custom hook and encapsulate this api call and stuff in that hook? 

something like `const [enterpriseUUID, fetchError] = useEnterpriseUUID(enterpriseSlug)`

[a example of custom hook](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/60694bad245bd54de8d227b135eb136d53395c4c/src/components/program-progress/data/hooks.jsx#L28)

and the logic of page can be cleaner

```
const [enterpriseUUID, fetchError] = useEnterpriseUUID(enterpriseSlug)

if (!enterpriseUUID) {
return Skeleton
} else {

// dump the logic of processEnterpriseAdmin
}

return Skeleton
```

I feel like it would look cleaner plus hook will be reusable.

irfanuddinahmad: No

src/components/AdminRegisterPage/index.jsx
irfanuddinahmad: if we donot explicitly return, undefined is auto returned

PR #784, opened 2022-05-09
fix: use config uuid for canvas oauth state

johnnagro:
## Description

- we no longer using enterprise uuid so we can disambiguate which config is used in the case the customer has more than 1 configured

## Description

- [ENT-5610](https://2u-internal.atlassian.net/browse/ENT-5610)


src/components/settings/SettingsLMSTab/LMSConfigs/CanvasConfig.jsx
```diff
@@ -135,7 +135,7 @@ const CanvasConfig = ({ enterpriseCustomerUuid, onClick, existingData }) => {
       setOauthPollingTimeout(LMS_CONFIG_OAUTH_POLLING_TIMEOUT);
 
       const oauthUrl = `${canvasBaseUrl}/login/oauth2/auth?client_id=${clientId}&`
-      + `state=${enterpriseCustomerUuid}&response_type=code&`
```
alex-sheehan-edx, line 138: 🥳  nice

alex-sheehan-edx (approved):
👍

alex-sheehan-edx:
what's the order of op here? Is the backend side already updated?

johnnagro:
> what's the order of op here? Is the backend side already updated?

order should not matter ✨

PR #798, opened 2022-06-09
Removed Table and StatusAlert  deprecation of Paragon

abdullahwaheed:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102)

### What has changed
Removed deprecated `Table` component of paragon and updated it with `DataTable`
Removed deprecated `StatusAlert` component of paragon and updated it with `Alert`

### References
[Paragon Table](https://paragon-openedx.netlify.app/components/table/)
[Paragon DataTable](https://paragon-openedx.netlify.app/components/datatable/)

[Paragon Alert](https://paragon-openedx.netlify.app/components/alert/)
[Paragon StatusAlert](https://paragon-openedx.netlify.app/components/statusalert/)


davidjoy (approved):
This seems like a fine change - the owning team should also review so they're aware of the change, it should be tested by Fed-BOM and we can get it in.  I did a cursory review.

src/components/TableComponent/index.jsx
```diff
@@ -86,14 +86,15 @@ class TableComponent extends React.Component {
           <div className="col">
             {loading && <TableLoadingOverlay />}
             <div className="table-responsive">
-              <Table
+              <DataTable
```
adamstankiewicz, line 89: Similar to the suggestion in `StatusAlert`, what would it take to no longer have a custom `TableComponent/index.jsx` file/component, and instead only rely on `DataTable` directly where `TableComponent/index.jsx` was previously used?

This would be more tricky to do for tables than the alerts throughout the application, though, so I'd love to understand the level of effort involved and/or what functionality is in this `TableComponent/index.jsx` that can't be replicated with `DataTable` alone.

adamstankiewicz (changes requested):
Left some initial comments/suggestions. May you also provide screenshots of the proposed changes?

src/components/StatusAlert/index.jsx
```diff
@@ -14,34 +13,30 @@ const StatusAlert = (props) => {
     onClose,
   } = props;
 
+  const showIcon = () => (
+    <div className="icon">
+      <Icon className={iconClassName} />
+    </div>
+  );
```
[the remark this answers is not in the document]
abdullahwaheed: i have added this icon to make it consistent with the legacy component, now as we are planning to remove this component, it will be outdated

src/components/TableComponent/index.jsx
abdullahwaheed: yes it can be changed to DataTable but require much file changes, i would suggest we should do it in a separate PR to keep the scope of this PR intact

abdullahwaheed:
@adamstankiewicz any updates on this?

abdullahwaheed:
this PR is much outdated and there are some critical conflicts, closing this for now and will create a new one

PR #804, opened 2022-06-15
feat: hide learner credit management automatically, refactor sidebar

long74100:
# For all changes


src/components/EnterpriseApp/EnterpriseAppRoutes.jsx
```diff
@@ -0,0 +1,166 @@
...
+          component={SettingsPage}
+        />
+      )}
+
+      {canManageLearnerCredit && (
+      <Route
+        exact
+        path={`${baseUrl}/admin/learner-credit`}
+        component={LearnerCreditManagement}
+      />
```
adamstankiewicz, line 144: nit: could use extra indents

src/components/subsidy-requests/SubsidyRequestsContext.jsx
```diff
@@ -0,0 +1,43 @@
+import { createContext, useMemo } from 'react';
+import {
+  useSubsidyRequestConfiguration,
+  useSubsidyRequestsOverview,
+} from './data/hooks';
+
+export const SubsidyRequestsContext = createContext();
+
+export const useSubsidyRequestsContext = (enterpriseId) => {
```
adamstankiewicz, line 9: If this just a hook now versus a component like before, should it be in `./data/hooks` versus a component file?

src/data/services/EcommerceApiService.js
```diff
@@ -86,6 +86,18 @@ class EcommerceApiService {
     }
     return EcommerceApiService.apiClient().post(url, options);
   }
+
+  static fetchEnterpriseOffers(options = {
+    status: 'Open',
+  }) {
+    const { enterpriseId } = store.getState().portalConfiguration;
+    let url = `${EcommerceApiService.ecommerceBaseUrl}/api/v2/enterprise/${enterpriseId}/enterprise-admin-offers/`;
+    if (options) {
```
adamstankiewicz, line 95: [curious] If `options` has a default function argument to fallback on, will `options` ever be falsy here?

src/hooks/useOnMount.jsx
```diff
@@ -0,0 +1,14 @@
+import { useEffect, useRef } from 'react';
+
+const useOnMount = (callback) => {
```
adamstankiewicz, line 3: nit: maybe add a JSDoc describing how this is intended to be used for others / future us (e.g., that the callback function passed in is what's called on mount, etc.).

src/data/services/EcommerceApiService.js
adamstankiewicz: Agreed, sounds good :)

src/components/subsidy-requests/SubsidyRequestsContext.jsx
long74100: will leave as is for now

src/components/EnterpriseSubsidiesContext/index.jsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useMemo } from 'react';
+import { useEnterpriseOffers } from '../enterprise-offers/data/hooks';
```
adamstankiewicz, line 2: nit/suggestion: not blocking, but curious about having this hook live in `components/enterprise-offers`, if `enterprise-offers` folder doesn't actually contain any components? Would it make sense to co-locate this hook where it's called in `components/EnterpriseSubsidiesContext`?

adamstankiewicz (approved):
Left 1 non-blocking nit/suggestion around where `useEnterpriseOffers` might go, but LGTM otherwise.

PR #817, opened 2022-06-29
fix: no longer multiply `percentUtilized` by 100 since API already returns this value.

adamstankiewicz:
# For all changes


adamstankiewicz:
Closing in favor of changing the API data to return the percentage as 0-1.

PR #822, opened 2022-07-05
fix: guard against bad offer summary data

long74100:
# For all changes

### DIFF
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -12,10 +12,11 @@ import {
 export const transformOfferSummary = (offerSummary) => {
   if (!offerSummary) { return null; }
 
+  // Guard against potential bad data
   const totalFunds = parseFloat(offerSummary.maxDiscount);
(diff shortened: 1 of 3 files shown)


src/components/learner-credit-management/data/utils.js
```diff
@@ -12,10 +12,11 @@ import {
 export const transformOfferSummary = (offerSummary) => {
   if (!offerSummary) { return null; }
 
-  const totalFunds = parseFloat(offerSummary.maxDiscount);
-  const redeemedFunds = parseFloat(offerSummary.amountOfOfferSpent);
-  const remainingFunds = parseFloat(offerSummary.remainingBalance);
-  const percentUtilized = parseFloat(offerSummary.percentOfOfferSpent);
+  // Guard against potential bad data
```
adamstankiewicz, line 15: nit/curious: we could keep `parseFloat` juuuust in case the API decides to change one day to strings, it won't break the app.

adamstankiewicz (approved):

PR #825, opened 2022-07-06
fix: handle when max discount is not available in offer summary metrics

adamstankiewicz:
# For all changes


long74100 (approved):
LLTM

PR #826, opened 2022-07-06
chore: current config status during auth

alex-sheehan-edx:
Bug fix on current config data after a config is created during the authorization step. Clicking the authorization button triggers a creation of an LMS integration config on the backend. Since the page doesn't trigger a success signal until confirming the config has a refresh token, and the process of retrieving the refresh token can potentially fail, the current config data needs to be set between the time the new config is created and the page has confirmed the refresh token exists and triggers the success signal.

# For all changes

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
@@ -20,7 +20,7 @@ import {
 } from '../../data/constants';
 
 const BlackboardConfig = ({
-  enterpriseCustomerUuid, onClick, existingData, existingConfigs,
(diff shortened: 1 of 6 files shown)


src/components/settings/SettingsLMSTab/LMSConfigs/BlackboardConfig.jsx
```diff
@@ -134,7 +146,7 @@ const BlackboardConfig = ({
       if (!appKey) {
         try {
           const response = await LmsApiService.fetchBlackboardGlobalConfig();
-          appKey = response.data.results[0].app_key;
+          appKey = response.data.results[response.data.results.length - 1].app_key;
```
alex-sheehan-edx, line 149: youch, caught this when testing. The backend is returning the list from oldest -> newest, so we need to fetch the last in the list not the first.

kiram15 (approved):
LGTM!

PR #832, opened 2022-07-12
feat: pull subsidies info into subsidies context, refactors, auto upd…

long74100:
…ate bnr config

- Pull enterprise subsidy information out of settings context and into the enterprise subsidies context, this will let us use subsidy information to conditionally render product tours, pages, etc.
- Hide BNR product tour if the enterprise has no subsidies (still need to hide the settings page but will defer since pr is getting too big)
- Auto-update subsidy request configuration if an enterprise adds a new subsidy 
- Disallow universal link generation if subsidies have all expired or if there are none

# For all changes


src/components/ProductTours/ProductTours.jsx
```diff
@@ -42,6 +47,7 @@ const ProductTours = ({ enterpriseSlug }) => {
 
 ProductTours.propTypes = {
   enterpriseSlug: PropTypes.string.isRequired,
+  enableLearnerPortal: PropTypes.bool.isRequired,
```
adamstankiewicz, line 50: If this component is already connected to the Redux store to get `enterpriseSlug`, should we grab `enableLearnerPortal` from the store versus using prop drilling the value through the parent component?

src/components/settings/SettingsAccessTab/SettingsAccessGenerateLinkButton.jsx
```diff
@@ -57,10 +57,11 @@ const SettingsAccessGenerateLinkButton = ({
 
 SettingsAccessGenerateLinkButton.defaultProps = {
   disabled: false,
+  formattedLinkExpirationDate: null,
 };
 SettingsAccessGenerateLinkButton.propTypes = {
   enterpriseUUID: PropTypes.string.isRequired,
-  formattedLinkExpirationDate: PropTypes.string.isRequired,
+  formattedLinkExpirationDate: PropTypes.string,
```
adamstankiewicz, line 64: [curious] Should this be allowed to be `null` if `expiration_date` is a required field on the `EnterpriseCustomerInviteKey` model ([source](https://github.com/openedx/edx-enterprise/blob/master/enterprise/models.py#L3323-L3329))?

adamstankiewicz (commented):
Nice clean up here! Just a couple questions.

src/components/ProductTours/ProductTours.jsx
long74100: yeah we should do that.

src/components/settings/SettingsAccessTab/SettingsAccessGenerateLinkButton.jsx
long74100: My original thought was the disabled should always = true if formattedLinkExpirationDate is null. But I think we can enforce it in this component so that the button is disabled if formattedLinkExpirationDate is null.

adamstankiewicz: Got it, thanks for the clarification. Makes sense that we'll disable the button when formattedLinkExpirationDate is null. 

Currently, though, this component is largely controlled in that it's solely up to its parent to control its disabled behavior and would probably prefer not to mix controlled with uncontrolled component usage here. That is, still rely solely on the parent to control whether this button is disabled or not (based on a null `formattedLinkExpirationDate`). Just worried that we're not mixing the `disabled` behavior to not always be based on what's passed into the `disabled` prop itself from the parent component.

Could we instead make the `formattedLinkExpirationDate` prop conditionally required if `disabled` is false?

See [this example](https://github.com/openedx/paragon/blob/a95624d0721dcfddb7a201543fb23cf44f07c931/src/utils/propTypes/utils.js#L51) of a custom prop type util function to do this type of conditional prop checking from Paragon and [these docs](https://reactjs.org/docs/typechecking-with-proptypes.html) for more details on creating custom validators.

long74100: That is a good point, let me look into this.

adamstankiewicz (approved):

PR #840, opened 2022-07-21
docs: add ADRs about support for Enterprise Offer subsidies.

iloveagent57:
Adds ADRs about Enterprise Offer support and where data comes from.

# For all changes


docs/decisions/0003-learner-credit-management.rst
```diff
@@ -0,0 +1,37 @@
...
+Status
+******
+
+Accepted - July 2022
+
+Context
+*******
+
+* We decided to enable the consumption of Enterprise Offers in the Enterprise Learner Portal (TODO: link to LP ADR)
+* Enterprise admins need the ability to check the cash remaining on their Enterprise offer,
```
adamstankiewicz, line 13: The Learner Credit Management screen also currently supported offers with no `max_discount` set, in which case there is no notion of "cash remaining" or upper limit. Instead a single card with "Total Redeemed" is shown to show their current spend.

Might be worth expanding the context to include this bit.

docs/decisions/0004-learner-credit-data-from-analytics-service.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Consequences
+************
+
+* We're now bound to the data pipeline for this feature.
+* Fresh data only visible to an end-user after a prefect job transfers data from our data warehouse into
+  the Analytics API.  This was once/day at time of initial implementation, but has been increased to TODO times/day.
+* Pipeline failures mean this data will be stale until failure is resolved.
+* This puts us in a weird position where ecommerce is the source of truth about Enterprise Offer state
+  *except* for the remaining balance for a given Offer.
```
adamstankiewicz, line 31: Similarly, it's not just about remaining balance, but also the individual offer redemptions (in the table) and the total spend, etc.

Would it be worth saying the inverse here, as well, where the ecommerce data is only currently used for the offer display name and start/end dates?

docs/decisions/0003-learner-credit-management.rst
```diff
@@ -0,0 +1,41 @@
...
+*******
+
+* We decided to enable the consumption of Enterprise Offers in the Enterprise Learner Portal (TODO: link to LP ADR)
+* Enterprise admins need the ability to check the remaining balance and/or total balance redeemed
+  on their Enterprise offer, so they can plan their learning program accordingly.
+
+Decisions
+*********
+
+* Initially will only support "first-come, first-serve, org-wide budget" flavor of Offers.
```
long74100, line 19: Might be worth calling out that the page is also available for enterprises with 1 offer.

long74100 (approved):

PR #847, opened 2022-08-25
chore: adding default provider config values to self service

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-6207

adding default values
```* Debug Mode
* Skip hinted dialog
* Skip registration form
* Skip email verfication
* Send to registration first
* Automatic metadata refresh with metadata URL specified
```

to form data when creating/updating provider config forms in the self service settings page

Tested locally that both old and new forms update with the new default values


kiram15 (approved):
LGTM!

PR #848, opened 2022-08-29
chore: fix tests using waitFor without await

long74100:
Fixes test where we were not using await with waitFor.

### DIFF
--- a/src/components/settings/SettingsAccessTab/tests/ActionsTableCell.test.jsx
+++ b/src/components/settings/SettingsAccessTab/tests/ActionsTableCell.test.jsx
@@ -25,7 +25,7 @@ const TEST_INVITE_KEY_UUID = 'test-uuid-1';
 
 const ActionsTableCellWrapper = (props) => (
   <IntlProvider locale="en">
-    <ActionsTableCell {...props} />
+    <ActionsTableCell enterpriseUUID="test-enterprise-id" {...props} />
   </IntlProvider>
 );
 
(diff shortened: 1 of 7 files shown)


src/components/subscriptions/SubscriptionPlanRoutes.jsx
```diff
@@ -5,7 +5,7 @@ import { Route } from 'react-router-dom';
 import { connect } from 'react-redux';
 
 import MultipleSubscriptionsPage from './MultipleSubscriptionsPage';
-import SubscriptionDetailPage from './SubscriptionDetailPage';
+import SubscriptionDetailPageContainer from './SubscriptionDetailPage';
```
christopappas, line 8: does `Container` get appended to the component name when it is exported?

long74100: This is the default export so you can name it anything. Appending container here since it's using the component connected to redux.

christopappas (approved):

src/components/settings/SettingsAccessTab/tests/ActionsTableCell.test.jsx
```diff
@@ -64,10 +65,10 @@ describe('ActionsTableCell', () => {
       />,
     );
     const copyBtn = screen.getByText('Copy');
-    await act(async () => { userEvent.click(copyBtn); });
+    fireEvent.click(copyBtn);
```
adamstankiewicz, line 68: nit: was `fireEvent` necessary or could we have used `userEvent` here still?

src/components/subscriptions/SubscriptionPlanRoutes.jsx
adamstankiewicz: An alternative name might be `ConnectedSubscriptionDetailPage`. OK with either, though.

adamstankiewicz (approved):

PR #849, opened 2022-08-30
feat: Adding branding custom theme cards 

kiram15:
Test by passing feature flag in url http://localhost:1991/test-enterprise/admin/settings/appearance/?feature=SETTINGS_PAGE_APPEARANCE_TAB

Screen Shot 2022-08-30 at 4 55 01 PM

https://2u-internal.atlassian.net/browse/ENT-6085

# For all changes


src/components/settings/SettingsAppearanceTab/ThemeSvg.jsx
```diff
@@ -0,0 +1,29 @@
+import PropTypes from 'prop-types';
+import Color from 'color';
+import { DARK_COLOR, WHITE_COLOR } from '../data/constants';
+
+export default function ThemeSvg({ themeVars }) {
+  const bannerColor = (Color(themeVars.banner).isDark() ? Color(WHITE_COLOR) : Color(DARK_COLOR));
+  return (
+    <svg viewBox="87.68 236.596 356.789 153.834" xmlns="http://www.w3.org/2000/svg">
+      <g>
+        <rect id="background" x="87.719" y="236.819" width="356.605" height="153.611" style={{ fill: 'rgb(255, 255, 255)' }} />
```
alex-sheehan-edx, line 10: are we not worrying about mobile? Not a huge issue, I assume yes but just curious how the set width height will affect things. And just small follow up- are we sure we don't want these to grow with screen size?

alex-sheehan-edx (approved):
one context question but otherwise LGTM!

src/components/settings/SettingsAppearanceTab/ThemeSvg.jsx
kiram15: They do resize! You can view the cards in a really small window and see the effect.

PR #850, opened 2022-09-06
feat: Render 404 page on non-existent enterprise-slug

brobro10000:
https://2u-internal.atlassian.net/browse/ENT-3600

- Add unified logic to redirect to 404 page on bad enterprise slug
- Use pre-existing component to render 404 page
- Checks enterprise existence before provider renders

Screen Shot 2022-09-06 at 16 06 54 PM

# For all changes

- [X] Ensure adequate tests are in place (or reviewed existing tests cover changes)

- [X] Ensure to attach screenshots
- [X] Ensure to have UX team confirm screenshots

### DIFF
--- a/src/components/EnterpriseApp/EnterpriseApp.test.jsx
+++ b/src/components/EnterpriseApp/EnterpriseApp.test.jsx
@@ -36,7 +36,7 @@ jest.mock('../ProductTours/ProductTours', () => ({
 
 jest.mock('./EnterpriseAppContextProvider', () => ({
   __esModule: true,
-  default: ({ children }) => <EnterpriseSubsidiesContextProvider>{ children }</EnterpriseSubsidiesContextProvider>,
+  default: ({ children }) => <EnterpriseSubsidiesContextProvider>{children}</EnterpriseSubsidiesContextProvider>,
 }));
 
 jest.mock('../../containers/Sidebar', () => ({
@@ -50,7 +50,6 @@ describe('<EnterpriseApp />', () => {
       url: '',
       params: {
         enterpriseSlug: 'foo',
-        page: 'settings',
       },
     },
     location: {
@@ -67,6 +66,11 @@ describe('<EnterpriseApp />', () => {
     enterpriseName: 'test-enterprise',
   };
 
+  const invalidEnterpriseId = {
+    ...basicProps,
+    enterpriseId: null,
+    enterpriseName: null,
+  };
   beforeEach(() => {
     getAuthenticatedUser.mockReturnValue({
       username: 'edx',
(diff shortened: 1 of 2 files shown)


christopappas:
Looking good to me. Can we update the test file to perhaps cover the case where we return the <PageNotFound> component? Also, probably makes sense to remove `page` from the testprops if it's no longer required or used by this component

src/components/EnterpriseApp/index.jsx
```diff
@@ -125,6 +126,10 @@ class EnterpriseApp extends React.Component {
       return <EnterpriseAppSkeleton />;
     }
 
+    if (!loading && enterpriseId === null) {
```
adamstankiewicz, line 129: nit: we can assume `loading` is false if we're getting to this conditional due to the conditional on line 125, so its unnecessary to check on this line :)

curious: it may also make sense to use `if (!enterpriseId) { ... }` rather than specifically checking for `null` to catch cases where it may be `undefined` or `null` (in both situations, we'd want to display the `NotFoundPage`.

src/components/EnterpriseApp/index.jsx
```diff
@@ -125,6 +126,10 @@ class EnterpriseApp extends React.Component {
...
+      return <NotFoundPage />;
```
adamstankiewicz, line 130: It's probably worth adding a test case in `EnterpriseApp.test.jsx` for this scenario of rendering the `NotFoundPage`.

adamstankiewicz (commented):
This looks great! Just a small nit/question and an ask to add a test :)

src/components/EnterpriseApp/EnterpriseApp.test.jsx
```diff
@@ -63,10 +62,16 @@ describe('<EnterpriseApp />', () => {
...
-    enterpriseId: 'uuid',
-    enterpriseName: 'test-enterprise',
   };
 
+  basicProps.enterpriseId = 'uuid';
+  basicProps.enterpriseName = 'test-enterprise';
+
+  const invalidEnterpriseNameProps = {
+    ...basicProps,
+    enterpriseId: null,
```
christopappas, line 72: nit/question: if the creation of invalidEnterpriseNameProps overwrites the values of `enterpriseId` and `enterpriseName`, is it necessary to pull out and re-add those 2 values out of basicProps?

christopappas (approved):
LGTM, just 1 quick question

src/components/EnterpriseApp/EnterpriseApp.test.jsx
adamstankiewicz: I second Chris' nit/question and raise a third nit 😉 If we don't have the enterprise ID, we likely also don't have the enterprise name so it could be confusing to have `null` `enterpriseId` but a valid `enterpriseName`.

adamstankiewicz (approved):
LGTM! :shipit:

PR #852, opened 2022-09-07
feat: adding dynamic functionality

kiram15:
https://user-images.githubusercontent.com/31229189/188977363-d57b1213-140e-48b7-a4d8-930cb06ebee3.mov

Talked to UX, they said that they want all primary and branded buttons to be the color selected with the exception of outline buttons at the moment. As you can see, after the page reloads, the logo and colors apply to the admin portal. Not seen, but this changes the values in Django admin, so the same changes will be happening on learner portal. 

To test this, you need to include with this feature flag http://localhost:1991/test-enterprise/admin/settings/appearance/?feature=SETTINGS_PAGE_APPEARANCE_TAB

# For all changes


src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -9,27 +11,67 @@ import {
...
+      if (response.status === 204) {
+        setConfigChangeSuccess(true);
+      }
+    } catch {
+      setConfigChangeSuccess(false);
+    }
+  }, [enterpriseId, theme]);
+
+  const saveChanges = () => {
+    window.location.reload();
```
kiram15, line 72: Not sure if this is the most graceful way to trigger a page refresh?

PR #854, opened 2022-09-12
feat: add custom theme and dynamic functionality to settings appearance page

kiram15:
https://2u-internal.atlassian.net/browse/ENT-6099

https://user-images.githubusercontent.com/31229189/189687576-5e686eaa-744f-4f28-b395-2372d9e24909.mov

# For all changes


alex-sheehan-edx:
Two quick comments which were discussed offline- I think it would be great if the entire theme card could be clickable rather than just the radio, especially if the card has an onHover behavior. Secondly, I think the confirmation toast is popping up after the theme submission but disappears on the refresh almost immediately. I'm not sure how to remedy this as the page is reloading but as it stands now it's difficult to read the text on the modal. Perhaps a query param or fragment id?

alex-sheehan-edx:
Not a direct comment on your code but seeing the `Portal Appearance` tab next to `Configure Access` and `LMS` really makes me want to change the LMS tab title to something more informative. I think UX has this but just thought I'd also say something  ¯\_(ツ)_/¯

alex-sheehan-edx (commented):
looking pretty good! Had a couple comments and questions

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+    >
+      <ModalDialog.Header>
+        <ModalDialog.Title>
+          Custom Theme
+        </ModalDialog.Title>
+        <p className="mt-3">{headerText}</p>
+      </ModalDialog.Header>
+
+      <ModalDialog.Body>
+        <Form>
```
adamstankiewicz, line 60: nit: btw, you should check out `Formik`. it's used by quite a few MFEs now for implementing forms to avoid having to write all those `useState` hooks yourself. We recently introduced it in frontend-app-learner-portal-enterprise as well in the `UserEnrollmentForm` component :) 10/10 would recommend again 😹 

https://formik.org/docs/overview

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+              defaultValue={customColors?.accent}
+            />
+            {!accentValid && (<Form.Control.Feedback type="invalid">{invalidMessage}</Form.Control.Feedback>)}
+          </Form.Group>
+        </Form>
+        <p className="mt-4 mb-1 d-flex x-small">
+          <Icon src={InfoOutline} className="mr-2" /> {infoText}
+          <Hyperlink
+            destination={HELP_CENTER_BRANDING_LINK}
+            style={{ display: 'contents' }}
```
adamstankiewicz, line 102: nit/curious: why was this style override necessary? just wondering if there's anything needed to be improved on in the `Hyperlink` component in Paragon.

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+            target="_blank"
+          >
+            Enterprise Help Center.
+          </Hyperlink>
+        </p>
+      </ModalDialog.Body>
+
+      <ModalDialog.Footer>
+        <ActionRow>
```
adamstankiewicz, line 111: Nice use of `ActionRow` even though there's only one button in it :)

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+CustomThemeModal.defaultProps = {
+  customColors: null,
+};
+
+CustomThemeModal.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  close: PropTypes.func.isRequired,
+  customColors: PropTypes.arrayOf(
+    PropTypes.strings,
```
adamstankiewicz, line 134: nit: Is `PropTypes.strings` a valid property? I haven't seen it plural before, only `PropTypes.string` 🤔

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
...
+
+  function deleteCustom() {
+    if (selected[0].title === 'Custom Theme') {
+      setTheme([SCHOLAR_THEME, null]);
+    } else {
+      setTheme([selected[0], null]);
+    }
+  }
+
+  // AFTER DELETE MUST CODE
```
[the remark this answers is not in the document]
adamstankiewicz: I love it, though; it's very authoritative haha 😄

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
 import PropTypes from 'prop-types';
-import { Card, Form } from '@edx/paragon';
+import { Button, Card, Form } from '@edx/paragon';
 import ThemeSvg from './ThemeSvg';
+import { SCHOLAR_THEME } from '../data/constants';
+
+export default function ThemeCard({
+  themeVars, selected, setTheme, openCustom,
+}) {
+  const isCustom = themeVars.title === 'Custom Theme';
```
adamstankiewicz, line 9: nit: Is it worth putting this string in a constant `CUSTOM_THEME_LABEL` or something to avoid having to repeat it twice (and in case a test needs it in the future, it could be worth exporting it as a named export as well, i.e. `export const CUSTOM_THEME_LABEL`).

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
...
+          checked={selected[0].title === themeVars.title}
+          onChange={() => selectTheme(themeVars)}
           data-testid={`radio-${themeVars.title}`}
           name="theme"
-          defaultChecked={themeVars.title === 'Scholar (Default)'}
           value={themeVars.title}
         />
       </Card.Footer>
+      {isCustom && (
+      <div className="ml-auto">
```
adamstankiewicz, line 38: Would this be a good place for `ActionRow`?

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
...
+        <Button className="" variant="tertiary" onClick={() => deleteCustom()}>Delete</Button>
+        <Button className="" variant="tertiary" onClick={() => openCustom()}>Edit</Button>
+      </div>
+      )}
+      {!isCustom && (<div className="pt-3" />)}
```
adamstankiewicz, line 43: nit: would it be worth conditionally adding a `pb-3` class to the above element when `isCustom` is false rather than rendering a new `div`? E.g., `className={classNames({ 'pb-3': !isCustom })}`

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
...
-  selected: PropTypes.shape({
-    title: PropTypes.string.isRequired,
     banner: PropTypes.string.isRequired,
     accent: PropTypes.string.isRequired,
-    button: PropTypes.string.isRequired,
   }).isRequired,
+  selected: PropTypes.arrayOf(
+    PropTypes.oneOfType([
+      PropTypes.string.isRequired,
+      PropTypes.arrayOf(PropTypes.string.isRequired),
```
adamstankiewicz, line 58: [curious] Similar to Alex's question, why does the data type of `selected` change? Is there a way to avoid this?

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,35 +1,86 @@
...
+  function getStartingTheme() {
+    for (let i = 0; i < curatedThemes.length; i++) {
+      if (curatedThemes[i].button === enterpriseBranding.primary_color
+        && curatedThemes[i].banner === enterpriseBranding.secondary_color
+        && curatedThemes[i].accent === enterpriseBranding.tertiary_color) {
+        return [curatedThemes[i], null];
+      }
+    }
+    const CUSTOM_THEME = {
+      title: 'Custom Theme',
```
adamstankiewicz, line 35: nit: maybe another place to use that constant?

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,35 +1,86 @@
...
+    return [CUSTOM_THEME, CUSTOM_THEME];
+  }
 
-  async function handleProcessUpload({
+  // this variable is [selected , null] if the user has not created a custom
+  // theme and it is [selectedTheme, [customName, #primary, #secondary, #tertiary]] if they have
+  // (they can't be two variables because of rules regarding setting react hooks)
+  const [theme, setTheme] = useState(getStartingTheme());
+
+  async function handleLogoUpload({
```
adamstankiewicz, line 48: Cool to see `Dropzone` in action for the first time :)

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,35 +1,86 @@
...
       handleError(error);
     }
   }
 
+  const saveChanges = () => {
+    const sendThemeData = async (formData) => {
+      const response = await LmsApiService.updateEnterpriseCustomerBranding(enterpriseId, formData);
+      if (response.status === 204) {
+        setConfigChangeSuccess(true);
+        window.location.reload();
```
adamstankiewicz: Discussed offline already, but can we avoid a full page reload in favor of updating the Redux store?

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,65 @@
...
       <Card.Section className="p-0">
         <ThemeSvg themeVars={themeVars} />
       </Card.Section>
-      <Card.Footer className="justify-content-between pt-2.5">
+      <Card.Footer className="justify-content-between pt-2.5 pb-0">
         <h3>{themeVars.title}</h3>
         <Form.Radio
-          checked={selected.title === themeVars.title}
-          onChange={() => setTheme(themeVars)}
+          checked={selected[0].title === themeVars.title}
```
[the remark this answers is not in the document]
kiram15: Sorry this is referenced in this comment but I can make this more readable!
```
  // this variable is [selected , null] if the user has not created a custom
  // theme and it is [selectedTheme, [customName, #primary, #secondary, #tertiary]] if they have
  // (they can't be two variables because of rules regarding setting react hooks)
```

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,48 +1,121 @@
...
+      banner: enterpriseBranding.secondary_color,
+      accent: enterpriseBranding.tertiary_color,
+    };
+    return [CUSTOM_THEME, CUSTOM_THEME];
+  }
 
-  async function handleProcessUpload({
+  // this variable is [selected , null] if the user has not created a custom
+  // theme and it is [selectedTheme, [customName, #primary, #secondary, #tertiary]] if they have
+  // (they can't be two variables because of rules regarding setting react hooks)
```
adamstankiewicz, line 45: Ah, I see these comments explaining the behavior elsewhere now. Anyway to make the return data types consistent here?

src/components/settings/data/hooks.js
```diff
@@ -56,6 +63,108 @@ export const useLinkManagement = (enterpriseUUID) => {
...
+      .bg-brand-primary {
+        background-color: ${brandColors.primary.regular.hex()} !important;
+      }
+      .border-brand-primary {
+        border-color: ${brandColors.primary.regular.hex()} !important;
+      }
+    `),
+  });
+
+  return styles;
```
adamstankiewicz, line 165: Because this is a re-usable hook, it may make sense to ensure `styles` is memoized (`useMemo`) such that it doesn't cause unnecessary re-renders (i.e., right now, every time the hook runs it will have a new object reference and if it were used in a `useEffect` by a consumer, for example, it could cause that consumer to run the `useEffect` more times than intended.

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,48 +1,121 @@
...
+
   return (
     <>
       <h2 className="py-2">Portal Appearance</h2>
       <p>
         Customize the appearance of your learner and administrator edX experiences with your
         organization’s logo and color themes.
       </p>
+      {configChangeSuccess === false && (
+      <Alert
```
adamstankiewicz, line 99: Let's use the `show` prop to control the `Alert` visibility rather than conditionally rendering the entire `Alert` component. It turns out this approach (versus using `show={true/false}` is that we lose the animation fade in/out.

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,48 +1,121 @@
...
+  };
+
   return (
     <>
       <h2 className="py-2">Portal Appearance</h2>
       <p>
         Customize the appearance of your learner and administrator edX experiences with your
         organization’s logo and color themes.
       </p>
+      {configChangeSuccess === false && (
```
adamstankiewicz, line 98: nit: i got a bit confused by this variable name. to clarify, this only gets set to `false` when the config change errored in some way? Would it be worth having a separate state variable to track the presence of an error rather than relying on a state variable that indicates success? This might make it possible to default the `configChangeSuccess` state variable to `false` instead of a `null` and avoid having the variable switch data types 🤔

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -53,11 +126,20 @@ export default function SettingsAppearanceTab({
           'image/*': ['.png'],
         }}
       />
+      )}
+      {uploadedFile && (
+      <p className="image-preview">
+        <Image
+          src={window.URL.createObjectURL(uploadedFile)}
```
adamstankiewicz, line 133: nit: `global` versus `window` _just in case_ this code ever runs in an environment where `window` doesn't exist.

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -53,11 +126,20 @@ export default function SettingsAppearanceTab({
...
+        />
+      </p>
+      )}
+
       <Toast
-        onClose={() => setConfigChangeSuccess(false)}
-        show={configChangeSuccess}
+        onClose={() => setConfigChangeSuccess(null)}
```
adamstankiewicz, line 139: nit: could this be set back to `false` instead of `null`?

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -77,10 +159,35 @@ export default function SettingsAppearanceTab({
         <ThemeCard themeVars={ACUMEN_THEME} selected={theme} setTheme={setTheme} />
         <ThemeCard themeVars={PIONEER_THEME} selected={theme} setTheme={setTheme} />
       </CardGrid>
+      <h3 className="py-2 pt-5 mb-0">
+        Custom Theme
+      </h3>
+      {theme[1] === null && (
+      <p className="mt-0">
+        Rather use your own colors?
+        <Button variant="link" onClick={openCustom} className="p-0 pl-1" style={{ verticalAlign: 'top' }}>
```
adamstankiewicz, line 168: nit/curious: why was this style override necessary?

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -77,10 +159,35 @@ export default function SettingsAppearanceTab({
...
+          Create a custom theme.
+        </Button>
+      </p>
+      )}
+      {theme[1] !== null && (
+        <ThemeCard className="mt-1" themeVars={theme[1]} selected={theme} setTheme={setTheme} openCustom={openCustom} />
```
adamstankiewicz, line 174: nit: just putting out another way to think about the `setTheme` approach here. wouldn't say one is better than other but leaving it as food for thought :) 

Rather than passing `setTheme` to `ThemeCard`, would it be simpler to reason about using callback? E.g., `ThemeCard` accepts an `onSelect` or `onClick` prop. Example:

```jsx
<ThemeCard onSelect={() => setTheme(theme[1])} {/* ... */} />
```

That way, `ThemeCard` doesn't have to know about `setTheme` and makes for a cleaner de-coupling between these components.

src/components/settings/SettingsAppearanceTab/index.jsx
```diff
@@ -1,48 +1,121 @@
...
-}) {
+export const SettingsAppearanceTab = ({
+  enterpriseId, enterpriseBranding, updatePortalConfiguration,
+}) => {
   const logoMessage = 'Your logo will appear on the upper left of every page for both learners and administrators. For best results, use a rectagular logo that is longer in width and has a transparent or white background.';
   const themeMessage = 'Select designer curated theme colors to update the look and feel of your learner and administrator experiences, or create your own theme.';
   const [configChangeSuccess, setConfigChangeSuccess] = useState(null);
-  const [theme, setTheme] = useState(SCHOLAR_THEME);
+  const [uploadedFile, setUploadedFile] = useState(undefined);
+  const [customIsOpen, openCustom, closeCustom] = useToggle(false);
```
adamstankiewicz, line 23: nit: `customModalIsOpen` to be a tad more explicit?

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+      <ModalDialog.Body>
+        <Form>
+          <Form.Group controlId="custom-button">
+            <Form.Control
+              floatingLabel="Button color"
+              isInvalid={!buttonValid}
+              onChange={(e) => {
+                setButton(e.target.value);
+                validateColor(e.target.value, setButtonValid);
```
adamstankiewicz, line 67: [curious] While we don't have a true color picker component, it'd be pretty trivial to at least show a colored square or circle or something with the inputted hex value for each `Form.Control` as a quick preview for the admin to sanity check its indeed the correct color they intended.

adamstankiewicz: Maybe a follow-on task, just food for thought 😃

src/components/settings/SettingsAppearanceTab/index.jsx
adamstankiewicz: Should we use the `size="inline"` prop instead?

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
kiram15: Because we have the icon and text as two containers in a flexbox, I want the hyperlink to be considered the same part of the text as the regular words, not its own separate container in the flex box.

src/components/BrandStyles/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+  enterpriseBranding: null,
+};
+
+BrandStyles.propTypes = {
+  enterpriseBranding: PropTypes.shape({
+    enterpriseId: PropTypes.string,
+    logo: PropTypes.string,
+    primary_color: PropTypes.string,
+    secondary_color: PropTypes.string,
+    tertiary_color: PropTypes.string,
```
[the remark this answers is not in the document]
kiram15: So I thought about that, and I think the way we would do it is to read primary, secondary, tertiary all in its own variables in portal configuration, which I just think is a worse way to do it, I like branding being grouped under one thing. But understand if you disagree, I can make the change

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
kiram15: Ok I think this is a good idea, but also I kind of want to make a ticket to do this will all forms in admin-portal....

src/components/settings/SettingsAppearanceTab/CustomThemeModal.jsx
kiram15: I think this is a good idea, but I'll include it in my next ticket with adding accessibility checks to the custom modal :)

alex-sheehan-edx: love the idea!

src/components/settings/SettingsAppearanceTab/ThemeCard.jsx
```diff
@@ -1,40 +1,71 @@
...
+    if (theme[0].title === CUSTOM_THEME_LABEL) {
+      setTheme([SCHOLAR_THEME, null]);
+    } else {
+      setTheme([theme[0], null]);
+    }
+  }
 
-export default function ThemeCard({ themeVars, selected, setTheme }) {
   return (
-    <Card style={{ width: '18rem' }} isClickable>
```
alex-sheehan-edx, line 7: 👍

alex-sheehan-edx (approved):
As long as all of Adam's comments are addressed I'm G2G

PR #855, opened 2022-09-12
feat: added new product tour for portal appearance feature

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-6100

Added product tour for portal appearance. Placed behind portal appearance feature flag.

Screen Shot 2022-09-09 at 2 55 37 PM

Notes for testing locally- if you missed the pop up you will most likely have to remove the `portal appearance seen` cookie which will be listed under localhost cookies. 

(Don't mind the cat logo)


kiram15 (approved):
Once you add some more tests this all LGTM!

kiram15 (changes requested):
One quick question though, when I navigate to settings, Portal Appearance is showing up even though I don't have the feature flag parameter in the url. Either we shouldn't merge this in until after we finish all the tickets, or this feature needs to recognize that feature flag

PR #860, opened 2022-09-19
feat: Filter out bad data for LCM table

brobro10000:
# For all changes
https://2u-internal.atlassian.net/browse/ENT-6260

Filter enterprise_enrollment_id for non-null items for DataTable sort functionality in LCM
Relabel null userEmail to empty string to remove DataTable required error and retain functionality of EmailAddressTableCell 'Email Hidden' label in LCM. 

- [X] Ensure adequate tests are in place (or reviewed existing tests cover changes)

- [NA] Ensure to attach screenshots
- [NA] Ensure to have UX team confirm screenshots


macdiesel (approved):
I think this is OK? Take a double check and make sure we don't check for null elsewhere to style the row or some other trickery.

package.json
```diff
@@ -66,6 +66,7 @@
     "regenerator-runtime": "0.13.7",
     "universal-cookie": "4.0.4",
     "url": "0.11.0",
+    "uuid": "^9.0.0",
```
adamstankiewicz, line 69: nit: let's keep this dependency pinned like the other dependencies in this file (i.e., remove the `^`)

src/components/learner-credit-management/data/utils.js
```diff
@@ -41,19 +41,24 @@ export const transformOfferSummary = (offerSummary) => {
...
  *
  * @param {array} results List of raw enrollment results from API.
  *
  * @returns List of transformed results for display in table.
  */
 export const transformUtilizationTableResults = results => results.map(result => ({
   created: result.created,
-  enterpriseEnrollmentId: result.enterpriseEnrollmentId,
-  userEmail: result.userEmail,
+  enterpriseEnrollmentId: result.enterpriseEnrollmentId === null ? parseInt(uuidv4().replace(/[^\d]+/g, getRandom(0, 9)).slice(0, 12), 10) : result.enterpriseEnrollmentId,
```
adamstankiewicz, line 60: Could this be written as `result.enterpriseEnrollmentId || uuidv4()`? I guess I'm not exactly following why we'd need `getRandom` if we're using `uuidv4`?

src/components/learner-credit-management/data/utils.js
```diff
@@ -41,19 +41,24 @@ export const transformOfferSummary = (offerSummary) => {
...
+  userEmail: result.userEmail === null ? '' : result.userEmail,
```
adamstankiewicz, line 61: [clarification] Why is it necessary to parse `userEmail` as an empty string here vs. keeping it `null`?

macdiesel:
FWIW (and I think you are already headed down this path) I think that it would be preferable to do the simplest thing possible here. Lets no parse things, I would be fine with you just selecting any random number or just using an auto-incrementing integer. 

Less is almost always more. Except when you need more.

PR #865, opened 2022-09-30
fix: adding error reporting functionality to LMS integrations

kiram15:
Three JIRA tickets are included in this PR 
[ENT-6214](https://2u-internal.atlassian.net/browse/ENT-6214) **|**  [ENT-6221](https://2u-internal.atlassian.net/browse/ENT-6221) **|** [ENT-6371](https://2u-internal.atlassian.net/browse/ENT-6371)

Screen Shot 2022-09-30 at 4 28 17 PM

Screen Shot 2022-09-30 at 4 28 26 PM

# For all changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@babel/plugin-transform-runtime": "7.12.1",
-        "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
+        "@edx/brand": "npm:@edx/brand-edx.org@^2.0.7",
         "@edx/frontend-enterprise-catalog-search": "3.1.0",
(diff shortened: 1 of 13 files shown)


alex-sheehan-edx (approved):
Base LMS integration landing page works and looks great. For the reporting modal- I had a weird testing edge case where course title/other fields were null, do we want default values for the table in those cases? Or is that something we don't need to bother with?

alex-sheehan-edx:
I would also love to be able to search the sync history modal table by course key!

kiram15:
> Base LMS integration landing page works and looks great. For the reporting modal- I had a weird testing edge case where course title/other fields were null, do we want default values for the table in those cases? Or is that something we don't need to bother with?

I feel like we have to ask John/Jenn on this one, but since this is just the base we can add in that functionality later. Also you can filter on course key! You just have to click the filter button

PR #870, opened 2022-10-11
feat: add in csv download button

kiram15:
[Jira Ticket
](https://2u-internal.atlassian.net/browse/ENT-6212)
Screen Shot 2022-10-11 at 1 55 56 PM

Screen Shot 2022-10-11 at 1 55 46 PM

# For all changes

### DIFF
--- a/src/components/settings/SettingsLMSTab/ErrorReporting/DownloadCsvButton.jsx
+++ b/src/components/settings/SettingsLMSTab/ErrorReporting/DownloadCsvButton.jsx
@@ -1,0 +1,93 @@
+import React, { useState, useEffect } from 'react';
+import { saveAs } from 'file-saver';
+import PropTypes from 'prop-types';
+import {
+  Toast, StatefulButton, Icon, Spinner, useToggle,
+} from '@edx/paragon';
+import { Download, Check } from '@edx/paragon/icons';
+
+const DownloadCsvButton = ({ data }) => {
+  const [buttonState, setButtonState] = useState('pageLoading');
+
+  useEffect(() => {
+    if (data !== null && data.length !== 0) {
+      setButtonState('default');
+    }
+  }, [data]);
+
+  const getCsvFileName = () => {
+    const currentDate = new Date();
+    const year = currentDate.getUTCFullYear();
+    const month = currentDate.getUTCMonth() + 1;
+    const day = currentDate.getUTCDate();
+    return `${year}-${month}-${day}-error-log.csv`;
+  };
+
+  const getCsvData = () => {
+    if (!Array.isArray(data) || !data.every((p) => typeof p === 'object' && p !== null)) {
+      return null;
+    }
+    const heading = Object.keys(data[0]).join(',');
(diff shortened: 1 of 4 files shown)


src/components/settings/SettingsLMSTab/ErrorReporting/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const month = currentDate.getUTCMonth() + 1;
+    const day = currentDate.getUTCDate();
+    return `${year}-${month}-${day}-error-log.csv`;
+  };
+
+  const getCsvData = () => {
+    if (!Array.isArray(data) || !data.every((p) => typeof p === 'object' && p !== null)) {
+      return null;
+    }
+    const heading = Object.keys(data[0]).join(',');
```
kiram15, line 30: Right now the heading is dynamic and not "pretty" because we assume staff might actually want to use the variable names.

src/components/settings/SettingsLMSTab/ErrorReporting/utils.jsx
```diff
@@ -0,0 +1,25 @@
+import * as timeago from 'timeago.js';
+import { CheckCircle, Error, Sync } from '@edx/paragon/icons';
+
+timeago.register('time-locale');
+
+export function isStrictlyArray(data) {
+  if (!Array.isArray(data) || !data.every((p) => typeof p === 'object' && p !== null)) {
+    return false;
+  }
+  return true;
```
marlonkeating, line 10: nit: Can save a few lines!
```suggestion
  return Array.isArray(data) && data.every((p) => typeof p === 'object' && p !== null);
```

src/components/settings/SettingsLMSTab/ErrorReporting/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,94 @@
...
+const DownloadCsvButton = ({ data }) => {
+  const [buttonState, setButtonState] = useState('pageLoading');
+
+  useEffect(() => {
+    if (data && data.length) {
+      setButtonState('default');
+    }
+  }, [data]);
+
+  const getCsvFileName = () => {
```
alex-sheehan-edx, line 19: is it worth to include channel code/config name/some other identifier in the CSV name? I'm not completely sure how I feel one way or the other, just a thought

alex-sheehan-edx (commented):
I have one open question and I agree with @marlonkeating's feedback! Otherwise thijs looks great 😄

src/components/settings/SettingsLMSTab/ErrorReporting/DownloadCsvButton.jsx
johnnagro: this is good feedback - i'd say for V1 lets not worry about additional columns. we are eventually going to power this by the backend and add a bunch more columns that dont appear in the view and we may add this stuff then.

johnnagro (approved):

PR #899, opened 2022-11-18
Removed Table and StatusAlert deprecation of Paragon

abdullahwaheed:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102)

### What has changed
Removed deprecated `Table` component of paragon and updated it with `DataTable`
Removed deprecated `StatusAlert` component of paragon and updated it with `Alert`

### References
[Paragon Table](https://paragon-openedx.netlify.app/components/table/)
[Paragon DataTable](https://paragon-openedx.netlify.app/components/datatable/)

[Paragon Alert](https://paragon-openedx.netlify.app/components/alert/)
[Paragon StatusAlert](https://paragon-openedx.netlify.app/components/statusalert/)

Based on #798


arbrandes:
@adamstankiewicz, another enterprisey one we need eyes on.

src/components/Admin/index.jsx
```diff
@@ -252,24 +252,30 @@ class Admin extends React.Component {
...
-        message={`Try refreshing your screen (${this.props.error.message})`}
-      />
+      <Alert
+        variant="danger"
+        icon={Error}
+      >
+        <Alert.Heading>
+          Unable to load overview
+        </Alert.Heading>
+        {`Try refreshing your screen (${this.props.error.message})`}
```
adamstankiewicz, line 262: nit: Render the alert message within a `<p>`, e.g.:

```jsx
<p>Try refreshing your screen ({this.props.error.message})</p>
```

src/components/Admin/index.jsx
```diff
@@ -252,24 +252,30 @@ class Admin extends React.Component {
...
-        title="Unable to Generate CSV Report"
-        message={`Please try again. (${message})`}
-      />
+        variant="danger"
+        icon={Error}
+      >
+        <Alert.Heading>
+          Unable to Generate CSV Report
+        </Alert.Heading>
+        {`Please try again. (${message})`}
```
adamstankiewicz, line 277: nit: Render the alert message within a `<p>`, e.g.:

```jsx
<p>Try refreshing your screen ({this.props.error.message})</p>
```

src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -136,20 +136,22 @@ export const BaseCourseSearchResults = (props) => {
...
-      <StatusAlert
-        alertType="danger"
-        iconClassName="fa fa-times-circle"
-        message={`${ERROR_MESSAGE} ${error.message}`}
-      />
+      <Alert
+        variant="danger"
+        icon={Error}
+      >
+        {`${ERROR_MESSAGE} ${error.message}`}
```
adamstankiewicz, line 143: nit: Render the alert message within a `<p>`, e.g.:

```jsx
<p>Try refreshing your screen ({this.props.error.message})</p>
```

src/components/CodeSearchResults/CodeSearchResults.test.jsx
```diff
@@ -77,12 +78,14 @@ describe('<CodeSearchResults />', () => {
         .create((
           <MemoryRouter>
             <Provider store={store}>
-              <CodeSearchResults
```
adamstankiewicz, line 80: [consideration] This might be a good opportunity to DRY up some of these tests, regarding the creating a reusable wrapper for `CodeSearchResults` without having to re-write `<MemoryRouter>`, `<Provider>`, and now, `<IntlProvider>` for each individual test.

src/index.scss
```diff
@@ -19,7 +19,6 @@ $fa-font-path: "~font-awesome/fonts";
 @import "./components/RequestCodesPage/RequestCodesPage";
 @import "./components/Sidebar/Sidebar";
 @import "./components/subscriptions/styles/";
-@import "./components/StatusAlert/StatusAlert";
 @import "./components/LoadingMessage/LoadingMessage";
 @import "./components/TableComponent/TableComponent";
```
adamstankiewicz, line 23: Could/should we also remove the custom `TableComponent` and `TableLoadingOverlay` styles here as well?

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
-import React from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
```
adamstankiewicz, line 1: This ESLint rule generally should not be ignored.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+              data={formatData(data)}
+              isSortable={tableSortable}
+              itemCount={totalCount}
+              isLoading={loading}
+              manualSortBy
+              fetchData={sortByColumn}
+            >
+              <DataTable.TableControlBar />
+              <DataTable.Table />
+              <DataTable.TableFooter>
```
adamstankiewicz, line 68: I believe we could rely on the stock `<DataTable.TableFooter />` here rather than defining our own. If we do need to define our own, we should be using the `DataTable`-specific pagination component as otherwise the pagination isn't properly wired up with how `DataTable` is supposed to be used.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+    <div className={className}>
+      <div className="row">
+        <div className="col">
+          <div className="table-responsive">
+            <DataTable
+              id={id}
+              className="table-sm table-striped"
+              columns={columns}
+              data={formatData(data)}
+              isSortable={tableSortable}
```
adamstankiewicz, line 60: Related to other questions about pagination, what about `isPaginated`?

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+  };
+
+  const renderTableContent = () => (
+    <div className={className}>
+      <div className="row">
+        <div className="col">
+          <div className="table-responsive">
+            <DataTable
+              id={id}
+              className="table-sm table-striped"
```
adamstankiewicz, line 57: `DataTable` itself has alternating striped rows already.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+              columns={columns}
+              data={formatData(data)}
+              isSortable={tableSortable}
+              itemCount={totalCount}
```
adamstankiewicz, line 61: nit: might recommend renaming `totalCount` -> `itemCount` for consistency

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+              isLoading={loading}
+              manualSortBy
```
adamstankiewicz, line 63: What about `manualPagination`?

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
-      loading,
-      enterpriseId,
-    } = this.props;
-
-    const sortByColumn = (column, direction) => {
+const TableComponent = ({
+  className, currentPage, pageCount, tableSortable, data, formatData, id, loading, enterpriseId, totalCount,
+  paginateTable, clearTable, columns, error, sortTable,
+}) => {
+  const sortByColumn = useCallback(args => {
```
adamstankiewicz, line 17: nit: if this function is passed to `fetchData`, it is technically responsible for more than just sorting; that is, it's responsible for fetching initial table data, sorting, pagination, filtering, etc.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+    const { sortBy } = args;
+    if (sortBy && sortBy.length > 0) {
+      const column = sortBy[0];
+      const ordering = `${column.desc ? '-' : ''}${column.id}`;
       updateUrl({
         page: 1,
-        ordering: direction === 'desc' ? `-${column.key}` : column.key,
+        ordering,
       });
+      sortTable(ordering);
```
adamstankiewicz, line 26: Ideally, our tables throughout the app would no longer need to dispatch Redux actions to do pagination/sorting/filtering, etc. but instead use what `DataTable` offers you via `fetchData` directly.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
-        {data && data.length > 0 && this.renderTableContent()}
-      </>
-    );
-  }
-}
+    </div>
+  );
+
+  const renderLoadingMessage = () => (
+    <TableLoadingSkeleton />
```
adamstankiewicz, line 86: We'd ideally rely on `DataTable`'s `isLoading` prop rather than defining our own loading state for the table at this point.

src/components/TableComponent/index.jsx
adamstankiewicz, line 60: Given that some tables support row selection (e.g., code management), I'm not seeing anything regarding the `isSelectable` prop here, which implies the selection logic is custom outside of `DataTable`; ideally, we'd rely on `DataTable`'s selection logic rather than defining our own.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+import { Pagination, DataTable, Alert } from '@edx/paragon';
+import { Error, WarningFilled } from '@edx/paragon/icons';
 import 'font-awesome/css/font-awesome.css';
 
 import TableLoadingSkeleton from './TableLoadingSkeleton';
-import TableLoadingOverlay from '../TableLoadingOverlay';
-import StatusAlert from '../StatusAlert';
 import { updateUrl } from '../../utils';
 
-class TableComponent extends React.Component {
```
adamstankiewicz, line 13: Long term, I'm not entirely sure the benefit of keeping `TableComponent`/`TableContainer` now that `DataTable` exists and am considering whether this refactor should move away from these components in favor of implementing `DataTable` directly where needed.

That way, we don't need to refactor/maintain a single table that supports every single possible table use case throughout the application vs. just implementing the bits and pieces of `DataTable` as needed for each individual use case instead.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+      There are no results.
+    </Alert>
+  );
+
+  return (
+    <>
+      {error && renderErrorMessage()}
+      {loading && !data && renderLoadingMessage()}
+      {!loading && !error && data && data.length === 0
+        && renderEmptyDataMessage()}
```
adamstankiewicz, line 113: Should we instead rely on `DataTable.EmptyTable` for the empty state messaging?

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
-                  tableId: id,
-                  page,
-                });
-              }}
-            />
+  useEffect(() => {
+    // Get initial data
+    paginateTable();
+    return clearTable;
+  }, []);
```
adamstankiewicz, line 39: `DataTable` calls `fetchData` on initial component mount to fetch any initial data. We shouldn't need a `useEffect` to trigger fetching initial data here.

src/components/TableComponent/index.jsx
```diff
@@ -1,166 +1,120 @@
...
+
+  const handlePageChange = (page) => {
```
adamstankiewicz, line 41: As mentioned else, pagination changes for `DataTable` should also come through the `fetchData` callback function rather than our existing custom implementation.

src/components/CodeSearchResults/CodeSearchResults.test.jsx
abdullahwaheed: Nice catch. but i think there are a lot of changes in this PR right now, we should create a separate PR for this

src/components/TableComponent/index.jsx
abdullahwaheed: its an old PR but as far as i remember, by adding those dependencies, the actual functionality was breaking. that's why i have added this ignore statement

abdullahwaheed:
this PR is outdated and there are a lot of merge conflicts. closing this one and create a new PR

PR #900, opened 2022-11-19
fix: ensure enterprise curation config loading state is set to false if feature is disabled

adamstankiewicz:
Ensures the loading state gets reset to `false` if `FEATURE_CONTENT_HIGHLIGHT` is disabled.

# For all changes


pwnage101 (approved):
Seems good to me!  Assuming the reverted code is just bypassed by the new condition that returns early.

adamstankiewicz:
> Seems good to me! Assuming the reverted code is just bypassed by the new condition that returns early.

@pwnage101 yep, the early return will prevent the `getOrCreateConfiguration` function from getting called when either an `enterpriseId` is missing or the highlights feature flag is disabled.

PR #901, opened 2022-11-19
chore: upgrade frontend-build and paragon to latest

adamstankiewicz:
* Upgrades from v11 to v12 of `@edx/frontend-build`, which upgrades ESLint to v8 and changes the standard ESLint rules.
* Moves away from our legacy `.eslintrc` config in favor of extending the shared one from `@edx/frontend-build`.
* Addresses new ESLint errors throughout the repo.
* Upgrades `@edx/paragon` to latest.

# For all changes


src/components/CodeReminderModal/index.jsx
```diff
@@ -214,35 +214,33 @@ export class BaseCodeReminderModal extends React.Component {
...
+        title={this.renderTitle()}
+        body={this.renderBody()}
+        buttons={[
+          <Button
+            key="remind-submit-btn"
+            disabled={submitting}
+            className="code-remind-save-btn"
+            onClick={handleSubmit(this.handleModalSubmit)}
+          >
+            <>
```
kiram15, line 229: Are we trying to get rid of all of these shorthand containers? Or just the ones that wrap an entire component?

src/components/CodeRevokeModal/index.jsx
```diff
@@ -232,36 +228,34 @@ class CodeRevokeModal extends React.Component {
...
+        title={this.renderTitle()}
+        body={this.renderBody()}
+        buttons={[
+          <Button
+            key="revoke-submit-btn"
+            disabled={submitting}
+            className="code-revoke-save-btn"
+            onClick={handleSubmit(this.handleModalSubmit)}
+          >
+            <>
```
kiram15, line 243: Same comment as above, maybe replace with a div or something?

src/components/FileInput/index.jsx
```diff
@@ -127,22 +127,20 @@ class FileInput extends React.Component {
...
+            <Button
+              variant="link"
+              className="remove-file-btn p-1 ml-2"
+              onClick={() => {
+                dispatch(clearFields(form, false, false, input.name));
+                input.onChange(null);
+                this.setState({ fileName: null });
+              }}
+            >
+              <>
```
kiram15, line 139: Unnecessary container wrapper

kiram15 (approved):
TBH this is a huge PR so I don't have a ton of the context on all these files, but tried to ask a couple questions! Other than a couple nits it looks good :)

__mocks__/react-instantsearch-dom.jsx
```diff
@@ -12,59 +13,71 @@ const advertised_course_run = {
   key: 'course-v1:edX+Bee101+3T2020',
 };
 
+/* eslint-disable camelcase */
 const fakeHits = [
```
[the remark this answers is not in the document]
adamstankiewicz: @kiram15 the ESLint rule was complaining about the snake_case `advertised_course_run` property below. ESLint _wants_ variables and property names to be camelCased, not snake_cased.

src/components/CodeReminderModal/index.jsx
adamstankiewicz: @kiram15 There's definitely still valid use cases for the React Fragments (`<></>`), like we see here on this line. We want to keep using it here because we want to use the `{...}` as a sibling to the "Remind" text, without getting wrapped by another DOM element (e.g., `div`).

In the case of the fragment that was removed, `Modal` was the only child element of the top-level fragment in this component so the fragment was essentially irrelevant/useless. 

The only reason to use a fragment at the top-level of a component is if the component has multiple sibling nodes at the top-level, but you don't want to include a wrapper `div` (or other element), e.g.:

```jsx
<>
  <Button>Hello</Button>
  <Button>World</Button>
</>
```

src/components/CodeRevokeModal/index.jsx
adamstankiewicz: @kiram15 Same comment as above, back at ya ;) This use of a fragment is valid, whereas the top-level fragment's usage was not.

src/components/CodeReminderModal/index.jsx
adamstankiewicz: @kiram15 I take that back actually. I think you're right that this Fragment could actually be removed, too (I had expected it to throw an ESLint error when I removed it just now, but it didn't!)

src/components/CodeRevokeModal/index.jsx
adamstankiewicz: @kiram15 And just like the other above comment's response, I take this one back too ;) I've since removed the fragment locally.

src/components/FileInput/index.jsx
adamstankiewicz: removed

PR #910, opened 2022-11-23
fix: Adding sync time data

kiram15:
https://2u-internal.atlassian.net/browse/ENT-6444
 - Added tab notification if there is evidence of an error on either the content or learner tab in the error reporting modal 
 - Changed toast text
 - Added last sync time info to card 

Screen Shot 2022-11-28 at 2 06 49 PM
Screen Shot 2022-11-28 at 10 33 58 AM
Screen Shot 2022-11-28 at 10 34 37 AM

# For all changes


src/components/settings/SettingsLMSTab/ErrorReporting/ErrorReportingModal.jsx
```diff
@@ -10,10 +10,11 @@ const ErrorReportingModal = ({
   isOpen, close, config, enterpriseCustomerUuid,
 }) => {
   const [key, setKey] = useState('contentMetadata');
-
+  const contentError = config?.lastContentSyncErroredAt == null ? '' : ' ';
+  const learnerError = config?.lastLearnerSyncErroredAt == null ? '' : ' ';
```
marlonkeating, line 14: clarification: Why single space for error, and empty string for no error?

src/components/settings/SettingsLMSTab/ExistingCard.jsx
```diff
@@ -50,7 +54,12 @@ const ExistingCard = ({
       setErrorModalText(errorToggleModalText);
       openError();
     } else {
-      onClick(TOGGLE_SUCCESS_LABEL);
+      // eslint-disable-next-line no-lonely-if
+      if (toggle) {
+        onClick(ACTIVATE_TOAST_MESSAGE);
+      } else {
+        onClick(INACTIVATE_TOAST_MESSAGE);
+      }
```
marlonkeating, line 62: nit: A bit terser option
```suggestion
      onClick(toggle ? ACTIVATE_TOAST_MESSAGE : INACTIVATE_TOAST_MESSAGE);
```

src/components/settings/SettingsLMSTab/ExistingCard.jsx
```diff
@@ -112,13 +121,23 @@ const ExistingCard = ({
     }
   };
 
+  const getLastSync = () => {
+    if (config.lastSyncErroredAt != null) {
+      const timeStamp = getTimeAgo(config.lastSyncErroredAt);
+      return <>Recent sync error:&nbsp; {timeStamp}<Icon className="small-icon text-danger-500" src={Error} /></>;
+    } if (config.lastSyncAttemptedAt != null) {
```
marlonkeating, line 128: nit: Don't strictly need this right now for flow reasons, but if we ever decide to make this function have only one return state in the future, can help avoid a bug.
```suggestion
    } else if (config.lastSyncAttemptedAt != null) {
```

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -89,26 +87,16 @@ const SettingsLMSTab = ({
...
-
     // If either the user has submit or canceled
-    if (input === '' || [SUCCESS_LABEL, DELETE_SUCCESS_LABEL, TOGGLE_SUCCESS_LABEL].includes(input)) {
+    if (input === '' || toastMessages.includes(input)) {
       // Re-fetch existing configs to get newly created ones
       fetchExistingConfigs();
     }
-    // If the user submitted
-    if (input === SUCCESS_LABEL) {
+    if (toastMessages.includes(input)) {
```
marlonkeating, line 95: Nice refactor!

src/components/settings/SettingsLMSTab/ErrorReporting/ErrorReportingModal.jsx
kiram15: This is the Tab component in paragon, which has [notification](https://paragon-openedx.netlify.app/components/tabs/#with-notification) as a subcomponent. Usually we use a string (like '1') to show notifications, which is why a string is passed and not just a boolean of if its present or not. But an empty string makes the notification disappear.

src/components/settings/SettingsLMSTab/ExistingCard.jsx
kiram15: The only thing is that with a return, the else isn't necessary, because if it falls into that, it means it didn't return from the previous if

marlonkeating (approved):

PR #895, opened 2022-11-15
feat: content highlights stepper updates, including refactor to use context selectors and connects to Algolia for course search results

brobro10000:
### v2

* Refactors to use `use-context-selector` for content highlights, which significantly reduces the performance impacts on component re-rendering when using our architectural approach to using vanilla React context; acts as a polyfill since official support for context selectors is still a proposal).
  * Note: This ultimately makes the reducer previously implemented unnecessary in favor of a more streamlined, state-like way to manage state than dispatching reducer actions.
  * Readings:
    * https://github.com/dai-shi/use-context-selector
    * https://github.com/reactjs/rfcs/pull/119
* Continues on with the title step validation logic. Adds a validation message to the `Stepper` step description.
* Persists data between the `Stepper` steps.
* Disables `DataTable` row selection when the cap is reached.
* Connects the "Select content" and "Confirm and and publish" steps with Algolia.
  * _Note: requires configuring Algolia API environment variables. I'm using the production Algolia index for now (since stage course data is janky) with a hardcoded (test) production enterprise customer UUID. Recommended to use `.env.private` to set these environment variables for local development._
* Sends `POST` to `/api/v/1/highlight-sets-admin/` on final CTA click from "Confirm and publish" step to create highlight, prepending it to any existing highlight sets in the `EnterpriseAppContext.enterpriseCuration` data without needing to refresh from API.
* Depends on upstream Paragon changes:
  * https://github.com/openedx/paragon/pull/1804
* Depends on the Algolia index having `aggregation_key` configured as an attribute for faceting (**PR incoming**).

### v1

Added Context, and refactored test/code to pass with context values for title and stepper logic
Initial work on data persist, more work is needed on it at this point
Set up reducer for persisting stepper data
Added logic to title to reflect limit on length of highlight title
Added additional UI changes to match language of the figma 
Added reducer for 'Last deleted item' to be used in toast on deletion

# For all changes

- [ ] Ensure adequate tests are in place (or reviewed existing tests cover changes)

# Only if submitting a visual change

- [ ] Ensure to attach screenshots
- [ ] Ensure to have UX team confirm screenshots

PR #912, opened 2022-12-05
feat: Add highlight content card to stepper and individual sets

brobro10000:
Added initial cards for user selection in the highlight stepper content review section.  
Added Price at the footer of the cards.
Added API integration with content highlight set items, displaying content highlight set items from highlight set.
Added Trashcan to remove items on the final step of the stepper
Added hyperlink on titles of cards
Added parameters/logics to take into account multiple content types for highlight cards

Added Relevant testing

Updating test reflecting changes and from reference PR https://github.com/openedx/frontend-app-admin-portal/pull/895

Highlight Set items
Screen Shot 2022-12-14 at 11 22 13 AM

Step 2 of Highlight Stepper
Screen Shot 2022-12-14 at 11 24 19 AM

Step 3 of Highlight Stepper
Screen Shot 2022-12-14 at 11 24 40 AM

# For all changes


src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -182,7 +182,7 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
           title={STEPPER_STEP_LABELS.CONFIRM_PUBLISH}
           index={steps.indexOf(STEPPER_STEP_LABELS.CONFIRM_PUBLISH)}
         >
-          <HighlightStepperConfirmContent />
+          <HighlightStepperConfirmContent enterpriseId={enterpriseId} />
```
adamstankiewicz, line 185: nit: One of the reasons `enterpriseId` is still in the global Redux store is to avoid needing to prop drill `enterpriseId` down the component tree.

If a subcomponent needs `enterpriseId`, we should rely more on pulling it from the Redux store instead of passing it via props to sub-components (e.g., look at the `mapStateToProps` and `connect` usages)

src/components/ContentHighlights/HighlightStepper/ContentSearchResultCard.jsx
```diff
@@ -11,13 +11,16 @@ const ContentSearchResultCard = ({ original }) => {
     cardImageUrl,
     originalImageUrl,
   } = original;
-
+  const extras = {
+    firstEnrollablePaidSeatPrice: original.firstEnrollablePaidSeatPrice,
+  };
```
adamstankiewicz, line 16: [curious] Any reason why we couldn't pass a `price` prop to `ContentHighlightCardItem` vs. passing this `extras` prop as an object?

```jsx
<ContentHighlightCardItem
  title={title}
  contentType={contentType}
  partners={partners}
  cardImageUrl={cardImageUrl || originalImageUrl}
  price={firstEnrollablePaidSeatPrice}
/>
```

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -43,18 +45,40 @@ const BaseReviewContentSelections = ({
...
+      };
+      original.extras = {
+        firstEnrollablePaidSeatPrice,
+      };
+      return original;
+    });
+    return cardData;
+  };
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
adamstankiewicz, line 69: Generally a code smell if you have to disable this ESLint rule. Hooks need to happen consistently across all renders, and this code here is after a conditional above. 

[curious] Any particular reason why you're passing `contentCards()` as React state if it never gets updated?

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -69,7 +69,7 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
       const newHighlightSet = {
         title: highlightTitle,
         isPublished: true,
-        // TODO: pass along the selected content keys!
+        content_key: Object.keys(currentSelectedRowIds).map(key => key.split(':')[1]),
```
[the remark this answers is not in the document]
brobro10000: Leaving Open as reference in the future

src/components/ContentHighlights/ContentHighlightCardItem.jsx
```diff
@@ -1,36 +1,49 @@
...
-        logoSrc={cardLogoSrc}
-        logoAlt={cardLogoAlt}
+        logoSrc={cardInfo.cardLogoSrc}
+        logoAlt={cardInfo.cardLogoAlt}
       />
       <Card.Header
-        title={<Truncate lines={3} title={title}>{title}</Truncate>}
-        subtitle={<Truncate lines={2} title={cardSubtitle}>{cardSubtitle}</Truncate>}
+        title={cardInfo.cardTitle}
+        subtitle={<Truncate lines={2} title={cardInfo.cardSubtitle}>{cardInfo.cardSubtitle}</Truncate>}
```
adamstankiewicz, line 40: [curious] Why is the `Truncate` component part of `cardInfo.cardTitle` but not `cardInfo.cardSubtitle`? I feel these should be consistent. E.g., maybe the variable stored in `cardInfo.cardTitle` is only the raw text and the `Truncate` component is rendered within the `title` prop itself, similar to how `subtitle` is currently set up.

src/components/ContentHighlights/ContentHighlightsCardItemsContainer.jsx
```diff
@@ -1,18 +1,14 @@
...
-const ContentHighlightsCardItemsContainer = () => {
-  const [highlightCourses] = useState(
-    camelCaseObject(TEST_COURSE_HIGHLIGHTS_DATA)[0]?.highlightedContent,
-  );
-
-  if (!highlightCourses) {
-    return null;
+const ContentHighlightsCardItemsContainer = ({ enterpriseSlug, isLoading, highlightedContent }) => {
+  if (!highlightedContent || highlightedContent?.length === 0) {
+    return <div data-testid="empty-highlighted-content" />;
```
adamstankiewicz, line 10: Rather than rendering an empty `div` here, I think it'd be worth rendering some sort of reasonable "empty state"  here, even though it theoretically shouldn't happen since you need to create a highlight set with some initial content.

Rendering an `Alert` in this situation would at least indicate something in the UI vs. a blank space, e.g.:

```jsx
<Alert variant="warning">
  There is no highlighted content for this highlight collection.
</Alert>
```

src/components/ContentHighlights/ContentHighlightsCardItemsContainer.jsx
```diff
@@ -21,19 +17,43 @@ const ContentHighlightsCardItemsContainer = () => {
         xl: 4,
```
adamstankiewicz, line 17: nit: should we align to using the same `columnSizes` from other components here? I.e., 

```jsx
columnSizes={{
  xs: 12,
  md: 6,
  lg: 4,
  xl: 3,
}}
```

[suggestion] maybe even moving this to a constant so that we can ensure the `CardGrids`  throughout all of `ContentHighlights` all have the same grid layout config?

src/components/ContentHighlights/ContentHighlightsCardItemsContainer.jsx
```diff
@@ -21,19 +17,43 @@ const ContentHighlightsCardItemsContainer = () => {
...
           key={uuid}
           cardImageUrl="https://picsum.photos/200/300"
           title={title}
-          contentType={contentType.toLowerCase()}
+          hyperlink={generateAboutPageUrl(enterpriseSlug, contentType?.toLowerCase(), contentKey)}
+          contentType={contentType?.toLowerCase()}
           partners={authoringOrganizations}
         />
       ))}
+      {isLoading && <div data-testid="card-item-skeleton" />}
```
adamstankiewicz, line 34: [curious] Should this be rendered higher up in the component? Example:

```jsx
const ContentHighlightsCardItemsContainer = ({ enterpriseSlug, isLoading, highlightedContent }) => {
  if (isLoading) {
    // render loading skeletons
  }

  if (!highlightedContent || highlightedContent?.length === 0) {
    // render reasonable empty state UI, e.g. warning `Alert`
  }

  return (
    <CardGrid>
      {highlightedContent.map(({
        uuid, title, contentType, authoringOrganizations, contentKey,
      }) => { /* ... */ })}
    </CardGrid>
  );
```

adamstankiewicz: generally, loading skeletons should mimic the general styles for the content that will once loading is complete, in this case skeleton cards.

Perhaps re-use/adapt the following from `HighlightStepperConfirmContent`:

```jsx
return (
  <div data-testid="skeleton-container">
    <CardGrid
      columnSizes={{
        xs: 12,
        md: 6,
        lg: 4,
        xl: 3,
      }}
    >
      {[...new Array(8)].map((element, index) => <SkeletonContentCard key={index} />)}
    </CardGrid>
  </div>
);
```

src/components/ContentHighlights/ContentHighlightCardItem.jsx
```diff
@@ -1,36 +1,49 @@
...
+    cardLogoAlt: partners?.length === 1 ? `${partners[0].name}'s logo` : undefined,
+    cardLogoSrc: partners?.length === 1 ? partners[0].logoImageUrl : undefined,
+    cardSubtitle: partners?.map(p => p.name).join(', '),
+    cardFooter: getContentHighlightCardFooter(price, contentType),
+  };
+  if (hyperlink) {
+    cardInfo.cardTitle = (
+      <Hyperlink destination={hyperlink} target="_blank">
+        <Truncate lines={3} title={title}>{title}</Truncate>
+      </Hyperlink>
```
adamstankiewicz, line 27: Hm, not sure a way around this, but it looks like the external link icon wraps onto a 4th line, since the text truncation doesn't consider the `Hyperlink` icon. And it looks like we can't have `Truncate` wrap `Hyperlink` instead.

image

I think for the purpose of this PR we should leave it as you have it, but let's confirm with Alena if this is OK. I personally feel having the external link icon sometimes wrap to a 4th line doesn't look all that clean and would possibly prefer a different approach to indicating opening in a new tab, e.g.:

image

Of course, let's defer to the decision is by Alena.

src/components/ContentHighlights/ContentHighlightSet.jsx
```diff
@@ -1,12 +1,19 @@
...
-  <Container fluid className="mt-5">
-    <CurrentContentHighlightItemsHeader />
-    <ContentHighlightsCardItemContainer />
-  </Container>
-);
+const ContentHighlightSet = () => {
+  const { highlightSetUUID } = useParams();
+  const { highlightSet, isLoading } = useHighlightSet(highlightSetUUID);
+  return (
+    <Container fluid className="mt-5">
```
adamstankiewicz, line 12: nit: IIRC `Container` is fluid by default.

src/components/ContentHighlights/CurrentContentHighlightItemsHeader.jsx
```diff
@@ -1,19 +1,20 @@
...
-
-  const highlightTitle = highlightSetUUID;
-
-  const titleName = `${highlightTitle} - Highlights`;
-
+const CurrentContentHighlightItemsHeader = ({ isLoading, highlightTitle }) => {
+  if (isLoading) {
+    return (
+      <div data-testid="header-skeleton">
+        <Skeleton />
```
adamstankiewicz, line 11: nit: perhaps something along the following instead?

```jsx
<ActionRow>
  <h2><Skeleton /></h2> // rendering `Skeleton` as child of `h2` means the `Skeleton` will inherit the `h2` styles (e.g., its height).
  <ActionRow.Spacer />
  <Skeleton />
</ActionRow>
```

such that the skeleton state more closely matches the styles that will replace the skeletons (i.e., an `h2` and a CTA button to the right).

src/components/ContentHighlights/data/constants.js
```diff
@@ -1,5 +1,17 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
+/* eslint-disable import/no-extraneous-dependencies */
 import { faker } from '@faker-js/faker';
+import { configuration } from '../../../config';
+/* eslint-enable import/no-extraneous-dependencies */
+
+// Generate URLs for about pages from the enterprise learner portal
+export const generateAboutPageUrl = (enterpriseSlug, contentType, contentKey) => {
```
adamstankiewicz, line 7: nit: might recommend changing the function signature here to use an object instead, so that when you pass the args, it's more obvious which argument is for what, e.g.:

```jsx
const generateAboutPageUrl = ({ enterpriseSlug, contentType, contentKey }) => { ... };
```

```jsx
generateAboutPageUrl({
  enterpriseSlug: 'sluggy',
  contentType: 'course',
  contentKey: 'edX+DemoX',
});
```

src/components/ContentHighlights/data/constants.js
```diff
@@ -28,7 +40,19 @@ export const FOOTER_TEXT_BY_CONTENT_TYPE = {
   learnerpathway: 'Pathway',
 };
 
-// Test Data for Content Highlights
+// High Card logic for footer text
+export const getContentHighlightCardFooter = (price, contentType) => {
```
adamstankiewicz, line 44: nit: This isn't so much a constant, but a utility function. Would it make sense to move into a `utils.js` file? Perhaps also consider `export const getContentHighlightCardFooter = ({ price, contentType }) => { ... };`

src/components/ContentHighlights/data/constants.js
adamstankiewicz: Similar to a comment below; this is more a util function than a constant; should it be moved to a `utils.js` file?

src/components/ContentHighlights/data/constants.js
```diff
@@ -28,7 +40,19 @@ export const FOOTER_TEXT_BY_CONTENT_TYPE = {
...
+    return formattedContentType;
+  }
+  return `$${price} · ${formattedContentType}`;
+};
+
+// Test Data for Content Highlights TODO: Remove when API is available || Remove when feature completed
+// Test entepriseId for Content Highlights to display card selections and confirmation
+export const testEnterpriseId = 'e783bb19-277f-479e-9c41-8b0ed31b4060';
+// Test Content Highlights data
 export const TEST_COURSE_HIGHLIGHTS_DATA = [
```
adamstankiewicz, line 56: [curious] Is this constant still in use anywhere at this point? If not, should it be removed?

src/components/ContentHighlights/data/constants.js
```diff
@@ -306,3 +330,68 @@ export const TEST_COURSE_HIGHLIGHTS_DATA = [
     ],
   },
 ];
+
+export const testCourseData = [
```
adamstankiewicz, line 334: [clarification] Is this only used by tests?

src/components/ContentHighlights/data/constants.js
```diff
@@ -306,3 +330,68 @@ export const TEST_COURSE_HIGHLIGHTS_DATA = [
...
+        logoImageUrl: 'https://prod-discovery.edx-cdn.org/organization/logos/44022f13-20df-4666-9111-cede3e5dc5b6-2cc39992c67a.png',
+      },
+    ],
+    cardImageUrl: 'https://prod-discovery.edx-cdn.org/media/course/image/3a31db71-de8f-45f1-ae65-11981ed9d680-31634d40b3bb.small.png',
+    originalImageUrl: 'https://prod-discovery.edx-cdn.org/media/course/image/3a31db71-de8f-45f1-ae65-11981ed9d680-b801bb328333.png',
+    firstEnrollablePaidSeatPrice: 199,
+  },
+];
+
+export const testCourseAggregation = {
```
adamstankiewicz, line 392: [clarification] Is this only used by tests?

src/components/ContentHighlights/data/hooks.js
```diff
@@ -30,12 +33,41 @@ export function useHighlightSetsForCuration(enterpriseCuration) {
...
+  };
+}
+
 /**
  * Defines an interface to mutate the `ContentHighlightsContext` context value.
  */
 export function useContentHighlightsContext() {
   const setState = useContextSelector(ContentHighlightsContext, v => v[1]);
-
+  const getState = useContextSelector(ContentHighlightsContext, v => v[0]);
```
adamstankiewicz, line 70: With `useContextSelector`, the 2nd argument is generally extracting the specific data from the state, not the entire state value itself. That way, it's smart enough to only subscribe to changes from a "slice" of state, rather than the entire state value, e.g.:

```jsx
const currentSelectedRowIds = useContextSelector(ContentHighlightsContext, v => v[0]. stepperModal.currentSelectedRowIds);
```

src/components/ContentHighlights/data/hooks.js
```diff
@@ -68,6 +100,18 @@ export function useContentHighlightsContext() {
     }));
   }, [setState]);
 
+  const deleteSelectedRowId = useCallback((rowId) => {
+    const x = getState.stepperModal.currentSelectedRowIds;
+    delete x[rowId];
```
adamstankiewicz, line 105: If we're need to rely on the previous state to set the updated state here, we should rely on the `prevState` argument in the callback function passed to `setState` per the React best practices. Since state setting is asynchronous, the `getState.stepperModal.currentSelectedRowIds` _could_ be different from the prev state coming through as the `s` argument in `setState`'s callback function below.

Perhaps consider something along the lines of:

```jsx
const deleteSelectedRowId = useCallback((rowId) => {
  setState((s) => {
    const updatedCurrentSelectedRowIds = s.stepperModal.currentSelectedRowIds;
    delete updatedCurrentSelectedRowIds[rowId];
    return {
      ...s,
      stepperModal: {
        ...s.stepperModal,
        currentSelectedRowIds: updatedCurrentSelectedRowIds,
      },
    };
  });
}, [setState]);
```

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -11,50 +13,71 @@ import {
...
   }
-
   if (!searchResults) {
-    return null;
+    return (<div data-testid="base-content-no-results" />);
   }
-
   const { hits } = camelCaseObject(searchResults);
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
adamstankiewicz, line 53: Is this comment to disable the rule still relevant?

src/components/ContentHighlights/ContentHighlightCardItem.jsx
brobro10000: I was debating leaving `cardInfo` for just data with altered logic or whether to include all card info.  I'll take this as confirmation that the `subtitle` also needs to be pulled into the `cardInfo` object

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -104,10 +130,11 @@ const SelectedContent = () => {
       filterString += ')';
     }
     return filterString;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
adamstankiewicz, line 133: [curious] Why is it necessary to disable this ESLint rule? Generally a code smell.

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -104,10 +130,11 @@ const SelectedContent = () => {
...
   }, [currentSelectedRowIds]);
 
   if (currentSelectedRowIds.length === 0) {
-    return null;
+    return (<div data-testid="selected-content-no-results" />);
```
adamstankiewicz, line 137: Similar to another comment, rather than rendering an empty div here, it may be better to render _something_ in the UI (e.g., in an error warning `Alert`: "You don't have any highlighted content selected. Go back to the previous step to select content.").

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+    title,
+    contentType,
+    partners,
+    cardImageUrl,
+    originalImageUrl,
+    firstEnrollablePaidSeatPrice,
+    aggregationKey,
+    key,
+  } = original;
+  // eslint-disable-next-line no-shadow
```
adamstankiewicz, line 24: [curious] why this disabling this ESLint rule necessary?

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+    if (deleteKey === aggregationKey) {
+      deleteSelectedRowId(deleteKey);
+      setIsDeleted(true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [deleteKey, aggregationKey]);
+  if (isDeleted) {
+    return (
+      null
+    );
```
adamstankiewicz, line 35: [curious] Might need some more clarity on what's happening here. When the selected row id is removed via `deleteSelectedRowId`, wouldn't the `algoliaFilters` passed to `<Configure />` get re-computed to not include the now-deleted content item, where this card would no longer be rendered regardless of the internal `isDeleted` state?

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+    partners,
+    cardImageUrl,
+    originalImageUrl,
+    firstEnrollablePaidSeatPrice,
+    aggregationKey,
+    key,
+  } = original;
+  // eslint-disable-next-line no-shadow
+  useEffect(() => {
+    if (deleteKey === aggregationKey) {
```
adamstankiewicz, line 26: [clarification] Why is the `useEffect` necessary vs. using a callback function? E.g.

```jsx
<IconButton
  invertColors
  isActive
  src={Delete}
  iconAs={Icon}
  alt="Delete"
  onClick={() => deleteSelectedRowId(aggregationKey)}
  className="ml-1 flex-shrink-0"
/>
```

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+        hyperlink={generateAboutPageUrl(enterpriseSlug, contentType?.toLowerCase(), aggregationKey?.split(':')[1])}
+        contentType={contentType}
+        partners={partners}
+        cardImageUrl={cardImageUrl || originalImageUrl}
+        price={firstEnrollablePaidSeatPrice}
+
+      />
+      <IconButton
+        invertColors
+        isActive
```
adamstankiewicz, line 50: [curious] Why use `invertColors` and `isActive` here?

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+        src={Delete}
+        iconAs={Icon}
+        alt="Delete"
```
adamstankiewicz, line 53: nit: a bit of an improvement in the `alt` text here might be something like, "`Remove ${title} from highlight collection`" so that each content card has a more explicit, unique `alt`.

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [deleteKey, aggregationKey]);
+  if (isDeleted) {
+    return (
+      null
+    );
+  }
+  return (
+    <React.Fragment key={key}>
```
adamstankiewicz, line 38: I'd probably recommend rendering a `div` vs. a `React.Fragment` here so there is some DOM element grouping the 2 together (they are related elements, so should be grouped in some way).

src/components/ContentHighlights/ContentHighlightCardItem.jsx
brobro10000: Going back to this again, i remembered why I structured it as so. In creating this `ContentHighlightCardItem`, I went with the mindset that not every component will be used if we hope to make the card component extensible in the future. So defining the title in the `Card.Header` as an object was done purposely for the `hyperlink` (it may or may not exist, giving users the object), while not defining the subtitle in the object due to its static nature in the `Card.Header` (no logic needed to define what goes in it beyond the `partners`.

src/components/ContentHighlights/ContentHighlightsCardItemsContainer.jsx
brobro10000: Decided to extract it out, could be a utility component in the future for cards, but basically created a new component called `SkeletonContentCardContainer` which contains the logic `HighlightStepperConfirmContent`

`
import SkeletonContentCard from './SkeletonContentCard';

const SkeletonContentCardContainer = ({ length }) => [
  ...new Array(length),
// eslint-disable-next-line react/no-array-index-key
].map((_, index) => <SkeletonContentCard key={index} />);

SkeletonContentCardContainer.propTypes = {
  length: PropTypes.number.isRequired,
};

export default SkeletonContentCardContainer;
`
This results in a much cleaner final implementation. Was tempted to add a `{children}` prop to map into instead of `SkeletonContentCard` but in this iteration I thought to keep it simple for now.

Final resulting condition is as such
`
  if (isLoading) {
    return (
      <CardGrid columnSizes={HIGHLIGHTS_CARD_GRID_COLUMN_SIZES}>
        <SkeletonContentCardContainer length={MAX_CONTENT_ITEMS_PER_HIGHLIGHT_SET} />
      </CardGrid>
    );
  }
  `

src/components/ContentHighlights/ContentHighlightCardItem.jsx
brobro10000: Leaving open for future reference

src/components/ContentHighlights/data/constants.js
brobro10000: I would advocate to keep it in the the repo for testing purposes. The enterpriseId allows the contents of the Highlights DataTable to be populated allowing for quick testing beyond the local environment with courses and existing meta data about courses.

Feel free to disagree, but I'd strongly encourage leaving it in.

src/components/ContentHighlights/data/constants.js
brobro10000: Yep, it reflects sample data coming from the algolia's search clients `hits` that have been destructured to Content cards metadata

src/components/ContentHighlights/data/constants.js
brobro10000: Yep, this represents the selectedRowIds between state changes of the stepper.

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: Using a div causes the items to render as the image below shows. I could of left an empty react fragment, but then the no key error occurs for each child element. React.Fragment was my solution to preventing both those issues.

Screen Shot 2022-12-16 at 4 20 54 PM

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: lol I don't know but they seem to be counteracting each other. I think inverting the isActive state makes it look like the default button, so I went with it without realizing that non-active state would of solved the issue.

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: Running through it without this code section it seems to be working. At some point I had floating cards with the container remaining without the cards reordering themselves filling in the empty gaps.

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: leftover remnants of poorly written code, currently has no effect removing it

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: I believe bad implementation of other parts of the code rendering the cards after it was deleted from the past required me to go down this route initially, which is now totally unnecessary. 

All suggested fixes regarding this page have been implemented and appears to be working correctly

src/components/ContentHighlights/data/hooks.js
brobro10000: After implementing the changes, the UI is showing the item is being removed but the test are failing now showing the content is not being removed. This is after implementing the changes ConfirmContentCard PR suggestions below. Putting the exact function above does not remove the items from the UI but removes them from the `selectedRowIds` updating the checkmark in step 2 but no re-rendering the page to show the removed items in the ConfirmContent Page. 

Might need to rubber duck on why this could be happening. It would also explain when I chose to return a null if the item `isDeleted` followed by a useEffect to match the deleted item to the aggregation key as a way to force a re-render after the item has been deleted.

brobro10000: Ended up reverting back to what I originally had in both `ContentConfirmContentCard` and `hooks`. Changing hooks resulted in an open gap when deleting items, while changing ContentConfirmContentCard resulted in failing test showing that the content still remained in the UI. Changing both resulted in the content not being removed at all. 
:(

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -69,7 +69,10 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
       const newHighlightSet = {
         title: highlightTitle,
         isPublished: true,
-        // TODO: pass along the selected content keys!
+        // TODO: This is a temporary solution to get the content keys from the
+        // selected row ids based on content_type (uuid, key etc.).
```
adamstankiewicz, line 73: FWIW, i think we can remove the `TODO` comment here for now. I don't imagine we'll change this anytime soon.

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -69,7 +69,10 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
+        content_keys: Object.keys(currentSelectedRowIds).map(key => key.split(':')[1]),
+        // content_keys: ['edX+DemoX'],
```
adamstankiewicz, line 75: nit: remove comment

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -78,7 +81,7 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
         isPublished: result.isPublished,
         title: result.title,
         uuid: result.uuid,
-        highlightedContentUuids: [],
+        highlightedContentUuids: result?.highlightedContentUuids || [],
```
adamstankiewicz, line 84: nit: `result` is always defined in the other object properties. why use `result?.` here?

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -7,54 +9,57 @@ import {
...
   }
-
   if (!searchResults) {
-    return null;
+    return (<div data-testid="base-content-no-results" />);
   }
-
   const { hits } = camelCaseObject(searchResults);
 
+  const transformedCardData = hits.map((highlightedContent) => {
```
adamstankiewicz, line 43: [curious] Is this variable necessary versus doing `hits.map(highlightedContent => <ContentConfirmContentCard key={highlightedContent.aggregationKey} original={highlightedContent} />)` on line 60 below?

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -92,7 +98,9 @@ const SelectedContent = () => {
    */
   /* eslint-enable max-len */
   const algoliaFilters = useMemo(() => {
-    let filterString = `enterprise_customer_uuids:${prodEnterpriseId}`;
+    // TODO: replace testEnterpriseId with enterpriseID before push,
+    // uncomment out import and replace with testEnterpriseId to test
+    let filterString = `enterprise_customer_uuids:${enterpriseId}`;
```
adamstankiewicz, line 103: 🙌

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.jsx
```diff
@@ -104,10 +112,14 @@ const SelectedContent = () => {
       filterString += ')';
     }
     return filterString;
-  }, [currentSelectedRowIds]);
+  }, [currentSelectedRowIds, enterpriseId]);
 
   if (currentSelectedRowIds.length === 0) {
-    return null;
+    return (
+      <Alert data-testid="selected-content-no-results">
```
adamstankiewicz, line 119: This `Alert` needs a `variant`, otherwise it has a white/transparent background. Try `variant="warning"`

src/components/ContentHighlights/data/constants.js
```diff
@@ -1,6 +1,22 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
+/* eslint-disable import/no-extraneous-dependencies */
```
adamstankiewicz, line 1: [curious] Why is this ESLint rule disabled/necessary?

src/components/ContentHighlights/data/hooks.js
```diff
@@ -30,12 +33,41 @@ export function useHighlightSetsForCuration(enterpriseCuration) {
   return highlightSets;
 }
 
+export function useHighlightSet(highlightSetUUID) {
+  const [highlightSet, setHighlightSetItems] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState(null);
+
+  const getHighlightSetItems = useCallback(async () => {
```
adamstankiewicz, line 41: nit: `getHighlightSet` since you're not just returning the highlight content associated with the `HighlightSet`, but rather all metadata about the `HighlightSet`

src/components/ContentHighlights/data/hooks.js
```diff
@@ -30,12 +33,41 @@ export function useHighlightSetsForCuration(enterpriseCuration) {
   return highlightSets;
 }
 
+export function useHighlightSet(highlightSetUUID) {
+  const [highlightSet, setHighlightSetItems] = useState([]);
```
adamstankiewicz, line 37: nit: `setHighlightSet`

src/components/ContentHighlights/data/utils.js
```diff
@@ -0,0 +1,20 @@
+import { configuration } from '../../../config';
+import { FOOTER_TEXT_BY_CONTENT_TYPE } from './constants';
+
+// High Card logic for footer text
```
adamstankiewicz, line 4: nit: "Highlight"?

src/components/ContentHighlights/data/utils.js
```diff
@@ -0,0 +1,20 @@
...
+export const getContentHighlightCardFooter = ({ price, contentType }) => {
+  const formattedContentType = FOOTER_TEXT_BY_CONTENT_TYPE[contentType?.toLowerCase()];
+  if (!price) {
+    return formattedContentType;
+  }
+  return `$${price} · ${formattedContentType}`;
+};
+// Generate URLs for about pages from the enterprise learner portal
```
adamstankiewicz, line 12: nit: new line between function and comment

src/components/ContentHighlights/HighlightStepper/ContentConfirmContentCard.jsx
brobro10000: Thought I did, now removed  👍🏼

src/components/ContentHighlights/HighlightSetSection.jsx
```diff
@@ -15,13 +16,7 @@ const HighlightSetSection = ({
...
     <div>
       <h3 className="mb-3">{sectionTitle}</h3>
-      <CardGrid
-        columnSizes={{
-          xs: 12,
-          lg: 6,
-          xl: 4,
-        }}
-      >
+      <CardGrid columnSizes={HIGHLIGHTS_CARD_GRID_COLUMN_SIZES}>
```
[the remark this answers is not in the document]
brobro10000: Leaving open for minor ego boost

src/components/ContentHighlights/data/constants.js
brobro10000: I had attempting to bring in another dependency beyond faker but had since removed it. But it still requires a es-lint disable on the next line at the very list to allow faker dependency to be brought into the folder.

Screen Shot 2022-12-19 at 12 47 57 PM

src/components/ContentHighlights/data/hooks.js
brobro10000: Thanks for the rubber ducking session to resolve this issue.

src/components/ContentHighlights/SkeletonContentCardContainer.jsx
```diff
@@ -0,0 +1,29 @@
+import PropTypes from 'prop-types';
+import { CardGrid } from '@edx/paragon';
+import { v4 as uuidv4 } from 'uuid';
+import SkeletonContentCard from './SkeletonContentCard';
+import { HIGHLIGHTS_CARD_GRID_COLUMN_SIZES } from './data/constants';
+
+const SkeletonContentCardContainer = ({ length, columnSizes }) => (
```
adamstankiewicz, line 7: nit: `length` -> `totalItems`

src/components/ContentHighlights/data/hooks.js
```diff
@@ -68,6 +101,18 @@ export function useContentHighlightsContext() {
     }));
   }, [setState]);
 
+  const deleteSelectedRowId = useCallback((rowId) => {
+    const currentRowIds = { ...currentSelectedRowState };
+    delete currentRowIds[rowId];
+    setState(s => ({
```
adamstankiewicz, line 107: I believe we can nix the `currentSelectedRowState` state variable in favor of something like below:

```jsx
const deleteSelectedRowId = useCallback((rowId) => {
    const currentRowIds = { ...currentSelectedRowState };
    setState((s) => {
      const currentRowIds = { ...s };
      delete currentRowIds[rowId];
      return {
       ...s,
       stepperModal: {
         ...s.stepperModal,
         currentSelectedRowIds: currentRowIds,
       },
      };
    });
});
```

adamstankiewicz (approved):

PR #920, opened 2022-12-16
fix: Remove updating width creating additional padding

brobro10000:
Ticket: https://2u-internal.atlassian.net/browse/ENT-6613

Removed updating width resulting in additional padding on content.
Screen Shot 2022-12-16 at 10 33 39 AM

# For all changes


src/components/EnterpriseApp/index.jsx
```diff
@@ -133,11 +133,6 @@ class EnterpriseApp extends React.Component {
                 <ProductTours />
                 <Sidebar
                   baseUrl={baseUrl}
-                  onWidthChange={(width) => {
-                    this.setState({
-                      sidebarWidth: width + defaultContentPadding,
-                    });
-                  }}
```
adamstankiewicz, line 140: Hm, so removing `onWidthChange` means the user can no longer use the hamburger menu icon in the header to persist the sidebar menu open. 

image

When the hamburger menu icon is clicked, it should treat the sidebar nav as an always-expanded menu with appropriate padding.

It might be worth checking what triggers the callback passed to `onWidthChange` in `Sidebar`. E.g., there is a `useEffect` with `onWidthChange` in its dependency array. But the callback function we're passing to `onWidthChange` in this component is not memoized so any re-render of `EnterpriseApp` would cause the callback function to have a new function reference, effectively triggering that `useEffect` in `Sidebar`.

I'd be curious if we abstracted this callback function out into a `useCallback` so the function reference persists on re-renders, and doesn't trigger that `useEffect` extra times.

adamstankiewicz (approved):

PR #928, opened 2022-12-21
fix: reverted back to original working toast

brobro10000:
Fixes original 6511 PR with functioning toast while retaining changes to fireEvent --> userEvent

# For all changes


src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -84,9 +84,9 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
         highlightedContentUuids: result.highlightedContent || [],
       };
       dispatchEnterpriseCuration(enterpriseCurationActions.addHighlightSet(transformedHighlightSet));
+      dispatchEnterpriseCuration(enterpriseCurationActions.setHighlightToast(transformedHighlightSet.uuid));
```
adamstankiewicz, line 87: May want to ensure this line gets test coverage (i.e., was `setHighlightToast` called with the expected `uuid`)?

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -47,6 +52,19 @@ function enterpriseCurationReducer(state, action) {
       return { ...state, enterpriseCuration: action.payload };
     case SET_FETCH_ERROR:
       return { ...state, fetchError: action.payload };
+    case SET_TOAST_TEXT: {
+      const existingHighlightSets = getHighlightSetsFromState(state);
+      const filteredHighlightSets = existingHighlightSets.filter(
+        highlightSet => highlightSet.uuid === action.payload,
+      );
```
adamstankiewicz, line 59: nit: considering we expect to only find 1 matching highlight set given the UUID passed in `action.payload`, should we opt for `existingHighlightSets.find` instead?

```jsx
const foundHighlightSet = existingHighlightSets.find(...);

return {
  ...state,
  enterpriseCuration: {
    ...state.enterpriseCuration,
    toastText: foundHighlightSet?.title,
  },
};
```

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -47,6 +52,19 @@ function enterpriseCurationReducer(state, action) {
       return { ...state, enterpriseCuration: action.payload };
     case SET_FETCH_ERROR:
       return { ...state, fetchError: action.payload };
+    case SET_TOAST_TEXT: {
```
adamstankiewicz, line 55: [curious] do we ever clear the `toastText` in this reducer? should we? (i defer to your judgement on this one for sure!)

brobro10000: Pushing for no on this only because within the `ContentHighlights` which contain the logic to render the `toasts` array, its dependency array is also the `enterpriseCuration.toastText`, we run the risk of rendering an empty/null toast.

adamstankiewicz: Makes sense to me!

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
adamstankiewicz: Deferring as a separate chunk of work since `handlePublish` doesn't currently have any coverage anyways.

adamstankiewicz (approved):

PR #936, opened 2023-01-05
feat: maintaining debug value on sso configs through self service

alex-sheehan-edx:
wonky backend default values meant not including debug = debug is false. this would happen every time you advance through the first step. Also would turn debug on whenever you advance through the last step. Now debug starts false and will maintain whatever value it has through the steps.

# For all changes


src/components/settings/SettingsSSOTab/SSOStepper.jsx
```diff
@@ -65,7 +65,7 @@ const SSOStepper = ({ enterpriseSlug, enterpriseId, enterpriseName }) => {
     // Values we want all provider configs to by default contain
     configFormData.append('enterprise_customer_uuid', enterpriseId);
     configFormData.append('enabled', true);
-    configFormData.append('debug_mode', true);
+    configFormData.append('debug_mode', providerConfig?.debug_mode || false);
```
mattdrayer, line 68: This seems like the right way to go.

We should do the same for the automatic refresh setting on line 73 -- `providerConfig?.automatic_refresh_enabled || false`

src/components/settings/SettingsSSOTab/SSOStepper.jsx
```diff
@@ -65,7 +65,7 @@ const SSOStepper = ({ enterpriseSlug, enterpriseId, enterpriseName }) => {
     // Values we want all provider configs to by default contain
     configFormData.append('enterprise_customer_uuid', enterpriseId);
     configFormData.append('enabled', true);
```
mattdrayer, line 67: Seems like the enabled/active setting shouldn't be forced to true -- if the config has been deactivated for some reason we probably should preserve that.  So maybe `providerConfig?enabled || true`?

src/components/settings/SettingsSSOTab/hooks.js
```diff
@@ -72,6 +72,8 @@ const useIdpState = () => {
     formData.append('sync_learner_profile_data', false);
     formData.append('enable_sso_id_verification', true);
 
+    formData.append('debug_mode', providerConfig?.debug_mode || false);
```
mattdrayer, line 75: I'm not seeing `automatic_refresh_enabled` in this list

mattdrayer (changes requested):

src/components/settings/SettingsSSOTab/hooks.js
alex-sheehan-edx: I can add it

src/components/settings/SettingsSSOTab/SSOStepper.jsx
alex-sheehan-edx: yup, happy to. Do we want it to default to true or false?

mattdrayer: I think it should default to false

src/components/settings/SettingsSSOTab/tests/NewSSOConfigForm.test.jsx
```diff
@@ -459,14 +459,15 @@ describe('SAML Config Tab', () => {
...
+      debug_mode: 'false',
       skip_hinted_login_dialog: 'true',
       skip_registration_form: 'true',
       skip_email_verification: 'true',
       send_to_registration_first: 'true',
-      automatic_refresh_enabled: 'true',
+      automatic_refresh_enabled: 'false',
     };
     mockUpdateProviderConfig.mock.calls[0][0].forEach((value, key) => {
+      console.log('WOW!', key);
```
mattdrayer, line 470: Should this be included in the merge or is it a dev artifact?

alex-sheehan-edx: hahaha oops!

PR #938, opened 2023-01-06
fix: update copy, icons, and use imported fallback img from @edx/brand for highlights

adamstankiewicz:
# For all changes


package.json
```diff
@@ -24,13 +24,13 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@babel/plugin-transform-runtime": "7.12.1",
-    "@edx/brand": "npm:@edx/brand-edx.org@^2.0.7",
+    "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
```
adamstankiewicz, line 27: The Open edX theme should always be installed by default for any MFE in the openedx Github organization. It gets aliased at build time in GoCD with `@edx/brand-edx.org`. These 2 packages share the same interface.

macdiesel (approved):

PR #946, opened 2023-01-10
fix: pass course run id instead of key for bulk enroll

adamstankiewicz:
# Description

For bulk enrollment, ensure the course run ID is passed instead of the course key. To do this, we take the Algolia results from the "Review step" and mutate the `selectedCourses` array in the `BulkEnrollContext` reducer to include the Algolia metadata such that `BulkEnrollmentSubmit` has access to `advertisedCourseRun` to get its course run key.

# For all changes

### DIFF
--- a/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
+++ b/src/components/BulkEnrollmentPage/CourseSearchResults.jsx
@@ -118,7 +118,6 @@ export const BaseCourseSearchResults = (props) => {
         id: rowId,
         values: {
           aggregationKey: rowId,
-          contentKey: rowId.split(':')[1],
         },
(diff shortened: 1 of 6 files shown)


src/components/BulkEnrollmentPage/CourseSearchResults.jsx
```diff
@@ -118,7 +118,6 @@ export const BaseCourseSearchResults = (props) => {
         id: rowId,
         values: {
           aggregationKey: rowId,
-          contentKey: rowId.split(':')[1],
```
adamstankiewicz, line 121: no longer needed

macdiesel (approved):

PR #949, opened 2023-01-10
feat: Add browser/modal confirmation for stepper exist

brobro10000:
Adds alert modal when user prematurely leaves stepper via the 'X' or back button without completing the steps to create a highlight set.

Also implements browser response indicating the user has unsaved changes (UI browser dependent).

Screen Shot 2023-01-10 at 12 26 41 PM
Screen Shot 2023-01-10 at 12 27 28 PM

# For all changes


src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
       setIsPublishing(false);
     }
   };
+  const closeStepper = () => {
+    open();
+  };
+
+  /* Start beforeunload event */
+  // This section triggers browser response to unsaved items when the stepper modal is open/active
```
adamstankiewicz, line 114: nit: use the standard [JSDoc format](https://jsdoc.app/about-getting-started.html) for comments like this.

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
+  useEffect(() => {
+    const preventUnload = () => {
+      /* eslint-disable no-restricted-globals */
+      /* Mandatory requirements to trigger response by browser, event.preventDefault && event.returnValue
+      A return value is required to trigger the browser unsaved data blocking modal response */
+      event.preventDefault();
```
adamstankiewicz, line 120: Where is `event` coming from here? Does it need to be passed as the function argument to `preventUnload`? E.g., `const preventUnload = (e) => { ... }`. This may make it possible to remove the above `eslint-disable no-restricted-globals`, as you'll no longer be referencing the global `event`.

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
+      event.returnValue = 'Are you sure? Your data will not be saved.';
```
adamstankiewicz, line 121: Bummer that browsers don't let you customize the actual message that shows, but I suppose it makes sense for security reasons.

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
+      /* eslint-enable no-restricted-globals */
+    };
+
+    /* Conditional MUST be set on event listener initialization.
+    Failure to provide conditional will trigger browser event on
+    all elements within ContentHighlightRoutes.jsx (essentially all of highlights) */
+    if (isStepperModalOpen) {
+      window.addEventListener('beforeunload', preventUnload);
```
adamstankiewicz, line 129: nit: use `global` > `window`. we've been bitten elsewhere in the past when React code gets executed at build time, where `window` doesn't exist, but `global` does (i.e., in a Node environment). If this component ever ends up with SSR, it _could_ (depending on whether `window` is called at build time or at runtime) throw an error. I tend to just use `global` to avoid the possibility 🤷‍♂️

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
-                }}
-                variant="primary"
-                onClick={handlePublish}
-                state={isPublishing ? 'pending' : 'default'}
-              />
-            </Stepper.ActionRow>
-          </>
+                <Button
+                  variant="tertiary"
+                  onClick={() => closeStepper()}
```
adamstankiewicz, line 157: nit: `onClick={closeStepper}`

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -46,6 +53,7 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
   const { location } = history;
   const [currentStep, setCurrentStep] = useState(steps[0]);
   const [isPublishing, setIsPublishing] = useState(false);
+  const [isOpen, open, close] = useToggle(false);
```
adamstankiewicz, line 56: nit: `[isCloseAlertOpen, openCloseAlert, closeCloseAlert]` or similar to be a tad more explicit

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
-        footerNode={(
-          <>
-            <Stepper.ActionRow eventKey={STEPPER_STEP_LABELS.CREATE_TITLE}>
-              <HighlightStepperFooterHelpLink />
-              <Stepper.ActionRow.Spacer />
-              {/* TODO: Eventually would need a check to see if the user has made any changes
+    <>
+      <Stepper activeKey={currentStep}>
+        <FullscreenModal
+          id="test"
```
adamstankiewicz, line 142: nit: "test" may not be a super descriptive `id` haha 😜

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
```diff
@@ -95,103 +106,149 @@ const ContentHighlightStepper = ({ enterpriseId }) => {
...
+          title="New highlight"
+          className="bg-light-200"
+          isOpen={isStepperModalOpen}
+          onClose={() => closeStepper()}
```
adamstankiewicz, line 146: `onClose={closeStepper}`

adamstankiewicz (approved):
Only one real comment about using the global `event` versus using the `event` that comes from the event listener itself; the rest are nits!

src/components/ContentHighlights/HighlightStepper/ContentHighlightStepper.jsx
brobro10000: remnants of rage testing

PR #951, opened 2023-01-13
fix: prevent infinite calls on customer-agreement API call when error occurs

adamstankiewicz:
# For all changes

### DIFF
--- a/src/components/subscriptions/data/hooks.js
+++ b/src/components/subscriptions/data/hooks.js
@@ -20,7 +20,7 @@ const subscriptionInitState = {
  * This hook provides all customer agreement and subscription data
  * for the authenticated user and given enterprise customer UUID.
  */
-export const useSubscriptions = ({ enterpriseId, errors, setErrors }) => {
+export const useSubscriptions = ({ enterpriseId, setErrors }) => {
   const [subscriptions, setSubscriptions] = useState({ ...subscriptionInitState });
 
   const [loading, setLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


src/components/subscriptions/data/hooks.js
```diff
@@ -54,16 +54,16 @@ export const useSubscriptions = ({ enterpriseId, errors, setErrors }) => {
...
           [SUBSCRIPTIONS]: NETWORK_ERROR_MESSAGE,
-        });
+        }));
       } finally {
         setLoading(false);
       }
     };
     fetchData();
-  }, [enterpriseId, errors, setErrors]);
+  }, [enterpriseId, setErrors]);
```
adamstankiewicz, line 66: This function relies on both reading `errors` but also mutating `errors` via `setErrors`. Due to this, `errors` needs to be in the dependency array which means the function reference is re-computed whenever the API throws an error. This has a cascading effect in that the `useEffect` below that calls `loadCustomerAgreementData` is constantly fired.

src/components/subscriptions/data/hooks.js
```diff
@@ -54,16 +54,16 @@ export const useSubscriptions = ({ enterpriseId, errors, setErrors }) => {
         setSubscriptions(subscriptionsData);
       } catch (err) {
         logError(err);
-        setErrors({
-          ...errors,
+        setErrors(s => ({
+          ...s,
```
adamstankiewicz, line 58: Use the callback function to retrieve the current state rather than relying on the current `errors` state variable (anti-pattern).

macdiesel (approved):

PR #962, opened 2023-01-25
fix: Input reload fix implemented

brobro10000:
Ticket: https://2u-internal.atlassian.net/browse/ENT-6724
Fixes issue with input as described in the ticket.
Passes all existing test and validation to match previous behavior

# For all changes

### DIFF
--- a/src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
+++ b/src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { useContextSelector } from 'use-context-selector';
 import { Form } from '@edx/paragon';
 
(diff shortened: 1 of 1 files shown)


src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,59 @@ const HighlightStepperTitleInput = () => {
...
+    }
+  }, [highlightTitle, setHighlightTitle, highlightValue]);
   return (
     <Form.Group
       isInvalid={isInvalid}
     >
       <Form.Control
         data-testid="stepper-title-input"
-        value={highlightTitle || ''}
+        value={highlightValue.highlightTitle || ''}
```
adamstankiewicz, line 60: With your refactor above, `highlightTitle` already falls back to `''`, right?

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,59 @@ const HighlightStepperTitleInput = () => {
...
 
+  useEffect(() => {
+    if (highlightValue.initialized) {
+      setHighlightTitle({
+        highlightTitle: highlightValue.highlightTitle,
+        titleStepValidationError: highlightValue.titleStepValidationError,
+      });
+      setTitleLength(highlightValue.highlightTitleLength);
+      setHighlightValue({
+        ...highlightValue,
```
adamstankiewicz, line 49: note: this is relying on the actual state variable in a state setter function. As such, there's no guarantee that `highlightValue` will be the correct/intended reference since the `useEffect` is re-triggered whenever the `highlightValue` itself changes.

Recommended to use the callback function approach to using the previous state to remove it from the dependency array of the `useEffect`:

```jsx
setHighlightValue(prevState = ({
 ...prevState,
  initialized: false,
}));
```

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,59 @@ const HighlightStepperTitleInput = () => {
...
   const [titleLength, setTitleLength] = useState(highlightTitle?.length || 0);
   const [isInvalid, setIsInvalid] = useState(false);
-
-  const handleChange = (e) => {
+  const [highlightValue, setHighlightValue] = useState({
+    initialized: false,
+    highlightTitle: highlightTitle || '',
+    titleStepValidationError: undefined,
+    highlightTitleLength: titleLength || 0,
+  });
```
adamstankiewicz, line 19: [curious/clarification] I'm a bit confused as to why we need to keep track of `titleLength` and `highlightTitle` in 2 separate state variables? In `titleLength` (`highlightValue.highlightTitleLength`) and `highlightTitle` (`highlightValue.highlightTitle`).

Do you mind clarifying this and/or adding some documentation as a comment to explain this logic?

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,59 @@ const HighlightStepperTitleInput = () => {
...
+  const handleChange = ((e) => {
+    e.persist();
```
adamstankiewicz, line 21: Is this `e.persist()` necessary? It appears to work as expected without it.

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
brobro10000: Yep, in order to separate the rendering from the internal component and external stepper, the internal input needed its own unique state as to not re-render the target value state. I will add additional documentation in reference to this.

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
brobro10000: Woops, leftover code from debugging

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
adamstankiewicz: Got it, cool. Yes, this explanation helps me better understand the changes here :)

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
brobro10000: Correct, verified

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,62 @@ const HighlightStepperTitleInput = () => {
...
+   *  Create seperate useState for FormInput as to differenciate between
+   *  the outer stepper context state and inner input useState.
+   * */
+  const [highlightValue, setHighlightValue] = useState({
+    initialized: false,
+    highlightTitle: highlightTitle || '',
+    titleStepValidationError: undefined,
+    highlightTitleLength: titleLength || 0,
+  });
+  const handleChange = ((e) => {
```
adamstankiewicz, line 24: nit: is there an extra set of parentheses around the function?

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
```diff
@@ -11,37 +11,62 @@ const HighlightStepperTitleInput = () => {
...
     if (e.target.value.length > MAX_HIGHLIGHT_TITLE_LENGTH) {
```
adamstankiewicz, line 25: nit: might recommend creating a variable similar to `const value = e.target.value;` for re-use and to avoid a possibility of needing to use `e.persist()` due to setting state and then trying to access the same `e` variable.

src/components/ContentHighlights/HighlightStepper/HighlightStepperTitleInput.jsx
adamstankiewicz: Haha, cool. The point here was that the `|| ''` fallback on this line is largely unnecessary because `highlightValue.highlightTitle` already defines the fallback elsewhere.

adamstankiewicz (approved):
LGTM with 3 additional nits

PR #970, opened 2023-02-13
feat: adding no sso config card

kiram15:
Screenshot 2023-02-13 at 12 29 19 PM

[Jira Ticket 
](https://2u-internal.atlassian.net/browse/ENT-6667)
# For all changes


src/components/settings/SettingsSSOTab/index.jsx
```diff
@@ -26,6 +29,9 @@ const SettingsSSOTab = ({ enterpriseId, setHasSSOConfig }) => {
         validConfigExists = true;
       }
     });
+    if (existingConfigs?.length < 1) {
```
marlonkeating, line 32: I'm a bit curious about the optional chaining construction here.  Do we want this expression to evaluate to false if existingConfigs is undefined or not an array-like object?

kiram15: to be honest this is just logic that is left over and I can add a null check but I feel like we can assume it's an array-like object

marlonkeating (changes requested):
Seeing the 'No SSO' card when I have an incomplete SSO config.

image

src/components/settings/SettingsSSOTab/index.jsx
```diff
@@ -26,6 +29,11 @@ const SettingsSSOTab = ({ enterpriseId, setHasSSOConfig }) => {
         validConfigExists = true;
       }
     });
+    if (existingConfigs?.length < 1 || !existingConfigs) {
```
marlonkeating, line 32: nit: Check for falsiness first
```suggestion
    if (!existingConfigs || existingConfigs?.length < 1) {
```

marlonkeating (changes requested):

marlonkeating (approved):

PR #984, opened 2023-04-20
doc: Lay out workflow changes for adding LMS Selector page

marlonkeating:
This is not for submission, this is just some notes and documentation for how I would generally approach the workflow changes for [ENT-6593](https://2u-internal.atlassian.net/browse/ENT-6593)


marlonkeating:
Work on this front has been completed

PR #990, opened 2023-05-26
fix: bug bash fixes

kiram15:
[Jira Ticket 
](https://2u-internal.atlassian.net/browse/ENT-7174)
# For all changes

Video to show that errors only appear after next button is clicked
https://github.com/openedx/frontend-app-admin-portal/assets/31229189/4381ab4d-0e65-41cb-90e5-1f04c5f809dc

### DIFF
--- a/src/components/settings/SettingsLMSTab/LMSConfigs/SAP/SAPConfigEnablePage.tsx
+++ b/src/components/settings/SettingsLMSTab/LMSConfigs/SAP/SAPConfigEnablePage.tsx
@@ -22,6 +22,16 @@ export const formFieldNames = {
   USER_TYPE: "userType",
 };
 
+export const validationMessages = {
+  displayNameRequired: 'Please enter Display Name',
+  baseUrlRequired: 'Please enter SAP Base URL',
+  companyIdRequired: 'Please enter SAP Company ID',
+  userIdRequired: 'Please enter SAP User ID',
+  keyRequired: 'Please enter OAuth Client ID',
+  secretRequired: 'Please enter OAuth Client Secret',
+  userTypeRequired: 'Please select SAP User Type',
+};
+
 export const validations: FormFieldValidation[] = [
   {
     formFieldId: formFieldNames.SAPSF_BASE_URL,
@@ -38,8 +48,8 @@ export const validations: FormFieldValidation[] = [
   {
     formFieldId: formFieldNames.DISPLAY_NAME,
     validator: (fields) => {
-      const displayName = fields[formFieldNames.DISPLAY_NAME];
-      return !displayName;
+      const error = !fields[formFieldNames.DISPLAY_NAME];
+      return error && validationMessages.displayNameRequired;
     },
   },
   {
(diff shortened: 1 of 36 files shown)


src/components/settings/SettingsLMSTab/LMSConfigs/SAP/SAPConfigEnablePage.tsx
```diff
@@ -22,6 +22,16 @@ export const formFieldNames = {
   USER_TYPE: "userType",
 };
 
+export const validationMessages = {
+  displayNameRequired: 'Please enter Display Name',
+  baseUrlRequired: 'Please enter SAP Base URL',
```
marlonkeating, line 27: Not used

src/components/settings/SettingsLMSTab/LMSConfigs/utils.tsx
```diff
@@ -175,13 +175,18 @@ export async function handleSaveHelper(
...
-        ? INVALID_NAME
-        : false;
+      let validName = true;
+      validName = !(existingConfigNames?.has(fields['displayName']));
+      if (fields.id && !validName) { // if we're editing an existing config
+        if (existingConfigNames.get(fields['displayName']) == fields.id) {
+          validName = true;
+        }
+      }
+      return validName ? '' : INVALID_NAME;
```
marlonkeating, line 189: Should use 'false' for validation that passes.  Technically, a falsy string does work at the moment ([since we're just filtering by truthiness without type checking](https://github.com/openedx/frontend-app-admin-portal/blob/013d73b106b22fd05e3295ac814b500f54517e0b/src/components/forms/data/reducer.ts#L27-L32)), but that might change in the future.
```suggestion
      return validName ? false : INVALID_NAME;
```

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -38,7 +38,7 @@ const SettingsLMSTab = ({
   const [configsExist, setConfigsExist] = useState(false);
   const [showNoConfigCard, setShowNoConfigCard] = useState(true);
   const [configsLoading, setConfigsLoading] = useState(true);
-  const [displayNames, setDisplayNames] = useState([]);
+  const [displayNames, setDisplayNames] = useState(new Map());
```
marlonkeating, line 41: Is the order of the key-value pairs in displayNames important?

src/components/settings/SettingsLMSTab/index.jsx
```diff
@@ -137,9 +137,11 @@ const SettingsLMSTab = ({
 
   useEffect(() => {
     // update list of used display names to prevent duplicates
+    const updatedMap = new Map();
     if (existingConfigsData[0]) {
-      setDisplayNames(existingConfigsData?.map((existingConfig) => existingConfig.displayName));
+      existingConfigsData?.map((existingConfig) => updatedMap.set(existingConfig.displayName, existingConfig.id));
```
marlonkeating, line 142: No reason to use .map if we're not using the functionality of .map to generate the Map.
```suggestion
      existingConfigsData?.forEach((existingConfig) => updatedMap.set(existingConfig.displayName, existingConfig.id));
```

marlonkeating (changes requested):

src/components/settings/SettingsLMSTab/LMSConfigs/utils.tsx
kiram15: I just did this because typescript throws warnings when it expects a string but gets a boolean. Should I just ignore that error?

src/components/settings/SettingsLMSTab/index.jsx
kiram15: No they're not, its just all existing configs and their IDs, it doesn't matter the order just that the info is there

src/components/settings/SettingsLMSTab/LMSConfigs/utils.tsx
marlonkeating: What's the error?  The return value of the Function explicitly accepts boolean.
https://github.com/openedx/frontend-app-admin-portal/blob/0a98549f38f5f1e42bb9c496a3d6b3f2c3b6b09e/src/components/forms/FormContext.tsx#L6-L11

src/components/settings/SettingsLMSTab/index.jsx
marlonkeating: Just curious, what is the benefit you were going for in using Map() over a plain {}?

kiram15: My idea was that it's nice that instead of having to loop through every pair to check the associated key which is O(n), it would be nicer to check the value using a map which is O(1)

marlonkeating (approved):

PR #991, opened 2023-05-26
Learner Credit pulls data from enterprise-access and enterprise-subsidy in the admin portal

emrosarioa:
The Learner Credit page in the Admin Portal was populating with the e-commerce API. This code instead checks enterprise-subsidy to see if there is a subsidy for the enterpriseId, and populates the page with the Subsidy data, if so. Otherwise, the data from the e-commerce API will be used.
Screenshot 2023-05-26 at 3 34 35 PM

# For all changes


src/components/EnterpriseSubsidiesContext/data/hooks.js
```diff
@@ -2,24 +2,53 @@ import { useEffect, useState } from 'react';
...
-          isCurrent: true,
-        });
-        const { results } = camelCaseObject(response.data);
-        setOffers(results);
+        const enterpriseAccessResponse = await EnterpriseAccessApiService.getSubsidyAccessPolicies(enterpriseId);
+        let { results } = camelCaseObject(enterpriseAccessResponse.data);
 
+        if (results.length === 0) {
+          // If there are no subsidy access policies in enterprise, fall back to the e-commerce API.
+          const response = await EcommerceApiService.fetchEnterpriseOffers({
```
adamstankiewicz, line 25: [curious/suggestion] Are these calls made sequentially? If so, we may want to make the calls in parallel, and only evaluate them once they are both loaded and then prioritize the `enterpriseAccessResponse`, for example: https://tanstack.com/query/v4/docs/react/guides/parallel-queries

src/components/EnterpriseSubsidiesContext/data/hooks.js
```diff
@@ -2,24 +2,53 @@ import { useEffect, useState } from 'react';
...
+            isCurrent: true,
+          });
+          results = camelCaseObject(response.data.results);
+          const offerData = {
+            id: results[0].id,
+            name: results[0].displayName,
+            start: results[0].startDatetime,
+            end: results[0].endDatetime,
+            isCurrent: results[0].isCurrent,
```
adamstankiewicz, line 34: nit: perhaps create a `result = results[0]` variable

src/components/EnterpriseSubsidiesContext/data/hooks.js
```diff
@@ -2,24 +2,53 @@ import { useEffect, useState } from 'react';
...
 
   useEffect(() => {
     const fetchOffers = async () => {
       try {
-        const response = await EcommerceApiService.fetchEnterpriseOffers({
-          isCurrent: true,
-        });
-        const { results } = camelCaseObject(response.data);
-        setOffers(results);
+        const enterpriseAccessResponse = await EnterpriseAccessApiService.getSubsidyAccessPolicies(enterpriseId);
```
adamstankiewicz, line 20: [clarification] I'm curious why we need to check if there are any policies before retrieving the subsidy metadata. Is there a way to get at the subsidy metadata without needing to call `enterprise-access` first? For example, the subsidy service's `subsidies` API might accept a `enterprise_customer_uuid` query parameter to only return subsidies for the requested enterprise, by passing the need to go through policies to get a subsidy uuid?

src/components/EnterpriseSubsidiesContext/data/hooks.js
emrosarioa: Great idea!

src/components/EnterpriseSubsidiesContext/data/hooks.js
emrosarioa: Excellent point, I'll call the `subsidies` API directly!

src/components/EnterpriseSubsidiesContext/data/tests/hooks.test.js
```diff
@@ -57,30 +79,87 @@ describe('useEnterpriseOffers', () => {
...
+
+    expect(SubsidyApiService.getSubsidyByCustomerUUID).toHaveBeenCalledWith(TEST_ENTERPRISE_UUID);
+    expect(result.current).toEqual({
+      offers: mockOffers,
+      isLoading: false,
+      canManageLearnerCredit: true,
+    });
+  });
+
+  it('should set canManageLearnerCredit to false if enterprise does not have exactly 1 offer', async () => {
```
adamstankiewicz, line 119: nit: perhaps update test case name to mention both offer or enterprise subsidy, given either of them would make `canManageLearnerCredit` truthy?

adamstankiewicz (approved):

PR #1004, opened 2023-07-10
fix: do not require change password to change reporting configuration

jajjibhai008:
**Description**

***Problem***: 
When we try to update a configuration report then the password is required even if I did not want to update it.

***Solution***: 
I make the password field conditionally required while updating a report.

# For all changes


bilaltahir21 (changes requested):
I think it would be better to use the existing variable `requiredFields` to check for password, it would keep validation logic in one place. 

Can we add some tests to check password checkbox scenario? It would be helpful in future if anything in related logic changes.

PR #1011, opened 2023-07-27
fix: removing moment for dayjs

kiram15:
Comparison of admin-portal's overall package size before and after the removal of Moment.js to Day.js. See ADR for more information on the decision to migrate and alternatives considered. 
https://2u-internal.atlassian.net/browse/ENT-5423

Screenshot 2023-07-28 at 11 24 51 AM

Screenshot 2023-07-28 at 11 27 27 AM


adamstankiewicz:
> Comparison of admin-portal's overall package size before and after the removal of Moment.js to Day.js Screenshot 2023-07-26 at 5 06 31 PM Screenshot 2023-07-28 at 10 27 41 AM

@kiram15 These screenshots only appear to only be showing the difference in "Stat size", where the more important metric for production is the "Gzipped size". What is the new "Gzipped" size of the vendor bundle?

src/components/Admin/SubscriptionDetails.jsx
```diff
@@ -37,7 +37,7 @@ const SubscriptionDetails = ({ enterpriseSlug }) => {
           <div className="mt-3 d-flex align-items-center">
             <div className="mr-5">
               <span>
-                {moment(subscription.startDate).format('MMMM D, YYYY')} - {moment(subscription.expirationDate).format('MMMM D, YYYY')}
+                {dayjs(subscription.startDate).format('MMMM D, YYYY')} - {dayjs(subscription.expirationDate).format('MMMM D, YYYY')}
```
bradenmacdonald, line 40: I'm curious why you're using a date library for this at all, instead of `react-intl`'s [`<FormattedDate>`](FormattedDate) component, which I believe is already included as a paragon peer dependency? (And uses the native `Intl` API behind the scenes, so has very little code size overhead).

src/components/learner-credit-management/tests/OfferDates.test.jsx
```diff
@@ -16,7 +16,7 @@ describe('<OfferDates />', () => {
 
   describe('with start date only', () => {
     it('date is past', () => {
-      const startDate = moment('2022-01-31');
+      const startDate = dayjs('2022-01-31');
```
bradenmacdonald, line 19: Personally, I am not a fan of any date libraries that use a single data type to represent both calendar dates and timestamps, which are very different things. Python gets this right but many JS libraries do not.

package.json
katrinan029, line 1: I think we can remove the `moment` from the dependencies.

katrinan029 (approved):
LGTM, I just have a comment in the `package.json` file.

PR #1019, opened 2023-08-18
Refactor LearnerCreditAllocationTable 

emrosarioa:
Refactor the `LearnerCreditAllocationTable` to have a single `Enrollment Details` column for both the user email or the course title, and also a single input for filtering the search results. 
This uses the API filtering implementation from here: https://github.com/openedx/edx-enterprise-data/pull/388

Previous table:
Screenshot 2023-08-29 at 4 40 17 PM

 Table after changes:

Screenshot 2023-08-31 at 9 54 22 AM

# For all changes


pwnage101:
I have this recollection from the mocks that the content title was more bolded than the email, but in the screenshot both the title and email have the same typeface.  Was this Intentional?

src/components/learner-credit-management/LearnerCreditAllocationTable.jsx
```diff
@@ -33,35 +39,34 @@ const LearnerCreditAllocationTable = ({
...
+      FilterStatusComponent={FilterStatus}
       columns={[
         {
-          Header: 'Email Address',
-          accessor: 'userEmail',
+          Header: 'Date',
+          accessor: 'enrollmentDate',
           // eslint-disable-next-line react/prop-types, react/no-unstable-nested-components
-          Cell: ({ row }) => <EmailAddressTableCell row={row} enterpriseUUID={enterpriseUUID} />,
+          Cell: ({ row }) => dayjs(row.values.enrollmentDate).format('MMMM DD, YYYY'),
```
adamstankiewicz, line 48: Per the Figma mocks, the date is shown as "Apr 15" and "Apr 2", without a year. I've commented in the [Figma](https://www.figma.com/file/8da9TyEpEzgxQld3DZE67w?node-id=1031:86533&mode=design#532322495) to clarify if UX intends to remove the year from the UI. Let's leave it in for now.

Also note that I believe we are including leading zeros on single digit days currently (e.g., `02`), but the Figma mocks show single digit days without leading zeros (e.g., `2`).

Also, looks like the intent is to include the abbreviated month name (e.g., `Apr`) in this column as well.

image

src/components/learner-credit-management/LearnerCreditAllocationTable.jsx
```diff
@@ -33,35 +39,34 @@ const LearnerCreditAllocationTable = ({
...
+            </>
+          ),
+          disableFilters: false,
+          disableSortBy: true,
         },
         {
-          Header: 'Course Price',
+          Header: 'Amount',
           accessor: 'courseListPrice',
           Cell: ({ row }) => `$${row.values.courseListPrice}`,
```
adamstankiewicz, line 67: Looks like we either will either need to these ESLint rules similar to line 54 above, or define this as a standalone React component (not defined inline).

src/components/learner-credit-management/LearnerCreditAllocationTable.jsx
```diff
@@ -34,35 +40,45 @@ const LearnerCreditAllocationTable = ({
...
+                  // eslint-disable-next-line react/prop-types
+                  destination={`${getConfig().ENTERPRISE_LEARNER_PORTAL_URL}/${enterpriseSlug}/course/${row.original.courseKey}`}
+                  target="_blank"
+                >
+                  {/* eslint-disable-next-line react/prop-types */}
+                  {row.original.courseTitle}
+                </Hyperlink>
+              </div>
+            </>
+          ),
```
adamstankiewicz, line 68: IMO, either we should update the ESLint comment above to ensure we disable the rule for an entire block of code, not just the next line. That way, you wouldn't need to include multiple comments for `eslint-disable-next-line react/prop-types` within the same sub-component.

Alternatively, it may make sense to also abstract this into its own standalone component to avoid these ESLint issues altogether. 

Example:

```js
{
  Cell: DescriptionTableCell,
}
```

adamstankiewicz (approved):
LGTM! Just a nit about disabled ESLint rules.

src/components/learner-credit-management/data/hooks.js
```diff
@@ -63,18 +63,15 @@ const applySortByToOptions = (sortBy, options) => {
 };
 
 const applyFiltersToOptions = (filters, options) => {
-  const userSearchQuery = filters?.find(filter => filter.id === 'userEmail')?.value;
-  const courseTitleSearchQuery = filters?.find(filter => filter.id === 'courseTitle')?.value;
   const courseProductLineSearchQuery = filters?.find(filter => filter.id === 'courseProductLine')?.value;
```
adamstankiewicz, line 66: Follow-up from Slack conversation:

> [clarification] in the current state of the (LCM) world, is the courseProductLine filter being used?
> `const courseProductLineSearchQuery = filters?.find(filter => filter.id === 'courseProductLine')?.value;`

[Emily] probably not.

[Adam] even if not used in the current state, since the "Product" column you're keeping in for now isn't exposed in the filters (i.e., disableFilters: true), Markhors next iteration probably will be using it again lolol. So maybe we just leave it in for now with that assumption it will be used again in the near future.

src/components/learner-credit-management/data/hooks.js
```diff
@@ -63,18 +63,15 @@ const applySortByToOptions = (sortBy, options) => {
...
-  if (userSearchQuery) {
-    Object.assign(options, { search: userSearchQuery });
-  }
-  if (courseTitleSearchQuery) {
-    Object.assign(options, { searchCourse: courseTitleSearchQuery });
-  }
+  const searchQuery = filters?.find(filter => filter.id.toLowerCase() === 'enrollment details')?.value;
```
adamstankiewicz, line 67: [curious/clarification] Do you happen to know where this filter id of "enrollment details" (with the space) is coming from? My understanding was that the `filter.id` was based on the `accessor` for the column (e.g., like `courseProductLine` above). But the accessor for the "Enrollment details" column is an object (first time seeing this).

adamstankiewicz (approved):
LGTM! Just one non-blocking question.

src/components/learner-credit-management/data/hooks.js
emrosarioa: Yeah I was surprised as well. It appears that if the accessor is a function (which in this case, it is), it defaults the `filter.id` to the `Header` of the column.

PR #1028, opened 2023-09-12
WIP: feat: Conditionally filter the budget enrollments

zamanafzal:
# For all changes

### DIFF
--- a/src/components/EnterpriseApp/EnterpriseAppRoutes.jsx
+++ b/src/components/EnterpriseApp/EnterpriseAppRoutes.jsx
@@ -14,6 +14,7 @@ import { PlotlyAnalyticsPage } from '../PlotlyAnalytics';
 import { ROUTE_NAMES } from './data/constants';
 import BulkEnrollmentResultsDownloadPage from '../BulkEnrollmentResultsDownloadPage';
 import LearnerCreditManagement from '../learner-credit-management';
+import BudgetDetailPage from '../learner-credit-management/BudgetDetailPage';
 import { EnterpriseSubsidiesContext } from '../EnterpriseSubsidiesContext';
 import ContentHighlights from '../ContentHighlights';
 
@@ -105,6 +106,11 @@ const EnterpriseAppRoutes = ({
         />
       )}
 
+        <Route
+        exact
+        path={`${baseUrl}/admin/${ROUTE_NAMES.learnerCredit}/:id`}
+        component={BudgetDetailPage}
+      />
       {enableContentHighlightsPage && (
         <Route
           path={`${baseUrl}/admin/${ROUTE_NAMES.contentHighlights}`}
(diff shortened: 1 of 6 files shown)


src/components/learner-credit-management/BudgetCard-V2.jsx
```diff
@@ -7,58 +7,45 @@ import {
...
+  const navigateToBudgetRedemptions = (id) => {
+    history.push(`/${enterpriseSlug}/admin/${ROUTE_NAMES.learnerCredit}/${id}`);
   };
 
-  const renderActions = (budgetType) => (
+  const renderActions = (id) => (
     <Button
       data-testid="view-budget"
-      onClick={() => navigateToBudgetRedemptions(budgetType)}
+      onClick={() => navigateToBudgetRedemptions(id)}
```
adamstankiewicz, line 42: [suggestion] I would recommend treating this `Button` as a browser hyperlink with the react-router's `Link` component rather than relying on a JavaScript-only `history.push`. Users might want to open the budget details page in a new tab (e.g., right click and "Open in new tab", or `Cmd + click` on the button).

```jsx
<Button
  data-testid="view-budget"
  as={Link}
  to={`/${enterpriseSlug}/admin/${ROUTE_NAMES.learnerCredit}/${id}`}
>
  View budget
</Button>
```

src/components/EnterpriseApp/EnterpriseAppRoutes.jsx
```diff
@@ -105,6 +106,11 @@ const EnterpriseAppRoutes = ({
         />
       )}
 
+        <Route
+        exact
+        path={`${baseUrl}/admin/${ROUTE_NAMES.learnerCredit}/:id`}
```
adamstankiewicz, line 111: nit: perhaps `:budgetId` to be slightly more explicit?

adamstankiewicz (commented):
Thanks for starting to tackle the routing! Left just 1 minor suggestion for now.

src/components/EnterpriseApp/EnterpriseAppRoutes.jsx
```diff
@@ -105,6 +106,11 @@ const EnterpriseAppRoutes = ({
         />
       )}
 
+        <Route
```
adamstankiewicz, line 109: [suggestion] Given this `Route` and the above `Route` both relate to the same Learner Credit Management (LCM) feature, I'm wondering if this component should be defining a single parent route for the overall LCM feature, and then the component passed to the parent route has its owned rendered sub-routes for the LCM-specific routes. That way, `EnterpriseAppRoutes`, doesn't really need any knowledge of the detail route.

Related, as is, this LCM detail page route would be available even when the LCM overview page route (defined above) is not since it isn't relying on the `canManageLearnerCredit` variable. If the overview page route isn't available; neither should the detail page route.

PR #1029, opened 2023-09-13
fix: Bug bash fixes

kiram15:
https://2u-internal.atlassian.net/browse/ENT-7682

# For all changes

### DIFF
--- a/src/components/settings/SettingsApiCredentialsTab/CopyButton.jsx
+++ b/src/components/settings/SettingsApiCredentialsTab/CopyButton.jsx
@@ -12,6 +12,8 @@ const CopyButton = ({ data }) => {
 
   const handleCopyLink = async () => {
     try {
+      const newData = data;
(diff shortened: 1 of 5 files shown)


src/components/settings/SettingsApiCredentialsTab/CopyButton.jsx
```diff
@@ -12,6 +12,8 @@ const CopyButton = ({ data }) => {
 
   const handleCopyLink = async () => {
     try {
+      const newData = data;
+      ['user', 'id'].forEach(prop => delete newData[prop]);
```
marlonkeating, line 16: Do we need to make a newData variable here?  Because you're just making a reference copy, this statement is equivalent to just deleting from the original data variable.

kiram15: We just had some eslint things about making direct changes to a parameter.

marlonkeating (approved):

PR #1030, opened 2023-09-14
fix: remove broken usage of useCache from getSubsidyRequestConfiguration

iloveagent57:
Fixes an issue where updates to course requests settings weren't reflected in the Settings/Access page.  We were previously setting `useCache` to always be true in the call to fetch subsidy request configs.

Broken behavior was possibly related to this change? https://github.com/openedx/frontend-platform/pull/361 

### DIFF
--- a/src/data/services/EnterpriseAccessApiService.js
+++ b/src/data/services/EnterpriseAccessApiService.js
@@ -7,11 +7,9 @@ class EnterpriseAccessApiService {
 
   static apiClient = getAuthenticatedHttpClient;
 
-  static getSubsidyRequestConfiguration({ enterpriseId, clearCacheEntry = false }) {
+  static getSubsidyRequestConfiguration({ enterpriseId }) {
     const url = `${EnterpriseAccessApiService.baseUrl}/customer-configurations/${enterpriseId}/`;
-    return EnterpriseAccessApiService.apiClient({
-      useCache: configuration.USE_API_CACHE,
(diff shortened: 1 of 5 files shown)


src/data/services/EnterpriseAccessApiService.js
```diff
@@ -9,9 +9,7 @@ class EnterpriseAccessApiService {
 
   static getSubsidyRequestConfiguration({ enterpriseId, clearCacheEntry = false }) {
     const url = `${EnterpriseAccessApiService.baseUrl}/customer-configurations/${enterpriseId}/`;
-    return EnterpriseAccessApiService.apiClient({
-      useCache: configuration.USE_API_CACHE,
-    }).get(url, { clearCacheEntry });
+    return EnterpriseAccessApiService.apiClient().get(url, { clearCacheEntry });
```
adamstankiewicz, line 12: [curious] If this API call never uses `useCache` anymore, is the `clearCacheEntry` logic necessary anymore in this function's signature, passed to this `get` function, or in the caller(s) of `getSubsidyRequestConfiguration`?

adamstankiewicz:
> Broken behavior was possibly related to this change? https://github.com/openedx/frontend-platform/pull/361

Hmm, yeah seems likely especially around the mention of "Usages of clearCacheEntry will have to be updated."... probably something that we missed 🙃

adamstankiewicz (approved):
LGTM! :shipit:

PR #1033, opened 2023-09-18
chore: update max allowed highlights per set and max number of highlight sets

macdiesel:
Pairs with: https://github.com/openedx/enterprise-catalog/pull/672


src/components/ContentHighlights/data/constants.js
```diff
@@ -43,10 +43,10 @@ export const TAB_TITLES = {
 export const MAX_HIGHLIGHT_TITLE_LENGTH = 60;
 
 // Max highlight sets per enteprise curation
```
adamstankiewicz, line 45: nit: while we're in here, typo in `enteprise` above ;)

adamstankiewicz (approved):
LGTM. I assume https://github.com/openedx/enterprise-catalog/pull/672 will be merged/released first?

PR #1034, opened 2023-09-19
add support for optional salesforce ids in csv upload

muhammad-ammar:
**JIRA:** https://2u-internal.atlassian.net/browse/ENT-7642

# For all changes

### DIFF
--- a/src/components/InviteLearnersModal/index.jsx
+++ b/src/components/InviteLearnersModal/index.jsx
@@ -10,7 +10,7 @@ import { camelCaseObject } from '@edx/frontend-platform/utils';
 import emailTemplate from './emailTemplate';
 import TextAreaAutoSize from '../TextAreaAutoSize';
 import FileInput from '../FileInput';
-import { returnValidatedEmails, validateEmailAddrTemplateForm } from '../../data/validation/email';
+import { extractSalesforceIds, returnValidatedEmails, validateEmailAddrTemplateForm } from '../../data/validation/email';
 import { normalizeFileUpload } from '../../utils';
 
 class InviteLearnersModal extends React.Component {
@@ -74,6 +74,10 @@ class InviteLearnersModal extends React.Component {
     };
 
     options.user_emails = returnValidatedEmails(formData);
+    const SFIDs = extractSalesforceIds(formData, options.user_emails);
+    if (SFIDs) {
+      options.user_sfids = SFIDs;
+    }
 
     /* eslint-disable no-underscore-dangle */
     return addLicensesForUsers(options, subscriptionUUID)
--- a/src/data/validation/email.js
+++ b/src/data/validation/email.js
@@ -82,6 +82,33 @@ const validateEmailAddresses = (emails) => {
   return result;
 };
 
+// Each row in textarea or csv can contain email plus an optional salesforce id
+// Email and salesforce id will be separated by comma. This function will read
+// each row, split it by comma and then return an object with three properties:
+//      textEmails: All emails extracted from textarea
+//      csvEmails: All emails extracted from CSV
+//      allEmails: Concatenation of `textEmails` and `csvEmails`
+const extractEmails = (formData) => {
+  let textEmails = [];
+  let csvEmails = [];
+  let allEmails = [];
+
+  if (formData[EMAIL_ADDRESS_TEXT_FORM_DATA] && formData[EMAIL_ADDRESS_TEXT_FORM_DATA].length) {
+    textEmails = formData[EMAIL_ADDRESS_TEXT_FORM_DATA].split(/\r\n|\n/).map(item => item.split(',')[0]);
+  }
+  if (formData[EMAIL_ADDRESS_CSV_FORM_DATA] && formData[EMAIL_ADDRESS_CSV_FORM_DATA].length) {
+    csvEmails = formData[EMAIL_ADDRESS_CSV_FORM_DATA].map(item => item.split(',')[0]);
+  }
+
+  allEmails = [...textEmails, ...csvEmails];
+
+  return {
+    textEmails,
+    csvEmails,
+    allEmails,
+  };
+};
+
 const validateEmailAddressesFields = (formData) => {
   // Validate that email address fields contain valid-looking emails.
   // Expects Redux form data
@@ -90,11 +117,11 @@ const validateEmailAddressesFields = (formData) => {
     _error: [],
   };
 
-  const textAreaEmails = formData[EMAIL_ADDRESS_TEXT_FORM_DATA] && formData[EMAIL_ADDRESS_TEXT_FORM_DATA].split(/\r\n|\n/);
-  const csvEmails = formData[EMAIL_ADDRESS_CSV_FORM_DATA];
+  const { csvEmails, textEmails } = extractEmails(formData);
+  const emails = textEmails.length ? textEmails : csvEmails;
   let {
     invalidEmailIndices,
-  } = validateEmailAddresses(textAreaEmails || csvEmails);
+  } = validateEmailAddresses(emails);
 
   // 1 is added to every index to fix off-by-one error in messages shown to the user.
   invalidEmailIndices = invalidEmailIndices.map(i => i + 1);
@@ -105,7 +132,7 @@ const validateEmailAddressesFields = (formData) => {
 ${invalidEmailIndices.length !== 0 ? `and ${lastEmail}` : `${lastEmail}`} \
 is invalid. Please try again.`;
 
-    errorsDict[textAreaEmails ? EMAIL_ADDRESS_TEXT_FORM_DATA : EMAIL_ADDRESS_CSV_FORM_DATA] = message;
+    errorsDict[textEmails.length ? EMAIL_ADDRESS_TEXT_FORM_DATA : EMAIL_ADDRESS_CSV_FORM_DATA] = message;
     errorsDict._error.push(message);
   }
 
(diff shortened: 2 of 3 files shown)


src/components/InviteLearnersModal/index.jsx
```diff
@@ -74,6 +74,10 @@ class InviteLearnersModal extends React.Component {
     };
 
     options.user_emails = returnValidatedEmails(formData);
+    const sfids = extractEmailIds(formData, options.user_emails);
```
zamanafzal, line 77: can we please use camelcase notation here to define the const?

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
...
-  let emails = [];
+  let emails = extractEmails(formData).allEmails;
+  emails = _.union(emails); // Dedup emails
+  return validateEmailAddresses(emails).validEmails;
+};
+
+// Combine all the rows from textarea and CSV and then make a map of email to salesforce id
+const parseData = (formData) => {
+  const rows = [];
+  const all = {};
```
zamanafzal, line 194: Can we please use a more descriptive variable name here?
e.g. `allRecords`

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
...
+  rows.forEach((row) => {
+    const [email, id] = row.split(',').map(item => item.trim());
+    all[email] = id;
+  });
+
+  return all;
+};
+
+// Extract salesforce ids for all validated emails
+const extractEmailIds = (formData, userEmails) => {
```
zamanafzal, line 213: can we change it to `extractSalesForceIds`?

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
...
     throw new SubmissionError(errorsDict);
   }
-  let emails = [];
+  let emails = extractEmails(formData).allEmails;
+  emails = _.union(emails); // Dedup emails
+  return validateEmailAddresses(emails).validEmails;
+};
+
+// Combine all the rows from textarea and CSV and then make a map of email to salesforce id
+const parseData = (formData) => {
```
iloveagent57, line 192: nit: maybe name this `getSalesforceIdsByEmail`

src/data/validation/email.test.js
```diff
@@ -233,4 +234,43 @@ describe('email validation', () => {
...
+      formData[EMAIL_ADDRESS_TEXT_FORM_DATA] = [
+        'abc@example.com,000000000000ABCABC',
+        'asdf@example.com,',
+        'zzz@example.com,000000000000XYZXYZ',
+      ].join('\n');
+      formData[EMAIL_ADDRESS_CSV_FORM_DATA] = [
+        'one@example.com,000000000000YYYYYY',
+        'two@example.com,000000000000ZZZZZZ',
+        'three@example.com,000000000000ABCDDD',
+        'wow@example.com,',
```
iloveagent57, line 250: Nit: consider adding one more test email in here without a trailing comma, like "nocomma@example.com" to make sure the splitting/parsing still works as expected.

iloveagent57 (approved):
Seems reasonable, couple of small suggestions!

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
...
+const extractSalesforceIds = (formData, userEmails) => {
+  const parsedData = parseData(formData);
+  const ids = [];
+
+  userEmails.forEach((email) => {
+    ids.push(parsedData[email]);
+  });
+
+  const allFalse = ids.every(item => !item);
+  return allFalse ? undefined : ids;
```
adamstankiewicz, line 222: nit: perhaps add a comment and/or rename this variable to explain this logic a bit explicitly? It's not clear to me what `allFalse` really means in this context.

src/components/InviteLearnersModal/index.jsx
```diff
@@ -74,6 +74,10 @@ class InviteLearnersModal extends React.Component {
     };
 
     options.user_emails = returnValidatedEmails(formData);
+    const SFIDs = extractSalesforceIds(formData, options.user_emails);
```
adamstankiewicz, line 77: [curious] Do you think we should have test coverage on this new, added logic to the `InviteLearnersModal` component? I.e., verify that we are indeed passing along the correct `options.user_sfids`, when applicable.

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
...
+
   if (formData[EMAIL_ADDRESS_CSV_FORM_DATA] && formData[EMAIL_ADDRESS_CSV_FORM_DATA].length) {
-    emails.push(...formData[EMAIL_ADDRESS_CSV_FORM_DATA]);
+    rows.push(...formData[EMAIL_ADDRESS_CSV_FORM_DATA]);
   }
-  emails = _.union(emails); // Dedup emails
-  return validateEmailAddresses(emails).validEmails;
+
+  rows.forEach((row) => {
+    const [email, id] = row.split(',').map(item => item.trim());
```
adamstankiewicz, line 205: nit: is it perhaps worth renaming `id` to `salesforceId` or similar, so reading the code has a bit more context added to what type of `id` it is (Salesforce).

src/data/validation/email.js
```diff
@@ -155,19 +183,48 @@ const returnValidatedEmails = (formData) => {
   if (errorsDict._error.length > 0) {
     throw new SubmissionError(errorsDict);
   }
-  let emails = [];
+  let emails = extractEmails(formData).allEmails;
+  emails = _.union(emails); // Dedup emails
```
adamstankiewicz, line 187: nit: would `const emails = _.union(extractEmails(formData.allEmails)) // Extract and dedup emails` be logically equivalent without needing to define and then re-define `emails`?

src/data/validation/email.js
```diff
@@ -82,6 +82,33 @@ const validateEmailAddresses = (emails) => {
...
 
+// Each row in textarea or csv can contain email plus an optional salesforce id
+// Email and salesforce id will be separated by comma. This function will read
+// each row, split it by comma and then return an object with three properties:
+//      textEmails: All emails extracted from textarea
+//      CSVEmails: All emails extracted from CSV
+//      allEmails: Concatenation of `textEmails` and `CSVEmails`
+const extractEmails = (formData) => {
+  let textEmails = [];
+  let CSVEmails = [];
```
adamstankiewicz, line 93: nit: `csvEmails`

src/data/validation/email.js
```diff
@@ -82,6 +82,33 @@ const validateEmailAddresses = (emails) => {
...
+  if (formData[EMAIL_ADDRESS_CSV_FORM_DATA] && formData[EMAIL_ADDRESS_CSV_FORM_DATA].length) {
+    CSVEmails = formData[EMAIL_ADDRESS_CSV_FORM_DATA].map(item => item.split(',')[0]);
+  }
+
+  allEmails = [...textEmails, ...CSVEmails];
+
+  return {
+    textEmails: textEmails.length ? textEmails : undefined,
+    CSVEmails: CSVEmails.length ? CSVEmails : undefined,
+    allEmails,
```
adamstankiewicz, line 108: [suggestion] Given `allEmails` will always return an array, even when empty, I'm wondering if all the properties should follow that convention for consistency, which might simplify this logic a bit.

```js
return {
  textEmails,
  csvEmails,
  allEmails,
}
```

src/data/validation/email.js
```diff
@@ -90,8 +117,9 @@ const validateEmailAddressesFields = (formData) => {
...
   };
 
-  const textAreaEmails = formData[EMAIL_ADDRESS_TEXT_FORM_DATA] && formData[EMAIL_ADDRESS_TEXT_FORM_DATA].split(/\r\n|\n/);
-  const csvEmails = formData[EMAIL_ADDRESS_CSV_FORM_DATA];
+  const extractedEmails = extractEmails(formData);
+  const textAreaEmails = extractedEmails.textEmails;
+  const csvEmails = extractedEmails.CSVEmails;
   let {
     invalidEmailIndices,
   } = validateEmailAddresses(textAreaEmails || csvEmails);
```
adamstankiewicz, line 125: [curious/clarification] Do we only support textarea _OR_ CSV upload? Aka, you can't provide textarea input _AND_ CSV upload together? This logic here would imply that it's one or the other, but I wonder if that's actually what we should be doing as a product requirement. Might be worth clarifying (not blocking/related to this PR, though!).

src/data/validation/email.js
muhammad-ammar: Initially I was returning the exact same object as you suggested but [this line](https://github.com/openedx/frontend-app-admin-portal/pull/1034/files#diff-b9e9f5222df1e9f068d27a79a69afc3bed637207da72c9646d5be5b9b5586c74R125) fails when textEmails is [] and csvEmails have some emails. For example the below code evaluates to `[]` which is not what is expected.

```javascript
[] || ['abc@example.com']
```

This [test](https://github.com/openedx/frontend-app-admin-portal/blob/ed98264e122e47f80edaf6186a1e80c81892a1b7/src/data/validation/email.test.js#L74C1-L74C1) fails if we just return empty array.

src/data/validation/email.js
muhammad-ammar: In existing logic, On the UI we say that either input emails in textarea or upload csv but in reality we pick all the emails from both textarea and csv and then send them to backend for license assignment.

src/data/validation/email.js
muhammad-ammar: this check ensure that if there are no salesforce ids present then do not send an empty array to backend.

src/data/validation/email.js
adamstankiewicz: Thanks for clarifying!

adamstankiewicz (approved):

PR #1046, opened 2023-10-02
feat: label and order each budget card on lCM page

jajjibhai008:
**Description**

1.  Label and order budgets

   - How to label each budget:
       - Budgets that have a start date in the future are labeled upcoming.
       - Budgets that are currently active have an Active label (within the date range and active
       - Budgets that have been marked inactive and/or are expired have an Expired label
   - What will be the order of these budget representations?
       - All active budgets first; followed by all upcoming budgets; followed by all expired budgets 
       - Then within each status, order by showing budget(s) expiring soonest (or most recently) first; 
       - Then if there are multiple budgets with the same end date, show them in alphabetical order
2. Update term date range to show only key date info for easier comprehension at a glance: 
- Active: “Expires [date]” 
- Expired: “Expired [date]”
- Upcoming: “Starts [date]” 

JIRA -> [ENT-7661](https://2u-internal.atlassian.net/browse/ENT-7661), [ENT-7662](https://2u-internal.atlassian.net/browse/ENT-7662)

Updated page
image

# For all changes

### DIFF
--- a/src/components/learner-credit-management/SubBudgetCard.jsx
+++ b/src/components/learner-credit-management/SubBudgetCard.jsx
@@ -6,6 +6,8 @@ import {
   Button,
   Row,
   Col,
+  Badge,
+  Stack,
 } from '@edx/paragon';
 
 import { BUDGET_STATUSES, ROUTE_NAMES } from '../EnterpriseApp/data/constants';
@@ -21,9 +23,8 @@ const SubBudgetCard = ({
   enterpriseSlug,
   isLoading,
 }) => {
-  const formattedStartDate = dayjs(start).format('MMMM D, YYYY');
-  const formattedExpirationDate = dayjs(end).format('MMMM D, YYYY');
-  const budgetStatus = getBudgetStatus(start, end);
+  const budgetLabel = getBudgetStatus(start, end);
+  const formattedDate = dayjs(budgetLabel?.date).format('MMMM D, YYYY');
 
   const renderActions = (budgetId) => (
     <Button
@@ -37,11 +38,12 @@ const SubBudgetCard = ({
 
   const renderCardHeader = (budgetType, budgetId) => {
     const subtitle = (
-      <div className="d-flex flex-wrap align-items-center">
+      <Stack direction="horizontal" gap={2.5}>
+        <Badge variant={budgetLabel.badgeVariant}>{budgetLabel.status}</Badge>
         <span data-testid="offer-date">
-          {formattedStartDate} - {formattedExpirationDate}
+          {budgetLabel.term} {formattedDate}
         </span>
-      </div>
+      </Stack>
     );
 
     return (
@@ -50,10 +52,10 @@ const SubBudgetCard = ({
         subtitle={subtitle}
         className="mb-3"
         actions={
-                    budgetStatus !== BUDGET_STATUSES.upcoming
-                      ? renderActions(budgetId)
-                      : undefined
-                }
+          budgetLabel.status !== BUDGET_STATUSES.scheduled
+            ? renderActions(budgetId)
+            : undefined
+        }
       />
     );
   };
@@ -83,7 +85,7 @@ const SubBudgetCard = ({
     >
       <Card.Body>
         {renderCardHeader(displayName || 'Overview', id)}
-        {budgetStatus !== BUDGET_STATUSES.upcoming && renderCardSection(available, spent)}
+        {budgetLabel.status !== BUDGET_STATUSES.scheduled && renderCardSection(available, spent)}
       </Card.Body>
     </Card>
   );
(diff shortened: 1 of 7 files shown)


src/components/learner-credit-management/data/utils.js
```diff
@@ -134,3 +149,28 @@ export const formatPrice = (price) => {
   });
   return USDollar.format(Math.abs(price));
 };
+
+export const orderOffers = (offers) => {
```
adamstankiewicz, line 153: It may be helpful to add a JSdoc comment above this function to describe its intent/purpose in a sentence or two. Otherwise, I'm assuming the status order below indicates active then upcoming then expired, but may be worth describing in words too.

src/components/learner-credit-management/data/utils.js
```diff
@@ -119,12 +119,27 @@ export const getBudgetStatus = (startDateStr, endDateStr) => {
   const endDate = new Date(endDateStr);
 
   if (currentDate < startDate) {
-    return BUDGET_STATUSES.upcoming;
+    return {
+      status: BUDGET_STATUSES.upcoming,
+      variant: 'info',
```
adamstankiewicz, line 124: nit: perhaps `badgeVariant` just to be explicit about what `variant` its referring to.

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -38,8 +38,16 @@ const SubBudgetCard = ({
   const renderCardHeader = (budgetType, budgetId) => {
     const subtitle = (
       <div className="d-flex flex-wrap align-items-center">
+        <span>
+          <Badge
+            className="p-2 mr-2"
```
adamstankiewicz, line 43: [curious] Is the `p-2` class name needed? It looks like the mocks are using the normal size of the `Badge` without any additional padding.

image

image

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -38,8 +38,16 @@ const SubBudgetCard = ({
   const renderCardHeader = (budgetType, budgetId) => {
     const subtitle = (
       <div className="d-flex flex-wrap align-items-center">
```
adamstankiewicz, line 40: I believe this `subtitle` could be simplified by using `Stack`, e.g.:

```jsx
<Stack direction="horizontal" gap={2.5}>
  <Badge variant="primary">Active</Badge>
  <span data-testid="offer-date">
    Expires Dec 30, 2023
  </span>
</Stack>
```

src/components/learner-credit-management/SubBudgetCard.jsx
jajjibhai008: when I use the Badge component without padding it looks like this (a bit weird) on my local machine. What do you think now?
```js
const subtitle = (
      <Stack direction="horizontal" gap={2.5}>
        <Badge variant={budgetLabel.badgeVariant}>{budgetLabel.status}</Badge>
        <span data-testid="offer-date">
          {budgetLabel.term} {formattedDate}
        </span>
      </Stack>
    );
```
image

src/components/learner-credit-management/SubBudgetCard.jsx
jajjibhai008: it makes sense to me. fixed

src/components/learner-credit-management/data/utils.js
jajjibhai008: Actually, I was working on this PR, and that's why this PR was in [WIP]. But now it is fixed.

src/components/learner-credit-management/SubBudgetCard.jsx
adamstankiewicz: @jajjibhai008 Ah, that's likely because you're running the MFE locally with the default Paragon theme, not the theme that will be installed on stage/prod (i.e., `@edx/brand-edx.org`). Refer to the Paragon documentation site and switch between the "Open edX" and "edX" themes to see the difference: https://paragon-openedx.netlify.app/components/badge/

Instead of running `npm start` to run the MFE locally, I would recommend using `npm run start:with-theme` instead to ensure the `@edx/brand-edx.org` is installed for the application without risking modifying any code.

Related, while not explicitly mocked currently... the badge color variant for "Upcoming" is intended to be `secondary` as opposed to `info`, so that it more closely maps to the variants used for subscription plan status badges ([source](https://github.com/openedx/frontend-app-admin-portal/blob/9cc4dad8060c0724d6b9ae87b4fb19e2fba6e910/src/components/subscriptions/data/constants.js#L52-L56)). 

We're also shooting for parity with Subscription Management in terms of the badge label itself. That is, we should update the "Upcoming" badge to say "Scheduled" instead 😄 

Jenn from Enterprise UX will be updating Figma mocks to include budgets that are scheduled (upcoming) and expired as well.

jajjibhai008: @adamstankiewicz Thanks for your explanation.
Here is the updated LCM page screenshot.

screencapture-localhost-1991-test-enterprise-admin-learner-credit-2023-10-03-16_19_09

saleem-latif (approved):

adamstankiewicz (commented):
LGTM! :shipit:

adamstankiewicz (approved):

mahamakifdar19 (approved):

PR #1051, opened 2023-10-06
feat: display AI analytics section on LPR page

mahamakifdar19:
#### Jira Ticket: 
https://2u-internal.atlassian.net/browse/ENT-7363

#### Description: 
Added AI analytics section on LPR page

#### UI Screens:

https://github.com/openedx/frontend-app-admin-portal/assets/113524403/ff3a3850-772b-4793-ad4c-fd4a7bb16923

Screenshot 2023-10-05 at 2 46 32 PM

# For all changes


src/components/AIAnalyticsSummary/data/hooks.js
```diff
@@ -0,0 +1,35 @@
...
+import LmsApiService from '../../../data/services/LmsApiService';
+
+const useAIAnalyticsSummary = (enterpriseId, insights) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [analyticsSummaryData, setAnalyticsSummaryData] = useState(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setIsLoading(true);
```
saleem-latif, line 12: I do not think we need this. Since, it is `true` by default.

src/components/AIAnalyticsSummary/data/hooks.js
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+import { logError } from '@edx/frontend-platform/logging';
+import LmsApiService from '../../../data/services/LmsApiService';
+
+const useAIAnalyticsSummary = (enterpriseId, insights) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [analyticsSummaryData, setAnalyticsSummaryData] = useState(null);
```
saleem-latif, line 7: I think it would be even better to use something like
```
const [error, setError] = useState(null);
```

And set the error value of there is an error while making the API call. And show the error to the user.

src/components/Admin/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,100 @@
...
+AnalyticsDetailCard.propTypes = {
+  onClose: PropTypes.func.isRequired,
+  data: PropTypes.string,
+};
+
+const AIAnalyticsSummary = ({ enterpriseId, insights }) => {
+  const [showSummarizeCard, setShowSummarizeCard] = useState(false);
+  const [showTrackProgressCard, setShowTrackProgressCard] = useState(false);
+
+  const analyticsSummary = useAIAnalyticsSummary(enterpriseId, insights);
```
saleem-latif, line 37: It would be more readable to use object deconstruction syntax here.

```
const { data: analyticsSummary, isLoading, error } = useAIAnalyticsSummary(enterpriseId, insights);
```

src/components/Admin/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,100 @@
...
+  </Card>
+);
+
+AnalyticsDetailCard.propTypes = {
+  onClose: PropTypes.func.isRequired,
+  data: PropTypes.string,
+};
+
+const AIAnalyticsSummary = ({ enterpriseId, insights }) => {
+  const [showSummarizeCard, setShowSummarizeCard] = useState(false);
```
saleem-latif, line 34: A better option here would be  `useToggle` hook (ref:https://paragon-openedx.netlify.app/components/hooks/usetoggle/)

src/components/Admin/index.jsx
```diff
@@ -309,6 +316,13 @@ class Admin extends React.Component {
                   <h2>Overview</h2>
                 </div>
               </div>
+              <div className="row mt-4">
+                <div className="col">
+                  {insightsLoading ? <AIAnalyticsSummarySkeleton /> : (
+                    insights && <AIAnalyticsSummary enterpriseId={enterpriseId} />
```
saleem-latif, line 322: We should an error message if `insights` are not loaded for some reason.

saleem-latif (changes requested):

src/components/Admin/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,100 @@
...
+import useAIAnalyticsSummary from '../AIAnalyticsSummary/data/hooks';
+
+const AnalyticsDetailCard = ({ onClose, data }) => (
+  <Card className="mt-3 mb-4">
+    <Card.Section>
+      <Badge variant="light" className="mb-3 font-weight-semibold">
+        Beta
+      </Badge>
+      <Stack gap={1} direction="horizontal">
+        <p className="card-text text-justify small">{data || 'No insights found..'}</p>
```
muhammad-ammar, line 18: Questions:

1. `'No insights found..'` message will render if we received the data from Analytics service but no data received from LMS? 
2. Are two dots intentional at the end of message?

src/components/Admin/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,100 @@
...
+        <Button variant="link" className="mb-4 ml-3" onClick={onClose}>
+          <span className="small font-weight-bold text-gray-800">Dismiss</span>
+        </Button>
+      </Stack>
+      <label className="x-small" htmlFor="poweredBylabel">Powered by OpenAI</label>
```
muhammad-ammar, line 23: Curious why we are not using `FormattedMessage` here and above for `Beta`?

src/components/Admin/AIAnalyticsSummary.jsx
muhammad-ammar, line 18: Don't we want to show a loading skeleton when we are loading data from LMS?

src/components/Admin/AIAnalyticsSummarySkeleton.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Skeleton, Stack } from '@edx/paragon';
+
+const AIAnalyticsSummarySkeleton = () => (
+  <Stack direction="horizontal" gap={2}>
+    <Skeleton height={40} width={250} />
+    <Skeleton height={40} width={250} />
+  </Stack>
```
jajjibhai008, line 8: instead of the custom skeleton, can we use a built-in skeleton like [this](https://github.com/openedx/frontend-app-admin-portal/blob/6a47c984164d0bf7c2448daaa089ac6908eb7fcd/src/components/learner-credit-management/BudgetCard-V2.jsx#L33) in the original component?

src/components/Admin/AIAnalyticsSummary.jsx
mahamakifdar19: @muhammad-ammar 
1. Yes, correct. Specifically if `learner_progress`/`learner_engagement` is null or empty.
2. Yes, they were intentional.

src/components/Admin/AIAnalyticsSummary.jsx
mahamakifdar19: @muhammad-ammar somehow missed using `FormattedMessage` here. Will do that!

src/components/Admin/AIAnalyticsSummarySkeleton.jsx
mahamakifdar19: @jajjibhai008 as far as I know the built in skeleton is for cards, here we want to show a skeleton for buttons. Please let me know if I have missed something.

src/components/Admin/index.jsx
mahamakifdar19: @saleem-latif as discussed with @muhammad-ammar, we are not showing `insights` section if we have not received data or encountered some error. Please let me know if this needs to change.

src/components/Admin/index.jsx
```diff
@@ -309,6 +316,13 @@ class Admin extends React.Component {
                   <h2>Overview</h2>
                 </div>
               </div>
+              <div className="row mt-4">
+                <div className="col">
```
jajjibhai008, line 320: Can we use the Row and Col components available in [Paragon](https://paragon-openedx.netlify.app/foundations/layout/#row)?

mahamakifdar19: @jajjibhai008  I used Bootstrap's `row` and `col` because the original structure in Admin `index.jsx` used them. To keep things consistent, I continued with Bootstrap's `row` and `col` classes rather than switching to Paragon's `Row` and `Col` components.
Screenshot 2023-10-10 at 12 29 16 PM

src/components/Admin/index.jsx
saleem-latif: @mahamakifdar19 Here is what I think the current behavior is when there is API error, let me know if my understanding is not correct.

User clicks on the button to view analytics, we will show the user loading card and when the call fails we will remove the loading card from the DOM without any indication of what happened.

I think that will confuse the users, if we instead show an error message after removing the loading card that the API call failed and user can try again later, then it would be a better UX.

saleem-latif (approved):

src/components/Admin/index.jsx
mahamakifdar19: @saleem-latif this section is for displaying analytics buttons "Summarize Analytics" and "Track Progress" so if there is an error or no data available, we will restrict showing the entire section.

muhammad-ammar (approved):

jajjibhai008 (approved):

PR #1054, opened 2023-10-11
feat: add assigned table to LCM activity tab

adamstankiewicz:
# Description

Tickets: 
* [ENT-7611](https://2u-internal.atlassian.net/browse/ENT-7611)
* [ENT-7610](https://2u-internal.atlassian.net/browse/ENT-7610) 

Contents:
* Adds "Assigned" table UI under the Learner Credit Management "Activity" page route with API integration.
* Implements API integration with assignments REST API to populate the table data.
* Implements API integration with subsidy access policy REST API to retrieve the metadata of an individual budget, which returns an assignment configuration UUID to be called by the assignments table.
* Implements a `useSubsidyAccessPolicy` hook that utilizes `useQuery` to act as a state manager for the subsidy access policy from the detail API. This is possible by expanding the `staleTime` on queries. This allows us to expose commonly used metadata across subcomponents to avoid prop-drilling and context (e.g., making the `subsidyAccessPolicy` metadata available, including its `catalog_uuid` in the "Catalog" tab).
* Ensures "Catalog" tab and "Assigned" table are only exposed in the UI if the viewed budget is assignable, based on its budget type (i.e., `AssignedLearnerCreditAccessPolicy`).

### With feature flag ENABLED

image

image

### With feature flag DISABLED

image

## Testing instructions

1. Ensure `enterprise.top_down_assignment_real_time_lcm` Waffle-flag is enabled within the LMS for your user.
2. Find a subsidy access policy UUID of an assignable budget.
3. Visit `localhost:1991/:enterpriseSlug/admin/learner-credit/:subsidyAccessPolicyUUID` to load the budget detail page.
4. You should see:
  - Subsidy access policy's display name, if any.
  - Initial "Assigned" table with any allocated/errored assigments.
  - "Spent" table (may be empty if you don't have analytics API data).

# For all changes


src/components/Sidebar/index.jsx
```diff
@@ -70,59 +70,59 @@ const Sidebar = ({
     {
       title: 'Learner Progress Report',
       to: `${baseUrl}/admin/${ROUTE_NAMES.learners}`,
-      icon: <Icon src={TrendingUp} fixedWidth />,
```
adamstankiewicz, line 73: [context] Removing the `fixedWidth` prop resolves a prop type warning since these icons were migrated away from font-awesome.

src/components/learner-credit-management/AssignmentDetailsTableCell.jsx
```diff
@@ -0,0 +1,39 @@
...
+import { connect } from 'react-redux';
+import { Hyperlink } from '@edx/paragon';
+
+import { configuration } from '../../config';
+
+const AssignmentDetailsTableCell = ({ row, enterpriseSlug }) => {
+  const { ENTERPRISE_LEARNER_PORTAL_URL } = configuration;
+  return (
+    <>
+      <div className="font-weight-bold">{row.original.learnerEmail}</div>
```
adamstankiewicz, line 12: [context] We are assuming that `learnerEmail` always exists here for now, though this may change in the future once product requirements around PII scrubbing (i.e., retirement pipeline) is better understood. In the future, if we need to account for missing learner email here, we can rely on the `EmailAddressTableCell` component which has messaging already in place for this case.

src/components/learner-credit-management/SpendTableEnrollmentDetails.jsx
```diff
@@ -12,17 +12,24 @@ const SpendTableEnrollmentDetailsContents = ({
   enterpriseSlug,
 }) => (
   <>
-    <EmailAddressTableCell row={row} />
+    <EmailAddressTableCell
+      tableId="spent"
+      userEmail={row.original.userEmail}
+      enterpriseEnrollmentId={row.original.enterpriseEnrollmentId}
+      fulfillmentIdentifier={row.original.fulfillmentIdentifier}
```
adamstankiewicz, line 19: [context] When the "Spent" table migrates its data source to pull from transactions list API via enterprise-subsidy, the row data will contain a `fulfillmentIdentifier`, not a `enterpriseEnrollmentId` though they have similar intents (i.e., uniquely identify some form of enrollment for the redemption). Proactively handling when the row either has the `enterpriseEnrollmentId` (analytics API) vs. the `fulfillmentIdentifier` (transactions API).

src/components/App/index.jsx
```diff
@@ -55,64 +62,67 @@ const AppWrapper = () => {
...
-            </Switch>
-          )}
-        />
-        <AuthenticatedPageRoute
-          path="/"
-          render={(routerProps) => <EnterpriseIndexPage {...routerProps} />}
-          authenticatedAPIClient={apiClient}
-          redirect={process.env.BASE_URL}
+    <QueryClientProvider client={queryClient}>
+      <ReactQueryDevtools />
```
adamstankiewicz, line 66: [context] Wraps the existing `App` component with `QueryClientProvider` and includes `ReactQueryDevtools`. Everything else is unchanged.

src/components/learner-credit-management/data/hooks/useOfferSummary.test.js
```diff
@@ -0,0 +1,67 @@
+import { renderHook } from '@testing-library/react-hooks/dom';
```
adamstankiewicz, line 1: [context] Contents are unchanged from original tests. Simply moving tests for `useOfferSummary` into its own standalone test file, so we don't have a giant `hooks.test.js` file over time.

src/components/learner-credit-management/search/CatalogSearchResults.jsx
```diff
@@ -134,14 +133,10 @@ BaseCatalogSearchResults.propTypes = {
...
     message: PropTypes.string,
   }),
-
   searchState: PropTypes.shape({
     page: PropTypes.number,
   }).isRequired,
   paginationComponent: PropTypes.func,
-  // eslint-disable-next-line react/no-unused-prop-types
-  row: PropTypes.string,
-  contentType: PropTypes.string.isRequired,
```
adamstankiewicz, line 144: [context] As far as I can tell, these props are unused. Was seeing an erroneous "`contentType` prop is required warning."

src/data/hooks.js
```diff
@@ -45,12 +41,3 @@ export function useTimeout(callback, delay) {
     timeoutIdRef.current = null;
   }, [callback, delay]);
 }
-
-export const useSelectedCourse = () => {
```
adamstankiewicz, line 49: `useSelectedCourse` is unused anywhere in the app...

src/components/learner-credit-management/BudgetDetailPageHeader.jsx
```diff
@@ -0,0 +1,50 @@
...
+  Row, Col, Breadcrumb, Stack,
+} from '@edx/paragon';
+
+import { ROUTE_NAMES } from '../EnterpriseApp/data/constants';
+import { useBudgetId, useSubsidyAccessPolicy } from './data';
+
+const BudgetDetailPageHeader = ({ enterpriseSlug }) => {
+  const { subsidyAccessPolicyId } = useBudgetId();
+  const { data: subsidyAccessPolicy } = useSubsidyAccessPolicy(subsidyAccessPolicyId);
+  const budgetDisplayName = subsidyAccessPolicy?.displayName || 'Overview';
```
katrinan029, line 15: [question] Not sure if this worth exploring to grab a different value other than `display_name` since most of our subsidies have `display_name` as `null`

src/components/learner-credit-management/data/hooks/useBudgetId.js
```diff
@@ -0,0 +1,15 @@
+import { useParams } from 'react-router-dom';
+import { isUUID } from '../utils';
+
+const useBudgetId = () => {
+  const { budgetId } = useParams();
+  const enterpriseOfferId = isUUID(budgetId) ? null : budgetId;
+  const subsidyAccessPolicyId = isUUID(budgetId) ? budgetId : null;
```
katrinan029, line 7: [nit]: I think it would be helpful to add a comment for context around the differences between `enterpriseOfferId` and `subsidyAccessPolicyId`

src/components/learner-credit-management/BudgetDetailActivityTabContents.jsx
```diff
@@ -1,47 +1,70 @@
...
     offerRedemptions,
     fetchOfferRedemptions,
   } = useOfferRedemptions(enterpriseUUID, enterpriseOfferId, subsidyAccessPolicyId);
 
+  const {
+    isLoading: isLoadingContentAssignments,
+    contentAssignments,
+    fetchContentAssignments,
+  } = useBudgetContentAssignments({
+    assignmentConfigurationUUID: subsidyAccessPolicy?.assignmentConfiguration?.uuid,
```
katrinan029, line 37: [question] can you share some insight on `assignmentConfiguration`. How is that configuration set in the BE? Just looking through the staged data, it looks like this value is always `null`.

katrinan029 (approved):

src/components/learner-credit-management/BudgetDetailPageHeader.jsx
adamstankiewicz: While most budgets (i.e., subsidy access policies) have `display_name=null` today, the expectation is that the provisioning tool in the support-tools MFE will be setting the `display_name` field for net-new budgets. For now, we are relying on ensuring `display_name` is manually set for customers via Django Admin and falling back to "Overview" similar to what the budget overview page does today.

src/components/learner-credit-management/data/hooks/useBudgetId.js
adamstankiewicz: I agree, adding an JSDoc for `useBudgetId` would be helpful! Will update in the morning.

src/components/learner-credit-management/BudgetDetailActivityTabContents.jsx
adamstankiewicz: There is a new SubsidyAccessPolicy type for assignable learner credit. When you create a new one ([example](https://enterprise-access-internal.stage.edx.org/admin/subsidy_access_policy/assignedlearnercreditaccesspolicy/add/)), there is an editable field to add/edit an `AssignmentConfiguration`. From there, you can manually create assignments for the `AssignmentConfiguration` [here](https://enterprise-access-internal.stage.edx.org/admin/content_assignments/learnercontentassignment/).

image

src/components/learner-credit-management/data/hooks/useBudgetContentAssignments.js
```diff
@@ -0,0 +1,65 @@
...
+      setContentAssignments(camelCaseObject(assignmentsResponse.data));
+      setIsLoading(false);
+    };
+    getContentAssignments();
+  }, [isEnabled, assignmentConfigurationUUID]);
+
+  const debouncedFetchContentAssigments = useMemo(
+    () => debounce(fetchContentAssignments, 300),
+    [fetchContentAssignments],
+  );
```
iloveagent57, line 56: Just curious, why do we need to debounce this fetch?

iloveagent57 (approved):

src/components/learner-credit-management/data/hooks/useBudgetContentAssignments.js
adamstankiewicz: Without debouncing, the API will be called on every keystroke of the search input unnecessarily. By debouncing, we only make the API call once the user has stopped typing for at least the 300ms threshold before we make the accompanying API call to fetch content assignments.

PR #1061, opened 2023-10-17
feat: add budget aggregates on the budgetDetail page

jajjibhai008:
**Description**
JIRA -> [ENT-7756](https://2u-internal.atlassian.net/browse/ENT-7756)

Locally implemented design screenshot
image

# For all changes

### DIFF
--- a/src/components/learner-credit-management/BudgetAggregates.jsx
+++ b/src/components/learner-credit-management/BudgetAggregates.jsx
@@ -1,0 +1,78 @@
+import { useParams } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import {
+  Card,
+  Stack,
+  Badge,
+  Icon,
+} from '@edx/paragon';
+import { Info } from '@edx/paragon/icons';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { formatDate, getBudgetStatus } from './data';
+import useOffersBudgets from './data/hooks/useOffersBudgets';
+import LearnerCreditAggregateCards from './LearnerCreditAggregateCards';
+
+const BudgetAggregates = ({
+  enterpriseUUID,
+  offers,
+}) => {
+  const { budgetId } = useParams();
+  const {
+    isLoading: isLoadingBudgets,
+    allBudgets,
+  } = useOffersBudgets(enterpriseUUID, offers);
+  const budget = allBudgets?.find(item => item.id === budgetId);
+  const budgetLabel = getBudgetStatus(budget?.start, budget?.end);
+  const renderCardHeader = () => {
+    const subtitle = (
+      <Stack direction="horizontal" gap={2.5}>
+        <Badge variant={budgetLabel?.badgeVarient} className="m-0">{budgetLabel?.status}</Badge>
+        <span data-testid="offer-date">
+          {formatDate(budget?.start)} - {formatDate(budget?.end)}
+        </span>
+      </Stack>
+    );
+
+    return (
+      <Card.Header
+        title={budget?.displayName || 'Overview'}
+        subtitle={subtitle}
+      />
+    );
+  };
+  const infoMessage = `Data last updated on ${formatDate(budget?.start)}. This data reflects the current active learner credit
+  credit only and does not include any spend by your organization (codes, manual enrollments,
+  past learner credit plans)`;
+
+  return (
+    <Card isLoading={isLoadingBudgets}>
+      {renderCardHeader()}
+      <Card.Section className="pt-0 d-flex flex-column">
+        <Stack direction="horizontal" gap={1} className="d-flex align-items-start my-3">
+          <Icon src={Info} className="mr-2 pt-1" />
+          <FormattedMessage id="adminPortal.detailPage.budgetInfo" defaultMessage={infoMessage} />
+        </Stack>
+        <div className="d-flex flex-wrap mx-n3">
+          <LearnerCreditAggregateCards
+            totalFunds={budget?.totalFunds}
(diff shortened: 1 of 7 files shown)


src/components/learner-credit-management/BudgetAggregates.jsx
```diff
@@ -0,0 +1,76 @@
...
+      />
+    );
+  };
+  return (
+    <Card isLoading={isLoadingBudgets}>
+      {renderCardHeader()}
+      <Card.Section className="pt-0 d-flex flex-column">
+        <Stack direction="horizontal" gap={1} className="d-flex align-items-start my-3">
+          <Icon src={Info} className="mr-2 pt-1" />
+          <span className="w-75"> Data last updated on Jan 03, 2023. This data reflects the current active learner credit
```
mahamakifdar19, line 49: Curious, are we supposed to hardcode this data?

src/components/learner-credit-management/BudgetAggregates.jsx
```diff
@@ -0,0 +1,76 @@
...
+    return (
+      <Card.Header
+        title={budget?.displayName || 'Overview'}
+        subtitle={subtitle}
+      />
+    );
+  };
+  return (
+    <Card isLoading={isLoadingBudgets}>
+      {renderCardHeader()}
```
mahamakifdar19, line 45: NIT: Instead of using a function to render the card header, you can create a separate React component for the customized header

```
const CardHeader = () => {
  const subtitle = (
    <Stack direction="horizontal" gap={2.5}>
      <Badge variant={budgetLabel?.badgeVariant} className="m-0">
        {budgetLabel?.status}
      </Badge>
      <span data-testid="offer-date">
        {formatDate(budget?.start)} - {formatDate(budget?.end)}
      </span>
    </Stack>
  );

  return (
    <Card.Header title={budget?.displayName || 'Overview'} subtitle={subtitle} />
  );
};
```

and then use it like
`<CardHeader />`

src/components/learner-credit-management/data/hooks/useOffersBudgets.js
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    if (!enterpriseOffers) {
+      setIsLoading(false);
+      return;
+    }
+
+    const fetchData = async () => {
+      try {
+        setIsLoading(true);
+        const budgetsPromises = enterpriseOffers.map(async (offer) => {
```
mahamakifdar19, line 21: NIT:

```
const flatBudgets = (
      await Promise.all(
        enterpriseOffers.map(async (offer) => {
          const response = await EnterpriseDataApiService.fetchEnterpriseOfferSummary(enterpriseUUID, offer.id);
          const data = camelCaseObject(response.data);
          return transformBudgetsData(offer, data);
        })
      )
    ).flat(1);

setAllBudgets(flatBudgets);
```
Also, can we rename `flatBudgets` to something more meaningful?

src/components/learner-credit-management/data/utils.js
```diff
@@ -187,3 +187,44 @@ export const orderOffers = (offers) => {
...
+      id: subsidyAccessPolicyUuid || data.offerId,
+      redeemedFunds: parseFloat(amountOfPolicySpent || data.amountOfOfferSpent),
+      remainingFunds: parseFloat(remainingBalance),
+      totalFunds: parseFloat(startingBalance || data.maxDiscount),
+      percentUtilized: parseFloat(percentOfPolicySpent || data.percentOfOfferSpent),
+      displayName: subsidyAccessPolicyDisplayName || offer.name,
+    };
+  };
+
+  if (offerDetails.budgets && offerDetails.budgets.length > 0) {
```
mahamakifdar19, line 221: NIT: `if (offerDetails.budgets?.length > 0)`

src/components/learner-credit-management/BudgetAggregates.jsx
jajjibhai008: Actually, this date value is not available in our API response now. So I am finding a way to make that value available then I will use it here.

src/components/learner-credit-management/BudgetAggregates.jsx
jajjibhai008: Actually, we are not supposed to use that component anywhere else. That's why I am using this as a function here. Also, we are following a similar pattern in other places like rendering values through functions.
 What do you think?

jajjibhai008:
> [clarification] My understanding is that the design work for the budget detail header page is still WIP by Jenn. We were also planning to save the budget detail header related work for last. Just wondering what the intent/plan is for this PR as a result (see [this Slack thread](https://twou.slack.com/archives/C049EGU04EQ/p1697730638584249)). FWIW, it would be helpful to include screenshots of this work in the PR description.
> 
> Joe's stance has largely been "no longer invest in legacy enterprise offers, if possible" so I'm a bit confused why this PR is implementing aggregates for those legacy enterprise offers at this point. I thought the intent moving forward was to only implement the budget detail page aggregates for subsidy access policy budgets, NOT enterprise offers.
> 
> This work feels largely duplicative of https://2u-internal.atlassian.net/browse/ENT-7827, given the final design for the budget aggregates is still WIP by UX (to ensure the design also appropriately takes into account assignments).

@adamstankiewicz thanks for clearing out things. Actually, this work was on our board, that's why I started working on this after getting instruction from Nathan Robertson. I wasn't aware of the new scenario in which we are not showing budget details for legacy offers. And today i saw your [PR](https://github.com/openedx/frontend-app-admin-portal/pull/1054) in which you made a couple of changes to the detail page.
So what I need to do now? Do I need to pause the work because the design is not final and also this work is a duplicate of ENT-7827?

PR #1066, opened 2023-10-20
feat: add course title to assigned table

katrinan029:
# Description
Display the content information associated with an assignment.

# Solution
Use the added field `content_title`  from this [PR](https://github.com/openedx/enterprise-access/pull/296) to display the course title.

Screenshot 2023-10-20 at 2 51 34 PM

# For all changes


src/components/learner-credit-management/tests/AssignmentDetailsTableCell.test.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
adamstankiewicz, line 1: [inform] Currently, the tests for the LCM budget detail page are largely integration-style testing rather than testing each sub-component individually. See the failing test from this PR in the `BudgetDetailPage.test.jsx` file where there is currently an assertion on "View course" that would be replaced with your changes (notably the `renders with assigned table data` test case).

adamstankiewicz (changes requested):

src/components/learner-credit-management/AssignmentDetailsTableCell.jsx
```diff
@@ -39,6 +40,7 @@ AssignmentDetailsTableCell.propTypes = {
       uuid: PropTypes.string.isRequired,
       learnerEmail: PropTypes.string.isRequired,
       contentKey: PropTypes.string.isRequired,
+      contentTitle: PropTypes.string.isRequired,
```
adamstankiewicz, line 43: Given `content_title` is currently an [optional field on the model (`null=True, blank=True`)](https://github.com/openedx/enterprise-access/blob/9ad845bfae5ec84028d58bcddbdcf5d98d04048a/enterprise_access/apps/content_assignments/models.py#L145-L152), the prop type may not need the `isRequired` bit.

We also may want to fallback to "View course" when there happens to be no content title as opposed to rendering an empty hyperlink.

adamstankiewicz (approved):
LGTM, with one suggested change to handle that `content_title` is an optional field on the model and may be worth having a fallback for the chance the assignment might not have one set.

PR #1070, opened 2023-10-26
feat: adding cancel and remind actions to the budget assignment table

alex-sheehan-edx:
Assigned selected:
image

Assigned and not assigned selected:
image

Not assigned selected:
image

https://2u-internal.atlassian.net/browse/ENT-7813

# For all changes

### DIFF
--- a/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Icon,
+  IconButton,
+  OverlayTrigger,
+  Stack,
+  Tooltip,
+} from '@edx/paragon';
+import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
+
+const AssignmentRowActionTableCell = ({ row }) => {
+  const cancelButtonMarginLeft = row.original.state === 'allocated' ? 'ml-2.5' : 'ml-auto';
+  return (
+    <div className="d-flex">
+      {row.original.state === 'allocated' && (
+        <>
+          <OverlayTrigger
+            key="Remind"
+            placement="top"
+            overlay={<Tooltip variant="light" id="tooltip-remind">Remind learner</Tooltip>}
+          >
+            <IconButton
+              className="ml-auto mr-0"
+              src={Mail}
+              iconAs={Icon}
+              alt="Remind learner"
+              // eslint-disable-next-line no-console
+              onClick={() => console.log(`Reminding ${row.original.uuid}`)}
+              data-testid={`remind-assignment-${row.original.uuid}`}
+            />
+          </OverlayTrigger>
+          <Stack direction="horizontal" gap={1} />
+        </>
+      )}
+      <OverlayTrigger
+        key="Cancel"
+        placement="top"
+        overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
+      >
+        <IconButton
+          className={`${cancelButtonMarginLeft} mr-1 text-danger-500`}
+          src={DoNotDisturbOn}
+          iconAs={Icon}
+          alt="Cancel assignment"
+          // eslint-disable-next-line no-console
+          onClick={() => console.log(`Canceling ${row.original.uuid}`)}
+          data-testid={`cancel-assignment-${row.original.uuid}`}
+        />
+      </OverlayTrigger>
+    </div>
+  );
+};
+
+AssignmentRowActionTableCell.propTypes = {
+  row: PropTypes.shape({
+    original: PropTypes.shape({
+      uuid: PropTypes.string.isRequired,
+      state: PropTypes.string.isRequired,
+    }).isRequired,
+  }).isRequired,
+};
+
+export default AssignmentRowActionTableCell;
--- a/src/components/learner-credit-management/BudgetAssignmentsTable.jsx
+++ b/src/components/learner-credit-management/BudgetAssignmentsTable.jsx
@@ -1,11 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { DataTable } from '@edx/paragon';
-
 import TableTextFilter from './TableTextFilter';
 import CustomDataTableEmptyState from './CustomDataTableEmptyState';
 import AssignmentDetailsTableCell from './AssignmentDetailsTableCell';
 import AssignmentStatusTableCell from './AssignmentStatusTableCell';
+import AssignmentRowActionTableCell from './AssignmentRowActionTableCell';
+import AssignmentTableRemindAction from './AssignmentTableRemind';
+import AssignmentTableCancelAction from './AssignmentTableCancel';
 import { DEFAULT_PAGE, PAGE_SIZE, formatPrice } from './data';
 import AssignmentRecentActionTableCell from './AssignmentRecentActionTableCell';
 
@@ -18,6 +20,7 @@ const BudgetAssignmentsTable = ({
 }) => (
   <DataTable
     isSortable
+    isSelectable
     manualSortBy
     isPaginated
     manualPagination
@@ -49,6 +52,13 @@ const BudgetAssignmentsTable = ({
         disableFilters: true,
       },
     ]}
+    additionalColumns={[
+      {
+        Header: '',
+        Cell: AssignmentRowActionTableCell,
+        id: 'action',
+      },
+    ]}
     initialTableOptions={{
       getRowId: row => row?.uuid?.toString(),
     }}
@@ -63,6 +73,10 @@ const BudgetAssignmentsTable = ({
     itemCount={tableData?.count || 0}
     pageCount={tableData?.numPages || 1}
     EmptyTableComponent={CustomDataTableEmptyState}
+    bulkActions={[
+      <AssignmentTableRemindAction />,
+      <AssignmentTableCancelAction />,
+    ]}
   />
 );
 
--- a/src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
+++ b/src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
@@ -705,4 +705,117 @@ describe('<BudgetDetailPage />', () => {
 
     expect(screen.getByText('loading budget activity overview')).toBeInTheDocument();
   });
+
+  it('displays remind row and bulk actions when allocated', async () => {
+    useParams.mockReturnValue({
+      budgetId: mockSubsidyAccessPolicyUUID,
+      activeTabKey: 'activity',
+    });
+    useOfferRedemptions.mockReturnValue({
+      isLoading: false,
+      offerRedemptions: mockEmptyOfferRedemptions,
+      fetchOfferRedemptions: jest.fn(),
+    });
+    useSubsidyAccessPolicy.mockReturnValue({
+      isInitialLoading: false,
+      data: mockAssignableSubsidyAccessPolicy,
+    });
+    useBudgetDetailActivityOverview.mockReturnValue({
+      isLoading: false,
+      data: {
+        contentAssignments: { count: 1 },
+        spentTransactions: { count: 0 },
+      },
+    });
+    useBudgetContentAssignments.mockReturnValue({
+      isLoading: false,
+      contentAssignments: {
+        count: 1,
+        results: [
+          {
+            uuid: 'test-uuid',
+            contentKey: mockCourseKey,
+            contentQuantity: -19900,
+            learnerState: 'active',
+            recentAction: { actionType: 'assigned', timestamp: '2023-10-27' },
+            actions: [],
+            state: 'allocated',
+          },
+        ],
+        numPages: 1,
+        currentPage: 1,
+      },
+      fetchContentAssignments: jest.fn(),
+    });
+    renderWithRouter(<BudgetDetailPageWrapper />);
+    const cancelRowAction = screen.getByTestId('cancel-assignment-test-uuid');
+    const remindRowAction = screen.getByTestId('remind-assignment-test-uuid');
+    expect(cancelRowAction).toBeInTheDocument();
+    expect(remindRowAction).toBeInTheDocument();
+
+    const checkBox = screen.getByTestId('datatable-select-column-checkbox-cell');
+    expect(checkBox).toBeInTheDocument();
+    userEvent.click(checkBox);
+    await waitFor(() => {
+      expect(screen.getByText('Remind (1)')).toBeInTheDocument();
+    });
+    await waitFor(() => {
+      expect(screen.getByText('Cancel (1)')).toBeInTheDocument();
+    });
+  });
+
+  it('hides remind row and bulk actions when allocated', () => {
+    useOfferRedemptions.mockReturnValue({
+      isLoading: false,
+      offerRedemptions: mockEmptyOfferRedemptions,
+      fetchOfferRedemptions: jest.fn(),
+    });
+    useBudgetDetailActivityOverview.mockReturnValue({
+      isLoading: false,
+      data: {
+        contentAssignments: { count: 1 },
+        spentTransactions: { count: 0 },
+      },
+    });
+    useParams.mockReturnValue({
+      budgetId: mockSubsidyAccessPolicyUUID,
+      activeTabKey: 'activity',
+    });
+    useSubsidyAccessPolicy.mockReturnValue({
+      isInitialLoading: false,
+      data: mockAssignableSubsidyAccessPolicy,
+    });
+    useBudgetDetailActivityOverview.mockReturnValue({
+      isLoading: false,
+      data: {
+        contentAssignments: { count: 1 },
+        spentTransactions: { count: 0 },
+      },
+    });
+    useBudgetContentAssignments.mockReturnValue({
+      isLoading: false,
+      contentAssignments: {
+        count: 1,
+        results: [
+          {
+            uuid: 'test-uuid',
+            contentKey: mockCourseKey,
+            contentQuantity: -19900,
+            learnerState: 'accepted',
+            recentAction: { actionType: 'assigned', timestamp: '2023-10-27' },
+            actions: [],
+            state: 'accepted',
+          },
+        ],
+        numPages: 1,
+        currentPage: 1,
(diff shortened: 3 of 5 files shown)


src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -13,46 +13,138 @@ const BudgetAssignmentsTable = ({
...
+      setRowsSelected(false);
+    }
+  };
+
+  const renderBulkActions = (selectedFlatRows) => {
+    if (rowsSelected) {
+      return ([
+        <RemindAction />,
+        <CancelAction />,
+        <ExtraAction />,
```
kiram15, line 51: We don't need any kind of of ExtraAction, Remind and Cancel should be the only options available if rows are selected because refresh is no longer an option.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -13,46 +13,138 @@ const BudgetAssignmentsTable = ({
...
+                    overlay={
+                      <Tooltip variant="light" id={`tooltip-cancel`}>
+                        Cancel assignment
+                      </Tooltip>
+                    }
+                  >
+                    <Icon
+                      className="ml-4.5 mr-2 text-danger-500"
+                      src={DoNotDisturbOn}
+                      onClick={() => console.log(`Cancelling ${Object.values(row.allCells)}`)}
```
kiram15, line 120: Ok this is a nit but just so we don't get confused later, this should be Canceling because of this annoying weirdness  https://www.scribbr.com/us-vs-uk/cancelled-or-canceled/#:~:text=Cancelling%20or%20canceling,-The%20same%20distinction&text=In%20UK%20English%2C%20%E2%80%9Ccancelling%E2%80%9D,is%20the%20more%20common%20spelling.

alex-sheehan-edx: oh lol, sure

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -13,46 +13,138 @@ const BudgetAssignmentsTable = ({
...
-    data={tableData.results}
-    itemCount={tableData.count}
-    pageCount={tableData.numPages}
-    EmptyTableComponent={CustomDataTableEmptyState}
-  />
-);
+}) => {
+  const [rowsSelected, setRowsSelected] = useState(false);
+
+  const CancelAction = ({ selectedFlatRows, ...rest }) => (
```
adamstankiewicz, line 19: Components should not be defined within another component, as described the ESLint rule:

```
Do not define components during render. React will see a new component type on every render and destroy the entire subtree’s DOM nodes and state (https://reactjs.org/docs/reconciliation.html#elements-of-different-types). Instead, move this component definition out of the parent component “BudgetAssignmentsTable” and pass data as props.
```

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
adamstankiewicz: Correct, the "Refresh" CTA is a table action, not a bulk action. It will only show when there are 0 selections, the default behavior of `DataTable`.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -13,46 +13,138 @@ const BudgetAssignmentsTable = ({
...
+          Header: 'Amount',
+          Cell: ({ row }) => `-${formatPrice(row.original.contentQuantity / 100, { maximumFractionDigits: 0 })}`,
+          disableFilters: true,
+        },
+      ]}
+      additionalColumns={[
+        {
+          id: 'action',
+          Header: '',
+          Cell: ({ row }) => (
```
adamstankiewicz, line 86: Should define `Cell` components as a standalone component, perhaps in its own file, similar to the comment about defining components inline:

```
Do not define components during render.
```

We should likely just pass the element name to the `Cell` property to avoid the pattern of defining components during render. Note: it will throw a prop type warning, though (should be fixed through [this issue](https://github.com/openedx/paragon/issues/2737) and [associated PR](https://github.com/openedx/paragon/pull/2749)).

Note x2: The `Amount` column above can define an inline function for its `Cell` attribute because it does not render JSX, so it's not technically a React component like this `action` column would be.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -13,46 +13,138 @@ const BudgetAssignmentsTable = ({
...
+
+  const rowSelected = (rowsSelected) => {
+    if (Object.keys(rowsSelected).length > 1) {
+      setRowsSelected(true);
+    } else {
+      setRowsSelected(false);
+    }
+  };
+
+  const renderBulkActions = (selectedFlatRows) => {
```
adamstankiewicz, line 46: As discussed in Slack, we should rely on `DataTable`'s default behavior with regard to bulk actions. That is, any actions passed into `bulkActions` prop by default will not display until one or more table rows are selected. Until then, anything passed into the `tableActions` prop is displayed (with 0 selections). We shouldn't have to manage selections in custom state outside of `DataTable`.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -0,0 +1,51 @@
...
+      />
+    </OverlayTrigger>
+  </div>
+);
+
+AssignmentRowActionTableCell.propTypes = {
+  row: PropTypes.shape({
+    original: PropTypes.shape({
+      uuid: PropTypes.string.isRequired,
+      learnerEmail: PropTypes.string.isRequired,
```
adamstankiewicz, line 44: nit: `learnerEmail` will not always be available/returned once the assignments stuff is integrated with the GDPR/PII retirement pipeline for when user's request to have their data scrubbed from our records.

Related, if `learnerEmail` isn't actually used within this component (same with `contentKey`), should they be excluded from the prop types for this component?

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -1,23 +1,41 @@
...
+    {`Cancel (${selectedFlatRows.length})`}
+  </Button>
+);
+
+const RemindAction = ({ selectedFlatRows, ...rest }) => (
+  // eslint-disable-next-line no-console
+  <Button onClick={() => console.log('Remind', selectedFlatRows, rest)}>
+    {`Remind (${selectedFlatRows.length})`}
+  </Button>
+);
```
adamstankiewicz, line 29: [suggestion] Might recommend creating these actions in standalone files, one per component, as I imagine these components will grow in size once they integrate with opening modals, etc.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -63,9 +88,30 @@ const BudgetAssignmentsTable = ({
     itemCount={tableData?.count || 0}
     pageCount={tableData?.numPages || 1}
     EmptyTableComponent={CustomDataTableEmptyState}
+    bulkActions={[
+      <RemindAction />,
+      <CancelAction />,
+    ]}
+    onSelectedRowsChanged={() => {}}
```
adamstankiewicz, line 95: Can remove this line now that it's no longer used.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -63,9 +88,30 @@ const BudgetAssignmentsTable = ({
...
   />
 );
 
+CancelAction.propTypes = {
+  selectedFlatRows: PropTypes.arrayOf(PropTypes.shape()),
```
adamstankiewicz, line 100: nit: may be helpful to include the contents of the `PropTypes.shape()`.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -701,4 +701,69 @@ describe('<BudgetDetailPage />', () => {
...
   });
+
+  it('displays bulk actions with remind when row selected and state allocated', () => {
+    useSubsidyAccessPolicy.mockReturnValue({
+      isInitialLoading: false,
+      data: {
+        uuid: 'a52e6548-649f-4576-b73f-c5c2bee25e9c',
+        policyType: 'AssignedLearnerCreditAccessPolicy',
+        isAssignable: true,
+      },
```
adamstankiewicz, line 712: nit: could re-use `mockAssignableSubsidyAccessPolicy` here instead of defining your own assignable subsidy access policy.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -701,4 +701,69 @@ describe('<BudgetDetailPage />', () => {
...
+  });
+
+  it('hides remind row action when state not allocated', () => {
+    useSubsidyAccessPolicy.mockReturnValue({
+      isInitialLoading: false,
+      data: {
+        uuid: 'a52e6548-649f-4576-b73f-c5c2bee25e9c',
+        policyType: 'AssignedLearnerCreditAccessPolicy',
+        isAssignable: true,
+      },
```
adamstankiewicz, line 746: nit: could re-use `mockAssignableSubsidyAccessPolicy` here instead of defining your own assignable subsidy access policy.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -701,4 +701,69 @@ describe('<BudgetDetailPage />', () => {
...
+    });
+    const mockLearnerEmail = 'edx@example.com';
+    const mockCourseKey = 'edX+DemoX';
```
adamstankiewicz, line 715: As ESLint is pointing out above, these variables already defined globally for the file? I see these same variables created on line 60 and 61 on `master`, available to use within any test cases in the file.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -701,4 +701,69 @@ describe('<BudgetDetailPage />', () => {
 
     expect(screen.getByText('loading budget activity overview')).toBeInTheDocument();
   });
+
+  it('displays bulk actions with remind when row selected and state allocated', () => {
```
adamstankiewicz, line 705: [clarification] What is the intent of this test case? Based on the title, I would expect this test to trigger a row to be selected, and assert that the bulk action CTAs to remind / cancel appear only after selection.

As is, though, I believe this test is asserting that the row-level remind/cancel `IconButtons` and associated tooltips are visible in the table row, which is not quite related to bulk actions as the test case's title would indicate.

If you're trying to test the bulk actions themselves in a test case(s), you'd need to trigger a user event to select a row.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -1,23 +1,41 @@
...
+  // eslint-disable-next-line no-console
+  <Button variant="danger" onClick={() => console.log('Cancel', selectedFlatRows, rest)}>
+    {`Cancel (${selectedFlatRows.length})`}
+  </Button>
+);
+
+const RemindAction = ({ selectedFlatRows, ...rest }) => (
+  // eslint-disable-next-line no-console
+  <Button onClick={() => console.log('Remind', selectedFlatRows, rest)}>
+    {`Remind (${selectedFlatRows.length})`}
```
adamstankiewicz, line 27: [clarification] Should we be assuming all selected rows can reminded? For example, if an admin selects a row with the "Status" column depicting "Notifying learner", "Failed: Bad email", etc., I'm figuring we wouldn't want to allow reminders on those rows. 

Related, I see these annotations in the Figma:

image
<br />
<br />

image

Is this something that intends to be addressed as part of this PR or are there plans around the conditional logic for these bulk actions as follow-up work?

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -1,23 +1,41 @@
...
 const FilterStatus = (rest) => <DataTable.FilterStatus showFilteredFields={false} {...rest} />;
 
+const CancelAction = ({ selectedFlatRows, ...rest }) => (
+  // eslint-disable-next-line no-console
+  <Button variant="danger" onClick={() => console.log('Cancel', selectedFlatRows, rest)}>
+    {`Cancel (${selectedFlatRows.length})`}
+  </Button>
+);
+
+const RemindAction = ({ selectedFlatRows, ...rest }) => (
```
adamstankiewicz, line 24: nit: would recommend defining these components as standalone files, not within `BudgetAssignmentsTable` so they have room to grow once they handle any additional logic (e.g., conditional rendering to disable the CTA, opening a confirmation modal, etc.). Will help keep the `BudgetAssignmentsTable` component file small.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
alex-sheehan-edx: oh interesting! This is copied from the other cell component so will we need to account for this there?

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
alex-sheehan-edx: ahh yeah perhaps- it was just so small I didn't think necessary but I agree

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
alex-sheehan-edx: and yeah, relic of copy paste tho and I'll remove

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -0,0 +1,51 @@
...
+        <Icon
+          className="ml-auto mr-0"
+          src={Mail}
+          // eslint-disable-next-line no-console
+          onClick={() => console.log(`Reminding ${row.original.uuid}`)}
+          data-testid={`remind-assignment-${row.original.uuid}`}
+        />
+      </OverlayTrigger>
+    )}
+    <div />
```
adamstankiewicz, line 23: [nit/question] Is this empty `div` just being used for spacing between these two icons? If so, I would recommend wrapping these icons with `<Stack direction="horizontal" gap={2}>` instead of relying on an empty div for styling.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -0,0 +1,51 @@
...
+    <OverlayTrigger
+      key="Cancel"
+      placement="top"
+      overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
+    >
+      <Icon
```
adamstankiewicz, line 29: [clarification] Pretty sure these are intended to be `IconButton` components, not standalone `Icon` components, thus giving them desired hover/focus states and a large click target:

image

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
adamstankiewicz: > This is copied from the other cell component so will we need to account for this there?

Ah, yeah... it was recently made consistent across all `learner-credit-management` components in a recently merged commit. Current state on `master` does not depict required `learnerEmail` anywhere.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
alex-sheehan-edx: > Related, if learnerEmail isn't actually used within this component (same with contentKey), should they be excluded from the prop types for this component?
I would say that this applies here as well? Currently selectedFlatRows isn't used for anything but the length of the array and logging the contents for debugging purposes.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
alex-sheehan-edx: I think this is a rebasing issue 👍

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
adamstankiewicz: It's likely more important once the integration with the API occurs and the `selectedFlatRows` are iterated, etc. Totally cool to defer.

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
alex-sheehan-edx: you're totally right in that we should not allow the reminding of assignments that aren't pending, however current UX is to hide the button, not disable

alex-sheehan-edx: I've added the behavior according to the AC on the ticket

kiram15: The first screenshot you have is from out an out of date mock

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
alex-sheehan-edx: It's about the hiding and showing of the remind action. I've updated the wording, let me know if it makes works for you

kiram15 (approved):

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
adamstankiewicz: Just a heads up, this behavior as implemented may need to be revisited after hearing from Mike Brown from UX about the intended bulk action behavior.

image

The Figma mocks above allude that the "Assigned" table bulk actions should largely have parity with subscription management's bulk actions. However, the behavior as spec'd/implemented (hiding the bulk action when non-actionable rows are selected) is **not** parity with subscription bulk actions.

There are possible usability concerns with hiding the bulk action (e.g., possible lack of discoverability of the bulk action, forcing admins to manually unselect one or more rows for the bulk action to re-appear). These were mitigated in subscriptions bulk actions by keeping the bulk action visible but exclude non-actionable rows. If there are no actionable rows, the bulk action becomes disabled.

Mike/UX is revisiting this interaction now. We may need to update this behavior in code once a final decision is made.

PR #1081, opened 2023-11-02
fix: altering bulk enrollment action behaviour

kiram15:
[Jira ticket with context here 
](https://2u-internal.atlassian.net/browse/ENT-7923)

[Continuation of this PR
](https://github.com/openedx/frontend-app-admin-portal/pull/1070#pullrequestreview-1708357713)
Clip with new bulk action recording (remind is only valid with "waiting for learner" status, and is disabled if the count of the rows selected is 0)
https://github.com/openedx/frontend-app-admin-portal/assets/31229189/d7114dfd-26ab-48fe-b575-dc4ed4289759

# For all changes

### DIFF
--- a/src/components/ContentHighlights/tests/DeleteHighlightSet.test.jsx
+++ b/src/components/ContentHighlights/tests/DeleteHighlightSet.test.jsx
@@ -97,7 +97,7 @@ describe('<DeleteHighlightSet />', () => {
     expect(sendEnterpriseTrackEvent).toHaveBeenCalledTimes(1);
   });
 
-  it('cancelling confirmation modal closes modal', () => {
+  it('canceling confirmation modal closes modal', () => {
     renderWithRouter(
       <DeleteHighlightSetWrapper />,
       { route: initialRouterEntry },
--- a/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
@@ -6,38 +6,35 @@ import {
 import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
 
 const AssignmentRowActionTableCell = ({ row }) => {
-  const learnerStateBool = row.original.learnerState === 'waiting';
-  const cancelButtonMarginLeft = learnerStateBool ? 'ml-2.5' : 'ml-auto';
+  const isLearnerStateWaiting = row.original.learnerState === 'waiting';
   return (
-    <Stack direction="horizontal" gap="1">
-      {learnerStateBool && (
-      <OverlayTrigger
-        key="Remind"
-        placement="top"
-        overlay={<Tooltip variant="light" id="tooltip-remind">Remind learner</Tooltip>}
-      >
-        <IconButton
-          className="ml-auto mr-0"
-          src={Mail}
-          iconAs={Icon}
-          alt="Remind learner"
+    <Stack direction="horizontal" gap="3" className="align-content-end mr-2">
+      {isLearnerStateWaiting && (
+        <OverlayTrigger
+          key="Remind"
+          placement="top"
+          overlay={<Tooltip id={`tooltip-remind-${row.original.uuid}`}>Remind learner</Tooltip>}
+        >
+          <IconButton
+            src={Mail}
+            iconAs={Icon}
+            alt={`Remind learner ${row.original.learnerEmail}`}
               // eslint-disable-next-line no-console
-          onClick={() => console.log(`Reminding ${row.original.uuid}`)}
-          data-testid={`remind-assignment-${row.original.uuid}`}
-        />
-      </OverlayTrigger>
+            onClick={() => console.log(`Reminding ${row.original.uuid}`)}
+            data-testid={`remind-assignment-${row.original.uuid}`}
+          />
+        </OverlayTrigger>
       )}
       <OverlayTrigger
         key="Cancel"
         placement="top"
-        overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
+        overlay={<Tooltip id={`tooltip-cancel-${row.original.uuid}`}>Cancel assignment</Tooltip>}
       >
         <IconButton
-          className={`${cancelButtonMarginLeft} mr-1`}
           variant="danger"
           src={DoNotDisturbOn}
           iconAs={Icon}
-          alt="Cancel assignment"
+          alt={`Cancel assignment ${row.original.learnerEmail}`}
           // eslint-disable-next-line no-console
           onClick={() => console.log(`Canceling ${row.original.uuid}`)}
           data-testid={`cancel-assignment-${row.original.uuid}`}
@@ -50,6 +47,7 @@ const AssignmentRowActionTableCell = ({ row }) => {
 AssignmentRowActionTableCell.propTypes = {
   row: PropTypes.shape({
     original: PropTypes.shape({
+      learnerEmail: PropTypes.string.isRequired,
       learnerState: PropTypes.string.isRequired,
       uuid: PropTypes.string.isRequired,
     }).isRequired,
--- a/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
@@ -7,8 +7,9 @@ import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
 
 const AssignmentRowActionTableCell = ({ row }) => {
   const isLearnerStateWaiting = row.original.learnerState === 'waiting';
+  const emailAltText = row.original.learnerEmail ? `for ${row.original.learnerEmail}` : '';
   return (
-    <Stack direction="horizontal" gap="3" className="align-content-end mr-2">
+    <Stack direction="horizontal" gap="3" className="justify-content-end">
       {isLearnerStateWaiting && (
         <OverlayTrigger
           key="Remind"
@@ -18,8 +19,8 @@ const AssignmentRowActionTableCell = ({ row }) => {
           <IconButton
             src={Mail}
             iconAs={Icon}
-            alt={`Remind learner ${row.original.learnerEmail}`}
-              // eslint-disable-next-line no-console
+            alt={`Send reminder ${emailAltText}`}
+            // eslint-disable-next-line no-console
             onClick={() => console.log(`Reminding ${row.original.uuid}`)}
             data-testid={`remind-assignment-${row.original.uuid}`}
           />
@@ -34,7 +35,7 @@ const AssignmentRowActionTableCell = ({ row }) => {
           variant="danger"
           src={DoNotDisturbOn}
           iconAs={Icon}
-          alt={`Cancel assignment ${row.original.learnerEmail}`}
+          alt={`Cancel assignment ${emailAltText}`}
           // eslint-disable-next-line no-console
           onClick={() => console.log(`Canceling ${row.original.uuid}`)}
           data-testid={`cancel-assignment-${row.original.uuid}`}
@@ -47,8 +48,8 @@ const AssignmentRowActionTableCell = ({ row }) => {
 AssignmentRowActionTableCell.propTypes = {
   row: PropTypes.shape({
     original: PropTypes.shape({
-      learnerEmail: PropTypes.string.isRequired,
-      learnerState: PropTypes.string.isRequired,
+      learnerEmail: PropTypes.string,
+      learnerState: PropTypes.string,
       uuid: PropTypes.string.isRequired,
     }).isRequired,
   }).isRequired,
--- a/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  Icon,
+  IconButton,
+  OverlayTrigger,
+  Stack,
+  Tooltip,
+} from '@edx/paragon';
+import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
+
+const AssignmentRowActionTableCell = ({ row }) => {
+  const cancelButtonMarginLeft = row.original.state === 'allocated' ? 'ml-2.5' : 'ml-auto';
+  return (
+    <div className="d-flex">
+      {row.original.state === 'allocated' && (
+        <>
+          <OverlayTrigger
+            key="Remind"
+            placement="top"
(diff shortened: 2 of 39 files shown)


src/components/ContentHighlights/tests/DeleteHighlightSet.test.jsx
```diff
@@ -97,7 +97,7 @@ describe('<DeleteHighlightSet />', () => {
     expect(sendEnterpriseTrackEvent).toHaveBeenCalledTimes(1);
   });
 
-  it('cancelling confirmation modal closes modal', () => {
```
kiram15, line 100: Just trying to make sure we're using only the US spelling throughout the code

adamstankiewicz: I never knew the double "l" was more used outside of the US! :)

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
-  OverlayTrigger,
-  Stack,
-  Tooltip,
+  Icon, IconButton, OverlayTrigger, Stack, Tooltip,
 } from '@edx/paragon';
 import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
 
 const AssignmentRowActionTableCell = ({ row }) => {
-  const cancelButtonMarginLeft = row.original.state === 'allocated' ? 'ml-2.5' : 'ml-auto';
+  const learnerStateBool = row.original.learnerState === 'waiting';
```
adamstankiewicz, line 9: nit: `isLearnerStateWaiting` might be a more explicit/descriptive variable

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
+  const cancelButtonMarginLeft = learnerStateBool ? 'ml-2.5' : 'ml-auto';
```
adamstankiewicz, line 10: I don't believe this custom styling is necessary given the use of `Stack` surrounding the `OverlayTrigger`(s). Only `Stack` should be responsible for the spacing between these icon buttons with its `gap` prop.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
-            overlay={<Tooltip variant="light" id="tooltip-remind">Remind learner</Tooltip>}
-          >
-            <IconButton
-              className="ml-auto mr-0"
-              src={Mail}
-              iconAs={Icon}
-              alt="Remind learner"
+    <Stack direction="horizontal" gap="1">
+      {learnerStateBool && (
+      <OverlayTrigger
```
adamstankiewicz, line 14: formatting nit: could use extra intent on this `OverlayTrigger`

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
+        key="Remind"
+        placement="top"
+        overlay={<Tooltip variant="light" id="tooltip-remind">Remind learner</Tooltip>}
```
adamstankiewicz, line 17: Based on the Figma spec and the Subscription Management table (which is what we're largely working against for consistency), the `Tooltip`'s variant should be the default such that the background is the dark color, as opposed to white.

image

<br />
<br />

image

adamstankiewicz, line 17: [sanity check, non-blocking] Given the "Remind" icon button's tooltip in each row would the same `id` (i.e., `tooltip-remind`), I'm wondering if there could be a possibility that more than 1 `Tooltip` is rendered at the same time? The concern is that `id`'s in HTML need to be unique (can't have duplicate IDs on the page). Even if there is guaranteed to be at most one `Tooltip` rendered at a time, it may be a good idea to make the `id` for these tooltips unique per row.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
+      >
+        <IconButton
+          className="ml-auto mr-0"
```
adamstankiewicz, line 20: nit: Given the use of `Stack`, we should only rely on it for spacing between these icons. This `className` likely can be removed.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
+          src={Mail}
+          iconAs={Icon}
+          alt="Remind learner"
```
adamstankiewicz, line 23: [suggested improvement, non-blocking] Given multiple rows may be displaying the same actions, I'm wondering if we may want to include the user's email and course title in the `alt` screen reader text for these icons such that any screen reader users would be read something along the lines of, "Send reminder to <learnerEmail> for <courseTitle>" and "Cancel <learnerEmail>'s assignment for <courseTitle>". Bit of an a11y usability improvement to consider.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
+      </OverlayTrigger>
       )}
       <OverlayTrigger
         key="Cancel"
         placement="top"
         overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
       >
         <IconButton
-          className={`${cancelButtonMarginLeft} mr-1 text-danger-500`}
+          className={`${cancelButtonMarginLeft} mr-1`}
```
adamstankiewicz, line 36: This class name likely can be removed given `Stack` is responsible for the spacing between these actions.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -790,12 +790,58 @@ describe('<BudgetDetailPage />', () => {
...
+      isLoading: false,
+      contentAssignments: {
+        count: 1,
+        results: [
+          {
+            uuid: 'test-uuid',
+            contentKey: mockCourseKey,
+            contentQuantity: -19900,
+            learnerState: 'notifying',
+            recentAction: { actionType: 'reminded', timestamp: '2023-10-27' },
```
adamstankiewicz, line 828: nit/clarification: Is it possible for the API to return `learnerState: 'notifying'` while the `recentAction` returns a "reminded" action? I believe the `recentAction` would return "assigned" if the `learnerState` is "notifying".

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -790,12 +790,58 @@ describe('<BudgetDetailPage />', () => {
...
+        currentPage: 1,
+      },
+      fetchContentAssignments: jest.fn(),
+    });
+    renderWithRouter(<BudgetDetailPageWrapper />);
+    const checkBox = screen.getByTestId('datatable-select-column-checkbox-cell');
+    expect(checkBox).toBeInTheDocument();
+    userEvent.click(checkBox);
+    expect(await screen.findByText('Remind (0)')).toBeInTheDocument();
+    expect(await screen.findByText('Remind (0)')).toBeDisabled();
```
adamstankiewicz, line 843: super tiny nit: could opt to only `findByText` once and store the element in a variable for re-use versus calling `findByText` multiple times.

```
const remindBulkAction = await screen.findByText('Remind (0)');
expect(remindBulkAction).toBeInTheDocument();
expect(remindBulkAction).toBeDisabled();
```

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -790,12 +790,58 @@ describe('<BudgetDetailPage />', () => {
     const checkBox = screen.getByTestId('datatable-select-column-checkbox-cell');
     expect(checkBox).toBeInTheDocument();
     userEvent.click(checkBox);
-    await waitFor(() => {
-      expect(screen.getByText('Remind (1)')).toBeInTheDocument();
+    expect(await screen.findByText('Remind (1)')).toBeInTheDocument();
+    expect(await screen.findByText('Cancel (1)')).toBeInTheDocument();
+  });
+
+  it('only allows cancel and not remind in certain states', async () => {
```
adamstankiewicz, line 797: nit: I'm wondering if it'd be a bit more DRY to rely on the existing above test case around displaying actions, but conditionally modify its behavior with arguments passed into the test versus writing a very similar test asserting on similar things. Example:

```jsx
it.each([
  {
    learnerState: 'waiting',
    shouldDisplayRemindAction: true,
  },
  { 
    learnerState: 'notifying',
    shouldDisplayRemindAction: false,
  },
])('displays remind and cancel row and bulk actions when appropriate (%s)', async ({ learnerState, shouldDisplayRemindAction }) => {
  // pass `learnerState` to the mock return value of `useBudgetContentAssignments`.
  // only assert that remind behavior when `shouldDisplayRemindAction=true`; when `false`, would recommend also including an explicit assertion that the remind action is _not_ in the document.
});
```

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -756,7 +756,7 @@ describe('<BudgetDetailPage />', () => {
     useBudgetDetailActivityOverview.mockReturnValue({
       isLoading: false,
       data: {
-        contentAssignments: { count: 1 },
+        contentAssignments: { count: 2 },
```
adamstankiewicz, line 759: [curious] Why the change to `2` here, when the `useBudgetContentAssignments` mock return value below is still only returning 1 result?

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
-            key="Remind"
-            placement="top"
-            overlay={<Tooltip variant="light" id="tooltip-remind">Remind learner</Tooltip>}
-          >
-            <IconButton
-              className="ml-auto mr-0"
-              src={Mail}
-              iconAs={Icon}
-              alt="Remind learner"
+    <Stack direction="horizontal" gap="1">
```
adamstankiewicz, line 12: I believe this should have `className="justify-content-end"` to force the contents of this table cell to be right-aligned in the table, based on the Figma mocks.

image

Note that the subscriptions table (for comparison) appears to not be properly right aligned as expected.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,45 +1,40 @@
...
-        </>
+          onClick={() => console.log(`Reminding ${row.original.uuid}`)}
+          data-testid={`remind-assignment-${row.original.uuid}`}
+        />
+      </OverlayTrigger>
       )}
       <OverlayTrigger
         key="Cancel"
         placement="top"
         overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
```
adamstankiewicz, line 33: Similar comment here about the `Tooltip` variant. I believe we should use the default variant here such that the tooltip's background is dark vs. white, for parity with the subscriptions table:

image

The tooltip background for "Cancel assignment" in the Figma is the danger variant but this isn't a supported variant by Paragon's `Tooltip`; it only supports the default variant (dark) or light.

src/components/ContentHighlights/tests/DeleteHighlightSet.test.jsx
kiram15: See I have to catch myself because apparently I use british spelling for a lot of stuff accidentally? Like grey (which is gray in the US?)

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
kiram15: If it's guaranteed that multiple can never be rendered, I don't see the possibility of confusion? I could tack on the uuid on the end of it if you want to make sure though

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
kiram15: Oops crust from earlier

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
adamstankiewicz: Yeah, I agree it's seemingly very low possibility to have more than 1 `Tooltip` rendered at once (the one instance that might be a factor that I can think of is that perhaps when quickly hovering over one `IconButton` to another in quick succession, due to the animations where one `Tooltip` fades in while the other `Tooltip` is fading out, where both could theoretically be rendered simultaneously.

Might be good to tack on the uuid to the end just in case.

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
kiram15: Good tip!

src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
```diff
@@ -1,62 +1,55 @@
...
-        overlay={<Tooltip variant="light" id="tooltip-cancel">Cancel assignment</Tooltip>}
+        overlay={<Tooltip id={`tooltip-cancel-${row.original.uuid}`}>Cancel assignment</Tooltip>}
       >
         <IconButton
-          className={`${cancelButtonMarginLeft} mr-1 text-danger-500`}
+          variant="danger"
           src={DoNotDisturbOn}
           iconAs={Icon}
-          alt="Cancel assignment"
+          alt={`Cancel assignment ${row.original.learnerEmail}`}
```
adamstankiewicz, line 37: nit: "Cancel assignment **for** <learnerEmail>"

src/components/learner-credit-management/AssignmentTableRemind.jsx
```diff
@@ -4,16 +4,15 @@ import { Button } from '@edx/paragon';
 import { Mail } from '@edx/paragon/icons';
 
 const AssignmentTableRemindAction = ({ selectedFlatRows, ...rest }) => {
-  const hideRemindAction = selectedFlatRows.some(
-    row => row.original.state !== 'allocated',
-  );
-  if (hideRemindAction) {
-    return null;
-  }
+  const selectedEnrollableRows = selectedFlatRows.filter(row => row.original.learnerState === 'waiting').length;
```
adamstankiewicz, line 7: nit: would this make sense to be `selectedRemindableRows` since the variable is determining whether the _Remind_ button is enabled/disabled?

PR #1087, opened 2023-11-07
feat: implementing gated top down allocation table subsidy service api usage

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7830

image

# For all changes


src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
```diff
@@ -29,19 +30,27 @@ const applySortByToOptions = (sortBy, options) => {
   });
 };
 
-const applyFiltersToOptions = (filters, options) => {
+const applyFiltersToOptions = (filters, options, fetchTransactions = null) => {
   const courseProductLineSearchQuery = filters?.find(filter => filter.id === 'courseProductLine')?.value;
-  const searchQuery = filters?.find(filter => filter.id.toLowerCase() === 'enrollment details')?.value;
+  const searchQuery = filters?.find(filter => filter.id.toLowerCase() === 'enrollmentdetails')?.value;
```
alex-sheehan-edx, line 35: small bug with filter ID - toLowerCase doesn't seem to be adding a space to camel cased words

adamstankiewicz: Great catch! Thanks for resolving for both data sources 💯 

[context] Originally, the `id` was based on the `Header` ("Enrollment details") which is why the `toLowerCase()` is there. Since then, the `accessor` attribute was added to the column which changed the `id` to `enrollmentDetails`.

[suggestion/nit, non-blocking] at this point we could probably do without the `.toLowerCase()` and just rely on the actual accessor being used instead (i.e., `enrollmentDetails`), similar to `courseProductLine` above.

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
```diff
@@ -29,19 +30,27 @@ const applySortByToOptions = (sortBy, options) => {
...
+    Object.assign(options, optionsObject);
   }
 };
 
-const useOfferRedemptions = (enterpriseUUID, offerId = null, budgetId = null) => {
+const useOfferRedemptions = (
+  enterpriseUUID,
+  offerId = null,
+  budgetId = null,
+  fetchTransactions = null,
```
adamstankiewicz, line 51: nit: Similar feedback as another comment ("perhaps `shouldFetchSubsidyTransactions = false`")

src/components/learner-credit-management/BudgetDetailRedemptions.jsx
```diff
@@ -3,16 +3,22 @@ import PropTypes from 'prop-types';
...
     offerRedemptions,
     fetchOfferRedemptions,
-  } = useOfferRedemptions(enterpriseUUID, enterpriseOfferId, subsidyAccessPolicyId);
-
+  } = useOfferRedemptions(
+    enterpriseUUID,
+    enterpriseOfferId,
+    subsidyAccessPolicyId,
+    enterpriseFeatures.topDownAssignmentRealTimeLcm,
+    subsidyAccessPolicy.subsidyUuid,
```
adamstankiewicz, line 20: [suggestion, non-blocking] If this component only relies on `subsidyAccessPolicy` just to pass it through to `useOfferRedemptions`, I'm wondering if it'd be a bit cleaner for the `useOfferRedemptions` hook itself to call `useSubsidyAccessPolicy` using the `subsidyAccessPolicyId` it's provided). It may help simplify tests too as this code path may need any mocks for `useSubsidyAccessPolicy`, provided `useOfferRedemptions` is mocked.

src/components/learner-credit-management/data/utils.js
```diff
@@ -95,6 +95,16 @@ export const transformUtilizationTableResults = results => results.map(result =>
   courseKey: result.courseKey,
 }));
 
+export const transformUtilizationTableTransactionResults = results => results.map(result => ({
+  created: result.created,
+  enterpriseEnrollmentId: result.fulfillmentIdentifier,
+  userEmail: result.lmsUserEmail,
+  courseTitle: result.contentTitle,
+  courseListPrice: result.unit === 'usd_cents' ? -1 * (result.quantity / 100) : -1 * results.quantity,
+  uuid: uuidv4(),
```
adamstankiewicz, line 104: [curious/question] Should this be relying on the `uuid` [returned by the `transactions` API](https://enterprise-subsidy.edx.org/api/schema/redoc/#tag/transactions/operation/api_v2_subsidies_admin_transactions_list)?

src/data/services/EnterpriseSubsidyApiService.js
```diff
@@ -6,8 +6,20 @@ import { configuration } from '../../config';
 class SubsidyApiService {
   static baseUrl = `${configuration.ENTERPRISE_SUBSIDY_BASE_URL}/api/v1`;
 
+  static subsidyV2Url = `${configuration.ENTERPRISE_SUBSIDY_BASE_URL}/api/v2/subsidies`;
```
adamstankiewicz, line 9: nit: if we need to have both a `v1` and a `v2` constant, perhaps something like might work. I feel it'd be preferable if we could migrate the existing `v1` call to use the `v2` endpoint so we don't necessarily need to support both `v1` and `v2` together.

```js
class SubsidyApiService {
  static baseUrl = `${configuration.ENTERPRISE_SUBSIDY_BASE_URL}/api/`;
  static baseUrlV1 = `${baseUrl}/v1/`;
  static baseUrlV2 = `${baseUrl}/v2/`;
}
```

adamstankiewicz, line 9: nit: In a similar vein, the other function in this class doesn't define on a constant for the `subsidies` endpoint it uses; I would recommend keeping the endpoint-specific part of the API URL in the specific function itself, or ensure both functions rely on the same constant.

src/data/services/EnterpriseSubsidyApiService.js
```diff
@@ -6,8 +6,20 @@ import { configuration } from '../../config';
...
   static apiClient = getAuthenticatedHttpClient;
 
+  static fetchCustomerTransactions(customerUuid, subsidyUuid, options = {}) {
+    const queryParams = new URLSearchParams({
+      ...snakeCaseObject(options),
+    });
+    const url = `${SubsidyApiService.subsidyV2Url}/${subsidyUuid}/transactions/?${queryParams.toString()}`;
+    return SubsidyApiService.apiClient({
+      useCache: configuration.USE_API_CACHE,
+    }).get(url, { clearCacheEntry: true });
```
adamstankiewicz, line 20: nit: fwiw, neither of these API calls to enterprise-subsidy rely on the cached version of the `httpClient` provided by `@edx/frontend-platform`. Not sure why we'd need to pass `useCache` as a result, since its not relevant when using `getAuthenticatedHttpClient`. Might be worth removing here and below since it's not needed and might cause some confusion around whether the endpoints are cached are not.

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
```diff
@@ -69,14 +78,27 @@ const useOfferRedemptions = (enterpriseUUID, offerId = null, budgetId = null) =>
...
         }
-        const response = await EnterpriseDataApiService.fetchCourseEnrollments(
-          enterpriseUUID,
-          options,
-        );
-        const data = camelCaseObject(response.data);
-        const transformedTableResults = transformUtilizationTableResults(data.results);
+        let data;
+        let transformedTableResults;
+        if (offerId === null && fetchTransactions) { // and do the feature flag check here
```
adamstankiewicz, line 85: [curious] Just wondering on this check against `offerId` being null versus having a non-null `budgetId` (subsidy access policy uuid). If this conditional is to determine whether we're dealing with a subsidy access policy, it might make sense to rely on the `budgetId` variable instead.

adamstankiewicz, line 85: nit: was a bit confused by the comment `// and do the feature flag check here`. The `fetchTransactions` variable currently represents whether the feature flag is enabled or not, correct? Perhaps make the comment more explicit by stating something along the lines of `// Check if the budget is a subsidy access policy and whether the feature flag is enabled`.

src/components/learner-credit-management/data/utils.js
```diff
@@ -95,6 +95,16 @@ export const transformUtilizationTableResults = results => results.map(result =>
   courseKey: result.courseKey,
 }));
 
+export const transformUtilizationTableTransactionResults = results => results.map(result => ({
```
adamstankiewicz, line 98: nit: `transformUtilizationTableSubsidyTransactionResults` just to be a bit more explicit compared to "offer transactions/redemptions" (reduce any naming ambiguity).

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
alex-sheehan-edx: relic of some copy pasta! Happy to rework

src/components/learner-credit-management/data/utils.js
alex-sheehan-edx: yup another copy pasta artifact, good catch

src/data/services/EnterpriseSubsidyApiService.js
alex-sheehan-edx: yaaaaaaaa that's a good callout

src/data/services/EnterpriseSubsidyApiService.js
alex-sheehan-edx: sure!

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
alex-sheehan-edx: ¯\_(ツ)_/¯ I had no particular method to the madness, just what came out first. Makes sense to have it the other way

alex-sheehan-edx: it was truly nothing more than a todo lol

adamstankiewicz (approved):
LGTM!

PR #1091, opened 2023-11-15
feat: display refunds in spent table when backed by transactions API

adamstankiewicz:
# Description

Ticket: [ENT-7644](https://2u-internal.atlassian.net/browse/ENT-7644) | [Figma](https://www.figma.com/file/8da9TyEpEzgxQld3DZE67w/LC2---Top-down-Assignment?type=design&node-id=2305-43752&mode=design&t=33MMwF8xQPMDSRCz-4)

Summary: 
* Handles the conditional display of refunds (reversals) on individual transaction rows per the Figma spec. Only relevant when the table is backed by the `transactions` API via enterprise-subsidy.
  * Note: Handling refunds/reversals for offers (backed by analytics API) is not supported.
* Resolves bug in the "Spent" table display where the enrollment date was always today's date when backed by the `transactions` API.
* Resolves bug in the "Spent" table where the search query parameter was not correctly using `?search_all` when viewing a enterprise offer backed by the analytics API (it was passing `?search` as it was were communicating to `transactions` API instead).
* Resolves bug where `useBudgetDetailActivityOverview` was calling the analytics API for spent transactions whereas the "Spent" table was pulling from the `transactions` (enterprise-subsidy) API. These should always rely on the same data source.
* Ensures the "Amount" column shows non-refunded transaction prices with a `-` to indicate that balance was drawn down (as seen in Figma spec and consistent with "Assigned" table).

image

# For all changes


src/components/learner-credit-management/LearnerCreditAllocationTable.jsx
```diff
@@ -73,10 +73,9 @@ LearnerCreditAllocationTable.propTypes = {
   tableData: PropTypes.shape({
     results: PropTypes.arrayOf(PropTypes.shape({
       userEmail: PropTypes.string,
-      courseTitle: PropTypes.string.isRequired,
+      courseTitle: PropTypes.string,
       courseListPrice: PropTypes.number.isRequired,
       enrollmentDate: PropTypes.string.isRequired,
-      courseProductLine: PropTypes.string.isRequired,
```
adamstankiewicz, line 79: [inform] "Product" column is no longer included in this table.

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
```diff
@@ -83,7 +86,7 @@ const useOfferRedemptions = (
         }
         let data;
         let transformedTableResults;
-        if (budgetId && shouldFetchSubsidyTransactions) {
+        if (budgetId && isTopDownAssignmentEnabled) {
```
pwnage101, line 89: why not just this?
```suggestion
        if (shouldFetchSubsidyTransactions) {
```

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
```diff
@@ -63,14 +63,17 @@ const useOfferRedemptions = (
...
           page: args.pageIndex + 1, // `DataTable` uses zero-indexed array
           pageSize: args.pageSize,
-          ignoreNullCourseListPrice: true,
         };
+        if (!shouldFetchSubsidyTransactions) {
+          options.ignoreNullCourseListPrice = true;
+        }
         if (budgetId !== null) {
-          options.budgetId = budgetId;
+          options[isTopDownAssignmentEnabled ? 'subsidyAccessPolicyUuid' : 'budgetId'] = budgetId;
```
pwnage101, line 76: Should this key off shouldFetchSubsidyTransactions instead?
```suggestion
          options[shouldFetchSubsidyTransactions ? 'subsidyAccessPolicyUuid' : 'budgetId'] = budgetId;
```

pwnage101 (changes requested):
practicing my rejections :). Couple comments, otherwise LGTM!

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
adamstankiewicz: Good catch!

src/components/learner-credit-management/data/hooks/useOfferRedemptions.js
adamstankiewicz: It could; I will update.

pwnage101 (approved):

PR #1100, opened 2023-11-21
chore: Add error message for SSO config step 3 authorization checkbox

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-7984)

# For all changes

### DIFF
--- a/src/components/forms/ValidatedFormCheckbox.tsx
+++ b/src/components/forms/ValidatedFormCheckbox.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import omit from 'lodash/omit';
+import isString from 'lodash/isString';
+
+import { Form } from '@edx/paragon';
+
+import { setFormFieldAction } from './data/actions';
+import { useFormContext } from './FormContext';
+
+type InheritedParagonCheckboxProps = {
+  className?: string;
+  children: string;
+};
+
+export type ValidatedFormCheckboxProps = {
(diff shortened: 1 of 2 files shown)


src/components/forms/ValidatedFormCheckbox.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  return (
+    <>
+      <Form.Label {...formCheckboxProps}>{formCheckboxProps.label}</Form.Label>
+      <Form.Checkbox {...formCheckboxProps} checked={value} onChange={onChange} />
+      {formCheckboxProps.fieldInstructions && (
+        <Form.Text>{formCheckboxProps.fieldInstructions}</Form.Text>
+      )}
+      {showErrors && showError && (
```
kiram15, line 53: Why is there showErrors twice?

kiram15: Ok reading this again, I see that its two different variables, but I think the {showError} below is confusing cause I would assume that's a boolean and not an error message

kiram15 (approved):
Other than one nit about variable names it looks good!

src/components/forms/ValidatedFormCheckbox.tsx
marlonkeating: That's a good point, went ahead and renamed the variable.

PR #1102, opened 2023-11-28
fix: use controlled selection components from DataTable for BudgetAssigmentsTable

adamstankiewicz:
# Description

https://2u-internal.atlassian.net/browse/ENT-8048

Ensures "Select all..." functions as expected, with exception of the Paragon bug related to the row count shown in the "Select all..." label (see [this Paragon bug](https://github.com/openedx/paragon/issues/2872) describing the issue with the "Select all..." label's count). Also, does not update the counts shown in the bulk actions or call a "full table selected" specific API endpoint.

https://github.com/openedx/frontend-app-admin-portal/assets/2828721/80cc5599-34db-4545-b982-674089da67ae

# For all changes

### DIFF
--- a/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentRowActionTableCell.jsx
@@ -1,13 +1,11 @@
-import React, { useContext } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import {
-  Icon, IconButton, OverlayTrigger, Stack, Tooltip, DataTableContext,
+  Icon, IconButton, OverlayTrigger, Stack, Tooltip,
 } from '@edx/paragon';
 import { Mail, DoNotDisturbOn } from '@edx/paragon/icons';
 
 const AssignmentRowActionTableCell = ({ row }) => {
-  const dataTableContext = useContext(DataTableContext);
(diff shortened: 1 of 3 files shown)


brobro10000 (approved):
👍🏽

iloveagent57 (approved):

katrinan029 (approved):
LGTM! I'm wondering if the "Select all 25" text should be updated to simply say "Select all". I think it's a bit misleading to indicate that only 25 items are being selected when in reality it could be more or less (i.e. the recording shows all 60 items selected).

adamstankiewicz:
> LGTM! I'm wondering if the "Select all 25" text should be updated to simply say "Select all". I think it's a bit misleading to indicate that only 25 items are being selected when in reality it could be more or less (i.e. the recording shows all 60 items selected).

@katrinan029 Yeah, the misleading "Select all" label is due to a bug upstream with Paragon and is documented in the GitHub issue linked above (https://github.com/openedx/paragon/issues/2872). Given this repo is blocked on upgrading to the latest versions of Paragon due to the React/React Router version upgrades, we'd have no control over the "Select all" label unless we opt to replicate and modify the default `DataTable.ControlledSelectionStatus` the table relies on (personally, I would avoid this approach in favor of waiting for the bug to be fixed upstream separately, once we can upgrade Paragon in the repo).

PR #1110, opened 2023-12-01
[Budget Detail Header] Display aggregates and budget overview

zwidekalanga:
# For all changes

**Assigned budget with no utilisation:**
Screenshot 2023-12-01 at 08 11 46

**Assigned budget with utilisation:**
Screenshot 2023-12-01 at 08 13 32

**Non assigned budget:**
Screenshot 2023-12-01 at 08 20 39


openedx-webhooks:
Thanks for the pull request, @zwidekalanga! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.


:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/openedx/.github/blob/master/CONTRIBUTING.md)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

    
Once you've signed the CLA, please allow 1 business day for it to be processed.  After this time, you can re-run the     CLA check by adding a comment here that you have signed it.  If the problem persists, you can tag the `@openedx/cla-problems` team in a     comment on your PR for further assistance.

adamstankiewicz:
[inform] The failing `openedx/cla` CI check should be resolved by https://github.com/openedx/axim-engineering/issues/988.

adamstankiewicz:
@zwidekalanga Note, the "Lint Commit Messages" CI check is failing as it expects commit messages to include a semantic commit type prefix. In this case, I would recommend amending your commit to include `feat: Add card banner...`.

src/components/learner-credit-management/BudgetDetailAssignments.jsx
```diff
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import BudgetAssignmentsTable from './BudgetAssignmentsTable';
 import AssignMoreCoursesEmptyStateMinimal from './AssignMoreCoursesEmptyStateMinimal';
 import { useBudgetContentAssignments, useBudgetId, useSubsidyAccessPolicy } from './data';
+import { useLocation } from "react-router-dom";
```
adamstankiewicz, line 8: image

src/components/learner-credit-management/BudgetDetailAssignments.jsx
```diff
@@ -26,20 +29,26 @@ const BudgetDetailAssignments = ({
...
-
+  
   if (!hasContentAssignments && hasSpentTransactions) {
     return (
       <AssignMoreCoursesEmptyStateMinimal />
     );
   }
-
+  
+  useEffect(() => {
```
adamstankiewicz, line 43: As is, this `useEffect` is breaking the Rules of Hooks (important ESLint error).

image

src/components/learner-credit-management/BudgetDetailPageOverviewUtilization.jsx
```diff
@@ -0,0 +1,108 @@
...
+      </Button>
+    );
+  };
+  
+  return (
+    <Collapsible
+      className='mt-4 budget-utilization-container'
+      styling="basic"
+      title={<h6 className="mb-0">Utilization details</h6>}
+      defaultOpen
```
adamstankiewicz, line 54: The `Collapsible` should be not be opened by default.

image

src/components/learner-credit-management/BudgetDetailPageOverviewUtilization.jsx
```diff
@@ -0,0 +1,108 @@
...
+          </Col>
+        </Row>
+      </Stack>
+    </Collapsible>
+  );
+}
+
+BudgetDetailPageOverviewUtilization.propTypes = {
+  budgetInfo: PropTypes.any.isRequired,
+  aggregates: PropTypes.any.isRequired, // Consider using a more specific prop type for policy if possible,
```
adamstankiewicz, line 104: Generally, `.any` isn't recommend as its better to be explicit with prop types.

src/components/learner-credit-management/BudgetDetailAssignments.jsx
```diff
@@ -26,20 +29,26 @@ const BudgetDetailAssignments = ({
...
+    if (locationState?.budgetActivityScrollToKey === 'assigned') {
+      assignedHeadingRef.current?.scrollIntoView({ behavior: 'smooth' });
```
adamstankiewicz, line 45: I believe we may need to `delete` the locationState after we've handled it here. Otherwise, for example, after I scroll to the "Assigned" / "Spent" tables, and then do a full page refresh, the `locationState` is unchanged and incorrectly scrolls me down to the "Assigned" / "Spent" tables (depending on which scroll-to link was last clicked).

src/components/learner-credit-management/styles/index.scss
```diff
@@ -37,3 +42,11 @@
     border: 1px solid $danger;
   }
 }
+
+.budget-overview-card {
+  .budget-utilization-container {
+    .collapsible-body {
+      padding: 0;
```
adamstankiewicz, line 49: Any way to avoid custom class names solely for a `padding: 0`? E.g., is it possible to pass a `className="p-0"` to the appropriate component?

adamstankiewicz (commented):
Here's a first pass at a review! Overall, it looks really good! Just a heads up, there are a fair number of ESLint errors that will have to be taken care of. Would recommend installing/using an ESLint extension in your IDE or ensuring `npm run lint` passes.

src/components/learner-credit-management/styles/index.scss
zwidekalanga: `div` rendered by the Collapsible component isn't accessible, maybe might be accessible if I use advance option, but felt its might be overkill since I'm only trying to align the available budget section and utilisation section.

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -0,0 +1,92 @@
...
+          Contact Support
+        </Button>
+      </div>
+    )
+  }
+  
+  return (
+    <div className="p-4">
+      <h4 className='text-primary-500'>Get people learning using this budget</h4>
+      <Button variant="brand" className="mr-2 mt-3" iconBefore={Add} as={Link} to={generatePath()}>
```
adamstankiewicz, line 61: The `generatePath` function is not given any arguments; the CTA button currently brings the user to the root of the MFE (http://localhost:1991) instead of to the "Catalog" tab.

src/components/learner-credit-management/styles/index.scss
```diff
@@ -1,3 +1,9 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@400;700&display=swap');
+
+.font-family-roboto {
+  font-family: 'Roboto Mono', monospace;
+}
```
adamstankiewicz, line 5: Instead of defining this custom class name and importing the Roboto font, I would recommend relying on the `.text-monospace` to ensure these elements get the monospace font family. Grabbed from the [CSS utilities](https://paragon-openedx.netlify.app/foundations/css-utilities/) page on the Paragon documentation site.


image

src/components/learner-credit-management/styles/index.scss
adamstankiewicz: Ah, you're right that the `.collapsible-body` element isn't accessible using the default `Collapsible` and agree its not worth refactoring to adopt the `Collapsible.Advanced` pattern here.

That said, it looks like you could consider using some negative margins to achieve the same effect on the parent `Stack` rendered as the child of `Collapsible`.

```jsx
return (
    <Collapsible
      className='mt-4'
      styling="basic"
      title={<h6 className="mb-0">Utilization details</h6>}
    >
      <Stack className="mt-2 mx-n2">
      // ...
    </Collapsible>
```

where `mx-n2` indicates negative left/right margin using the same spacer value as the `.collapsible-body` class name, effectively negating the `.collapsible-body` padding without any custom class names.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -181,9 +181,10 @@ describe('<BudgetDetailPage />', () => {
   });
 
   it.each([
-    { displayName: null },
-    { displayName: 'Test Budget Display Name' },
-  ])('renders budget header data (%s)', ({ displayName }) => {
+    { displayName: null, assigned: false },
+    { displayName: 'Test Budget Display Name', assigned: false },
+    { displayName: 'Assigned Budget', assigned: true },
```
adamstankiewicz, line 186: nit: `assigned` -> `isAssigned` to make the name more explicit that we're working with a boolean.

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -0,0 +1,92 @@
...
+  
+  if (!isAssignable) {
+    return (
+      <div className="p-4">
+        <h4 className='text-primary-500'>Get people learning using this budget</h4>
+        <p>
+          Funds from this budget are set auto allocate to registered
+          learners based on settings configured with your support team.
+        </p>
+        <Button variant="outline-dark" className="mr-2 mt-3" as={Hyperlink} destination={destinationUrl} target="_blank">
```
adamstankiewicz, line 51: nit: is the `mr-2` class name necessary? There aren't any elements positioned next to the button so I don't think the right margin is really doing much here.

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -0,0 +1,92 @@
...
+  </span>
+);
+
+BudgetInfo.propTypes = {
+  label: PropTypes.string.isRequired,
+  value: PropTypes.string.isRequired,
+};
+
+const BudgetDetail = ({ available, utilized, limit }) => (
+  <Stack className="border border-light-400 p-4">
```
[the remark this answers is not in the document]
zwidekalanga: I've tried Card out and it adds additional styling like `border-radius` that I'll have to remove.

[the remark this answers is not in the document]
adamstankiewicz: @zwidekalanga OK cool, yeah I missed that one when I was experimenting with it locally during review. In that case, let's leave it as is with the `Stack`!

src/components/learner-credit-management/search/CatalogSearch.jsx
```diff
@@ -19,6 +18,7 @@ const CatalogSearch = () => {
   } = useSubsidyAccessPolicy(subsidyAccessPolicyId);
   const searchFilters = `enterprise_catalog_uuids:${ENABLE_TESTING(subsidyAccessPolicy.catalogUuid)} AND content_type:course`;
   const displayName = subsidyAccessPolicy.displayName ? `${subsidyAccessPolicy.displayName} catalog` : 'Overview';
```
brobro10000, line 20: Maybe a question for @adamstankiewicz , but is this duplicate logic from the `useBudgetDetailPage` function? 
Is there a way we can reduce duplicate logic in this case?

brobro10000 (commented):
Great job, just a bunch of tiny nits, but nothing blocking on my end 👍🏽

src/components/learner-credit-management/search/CatalogSearch.jsx
adamstankiewicz: @brobro10000 There most definitely is some duplicate logic here, and it's duplicated in more places than just these 2 (i.e., given a display name string, fall back to "Overview" if not set). I would recommend splitting out the refactor to DRY up the fallback to `Overview` as a separate chunk of work from this ticket, though.

brobro10000: I can plus 1 on that 🦖

src/components/learner-credit-management/BudgetDetailPageOverviewUtilization.jsx
```diff
@@ -0,0 +1,132 @@
...
+  const renderActivityLink = ({ amount, type }) => {
+    if (amount <= 0) {
+      return null;
+    }
+
+    const linkText = (type === 'assigned') ? 'View assigned activity' : 'View spent activity';
+
+    return (
+      <Button
+        onClick={() => handleTableSkipLink(type)}
```
adamstankiewicz, line 47: It looks like it might be possible to rely on `Link` here as well, passing the location state via the `to` prop, ensuring the button acts more like a regular ol' hyperlink, e.g. 

```jsx
<Button
  as={Link}
  to={{
    pathname: generatePath(routeMatch.path, { budgetId, activeTabKey: 'activity' }),
    state: { budgetActivityScrollToKey: type },
  }}
  size="sm"
  variant="link"
  iconAfter={ArrowDownward}
  className="pl-0 pr-0"
>
  {linkText}
</Button>
```

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  return (
+    <div className="p-4">
+      <h4>Get people learning using this budget</h4>
+      <Button
+        variant="brand"
+        className="mt-3"
+        iconBefore={Add}
+        onClick={() => handleSkipToCatalog('catalog')}
```
adamstankiewicz, line 76: Related, it looks like when I'm already on the "Catalog" tab route and click this CTA button (in this current implement or the suggested implementation with `Link`), it doesn't seem to scroll down to the "Catalog" tab contents. Do you also see this behavior?

src/components/learner-credit-management/data/hooks/useBudgetDetailPage.js
```diff
@@ -0,0 +1,76 @@
...
+    badgeVariant: 'warning',
+    status: BUDGET_STATUSES.scheduled,
+    term: 'Ends',
+    date: '',
+    isAssignable: policy.isAssignable || false,
+    budgetDisplayName: policy?.displayName || 'Overview',
+  };
+
+  if (!policy) {
+    return defaultData;
```
adamstankiewicz, line 66: [clarification] Does it make sense to return any data from this hook if both `subsidyAccessPolicy` and `enterpriseOffer` are not defined? Doing so, could likely be misleading in that it returns an object that _seems_ like its valid policy metadata even though no policy/offer was given. Perhaps consider returning an empty object here instead?

For example, as if, if no `policy` is defined, this hook will return as if there is a budget that has a `term` of "Ends", which isn't really true in that case.

src/components/learner-credit-management/data/hooks/useBudgetDetailPage.js
```diff
@@ -0,0 +1,76 @@
...
+      available: 0,
+      utilized: 0,
+      limit: 0,
+    },
+    budgetAggregates: policy?.aggregates || {},
+    badgeVariant: 'warning',
+    status: BUDGET_STATUSES.scheduled,
+    term: 'Ends',
+    date: '',
+    isAssignable: policy.isAssignable || false,
```
adamstankiewicz, line 61: Similar to the below comment about `!policy`, if `policy` can be undefined/null, then this line may throw a JS error in that case since it assumes `policy` always exists, unlike line 62 below which uses the `?.` optional chaining syntax.

src/components/learner-credit-management/data/hooks/useOfferSummary.js
adamstankiewicz, line 1: This file can be deleted as it was replaced by `useSubsidySummaryAnalyticsApi` below.

src/components/learner-credit-management/data/hooks/useBudgetDetailPage.js
```diff
@@ -0,0 +1,76 @@
+import { getBudgetStatus } from '../utils';
+import { BUDGET_STATUSES } from '../../../EnterpriseApp/data/constants';
+
+const transformOfferToPolicy = (offer) => {
+  if (!offer) { return null; }
+
+  return {
+    displayName: null,
+    subsidyActiveDatetime: offer.dateCreated,
+    subsidyExpirationDatetime: offer.exportTimestamp,
```
adamstankiewicz, line 10: As discussed in Slack, the metadata needed about the offer regarding start/end dates are not available to this page route at the moment. We will need to integrate with the `enterprise-admin-offers` detail API to request the metadata about the currently viewed enterprise offer from ecommerce to get these dates (and the display name for the offer). I'd recommend implementing a simple `useEnterpriseOffer` hook for the budget detail page that makes the API request using `useQuery`.

Example:

```jsx
const useEnterpriseOffer = (enterpriseOfferId) => {
  return useQuery({
    queryKey: learnerCreditManagementQueryKeys.budgetEnterpriseOffer(enterpriseOfferId),
    queryFn: fetchEnterpriseOffer, // responsible for actually calling the ecommerce `enterprise-admin-offers` detail REST API, and camelCasing the response data.
    enabled: !!enterpriseOfferId, // keep the query disabled unless we are dealing with an enterprise offer
  });
};
```

See [Slack message](https://twou.slack.com/archives/D066M6FC0D7/p1701809304543899) from our DM for more context.

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -0,0 +1,121 @@
...
+};
+
+const BudgetActions = ({ budgetId, isAssignable }) => {
+  const routeMatch = useRouteMatch();
+  const history = useHistory();
+  const supportUrl = configuration.ENTERPRISE_SUPPORT_URL;
+
+  if (!isAssignable) {
+    return (
+      <div className="p-4">
```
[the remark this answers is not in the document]
zwidekalanga: I attempted to implement the following modification, but it appears to be not to be appropriate change, particularly in terms of the page's responsiveness.

**Before**
Screenshot 2023-12-06 at 20 00 04

**After**
Screenshot 2023-12-06 at 19 59 33

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
zwidekalanga: It appears that the `BudgetDetailCatalogTabContents` doesn't re-render when already on the catalog route, which in turn prevents the triggering of the scrollToView functionality within the component.

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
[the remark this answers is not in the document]
adamstankiewicz: @zwidekalanga In that case, you can play with the responsive CSS utility classes for spacing, like `pt-4 pt-lg-0` to give it responsive padding (`4` up until the large breakpoint, then switches to `0`).

adamstankiewicz:
@zwidekalanga I think there's just two last minor hurdles. 

1. When viewing an enterprise offer in the budget detail page and the analytics API errors/doesn't return data, there are a few JS errors thrown. I would recommend applying the following patch to get around this; otherwise, the page entire page will be unusable on stage for enterprise offers due to the JS error.
2. Incorrect field names were being associated for the `spendAvailableUsd ` and `amountRedeemedUsd ` fields within `transformSubsidySummaryToPolicy` (the patch resolves this too).

[budget-detail-header-failing-analytics-api-case.patch](https://github.com/openedx/frontend-app-admin-portal/files/13617324/budget-detail-header-failing-analytics-api-case.patch)

adamstankiewicz (approved):
Approving and merging this PR with [these 2 remaining feedback items](https://github.com/openedx/frontend-app-admin-portal/pull/1110#issuecomment-1847396566) still open; I will apply the linked patch/diff in a separate PR after merging this PR to resolve the 2 open issues.

PR #1114, opened 2023-12-04
feat: background refetching loading state on budgets cards

adamstankiewicz:
# Description

Introduces a background refetching loading state for the budget cards on the Learner Credit Management budgets list page. This is necessary to give the user immediate feedback that the budgets' metadata is refetching (e.g., after assignments were successfully allocated) as opposed to the current state where there is a delay between seeing the original budget aggregates vs. the updated budget aggregates after taking action on the budget (e.g., allocating assignments).

Considered using optimistic updates, where we manually update the query client cache instead of refetching the API, which would improve the perceived performance of the application. However, this preferred approach requires performing calculation(s) around remaining balance, etc. after successful assignment allocation that ideally would live in the backend API (e.g., after successful call to POST `allocate`, it could return the updated `aggregates` for the budget and the client-side optimistic update could rely on data given by the API vs. calculating itself).

Given the (possible) API changes to better support optimistic updates, I'm opting just to handle the background refetching loading state instead (lightest lift).

image

The success `Toast` notification assumes all emails entered were net-new assignments in its "Course successfully assigned to N learners". However, some learners may have already been assigned, where the `Toast` notification message is a bit misleading.

Taking inspiration from Subscriptions (which informs users of net-new subscription learners vs. subscription learners already invited), I've updated the `Toast` messaging to depict the count of learners already assigned to the course in addition to the count of learners with created+updated assignments (allocated or re-allocated, respectively):

image

# For all changes

### DIFF
--- a/src/components/EnterpriseSubsidiesContext/index.jsx
+++ b/src/components/EnterpriseSubsidiesContext/index.jsx
@@ -11,6 +11,7 @@ export const useEnterpriseSubsidiesContext = ({
 }) => {
   const {
     isLoading: isLoadingBudgets,
+    isFetching: isFetchingBudgets,
     data: budgetsOverview,
   } = useEnterpriseBudgets({
     enablePortalLearnerCreditManagementScreen,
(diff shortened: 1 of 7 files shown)


brobro10000 (approved):
LGTM 👍🏽 , connected asynchronously to converse relating to a potential naming conflict with the track event.

PR #1117, opened 2023-12-05
fix: enforce no fraction digits

adamstankiewicz:
# Description

Regarding the Learner Credit Management feature: Currently, if a price includes cents, we show the cents despite the mocks not including any fraction digits.

The `formatPrice` utility function attempts to resolve this by passing `minimumFractionDigits` to `Intl.NumberFormat`; however, really, for the intended behavior to _never_ show fraction digits in prices, we should be using `maximumFractionDigits`.

# For all changes


macdiesel (approved):

adamstankiewicz:
After discussing in Titans scrum, we decided we will continue to show the cents (when relevant) and no cents (when not relevant). We want to avoid issues with rounding when there are cents to avoid misleading data.

PR #1118, opened 2023-12-05
feat: Adds redirect logic for non-existent and inactive budgets

brobro10000:
Changes the way we handle "inactive" budgets returned from `subsidy-access-policy` api and browser behavior when attempting to navigate to an inactive or non-existent budget.

- Checks for if `active == true` before populating the list that generates the budget cards to only show active budgets
- Checks if there is a invalid `subsidyAccessPolicyId` or if the `subsidyAccessPolicy.active === false` before redirecting the user to the budget picker list. 

# For all changes


src/components/learner-credit-management/BudgetDetailPage.jsx
```diff
@@ -28,6 +32,13 @@ const BudgetDetailPage = () => {
     );
   }
 
+  // If the budget has a subsidyAccessPolicy but is not active, or the subsidyAccessPolicyId is invalid
+  // we should redirect the user to the budget list page.
+  if (!subsidyAccessPolicyId || subsidyAccessPolicy?.active === false) {
```
adamstankiewicz, line 37: Keep in mind that the `BudgetDetailPage` works with both subsidy access policies _AND_ enterprise offers. If the budget detail page route is rendered for an enterprise offer, I believe this conditional would execute essentially making it impossible to view budget detail pages for enterprise offers.

This component in particular doesn't really deal with enterprise offers, as the only thing visible for the enterprise offer case is the "Spent" table, backed by the analytics API. That said, you could determine whether you're dealing with an enterprise offer or a subsidy access policy by referring to the `useBudgetId` output, which returns both an `enterpriseOfferId` and the currently used `subsidyAccessPolicyId` (when viewing an offer, the `subsidyAccessPolicyId` is undefined/null).

PR #1128, opened 2023-12-07
fix: modifies product tours state to fix re-rendering error

brobro10000:
Updates how the product tour state is set within the component so on re-render, the component is passing the previously updated state of the `tours` object after it checks the localStorage flags to display the product tour. 
The refetching can still perform as intended but without re-rendering the product tours.

https://github.com/openedx/frontend-app-admin-portal/assets/82611798/b7f7ae61-93d7-4884-b86e-46630b2ead7d

# For all changes

### DIFF
--- a/src/components/ProductTours/ProductTours.jsx
+++ b/src/components/ProductTours/ProductTours.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 import { ProductTour } from '@edx/paragon';
 import { useHistory } from 'react-router-dom';
@@ -29,6 +29,7 @@ const ProductTours = ({
   enableLearnerPortal,
 }) => {
   const { FEATURE_CONTENT_HIGHLIGHTS } = getConfig();
+  const [tours, setTours] = useState([]);
   const enablePortalAppearance = features.SETTINGS_PAGE_APPEARANCE_TAB;
   const history = useHistory();
   const enabledFeatures = {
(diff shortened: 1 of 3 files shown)


macdiesel (approved):
👍 More lines but does not have any change any global variables that could side effect other things the app is doing. Much improved!

src/components/ProductTours/ProductTours.jsx
```diff
@@ -29,15 +29,23 @@ const ProductTours = ({
...
-    [HIGHLIGHTS_COOKIE_NAME]: useHighlightsTour(FEATURE_CONTENT_HIGHLIGHTS)[0],
-  };
-  const newFeatureTourCheckpoints = {
+
+  const portalAppearance = usePortalAppearanceTour({ enablePortalAppearance });
+  const browseAndRequest = useBrowseAndRequestTour({ enableLearnerPortal });
+  const learnerCredit = useLearnerCreditTour();
+  const highlightTour = useHighlightsTour(FEATURE_CONTENT_HIGHLIGHTS);
+
+  const enabledFeatures = useMemo(() => ({
```
adamstankiewicz, line 41: [curious/clarification] Are the changes in this component (beyond changing the return value of the `usePortalAppearanceTour`, `useBrowseAndRequestTour`, etc. hooks) still needed now that these hooks no longer make use of state?

I believe keeping this file as it exists on `master` alongside the changes to the hooks (removing state) still seems to resolve the issue (without requiring a re-render when calling `setTours`).

adamstankiewicz (approved):
LGTM! :shipit:

src/components/ProductTours/ProductTours.jsx
brobro10000: Updated the PR and tested multiple routes of it possibly re-rendering the product tours component. Identical behavior to the first fix in the PR with a cleaner outcome 👍🏽

PR #1129, opened 2023-12-07
fix: adds sane fallback to if learnerState is failed and no error reason exist

brobro10000:
Returns a generalized failed system chip if no error reason exist and the learnerState is 'failed'.
Also refactors return string (capitalize first letter) using and existing utilities function

# For all changes

### DIFF
--- a/src/components/learner-credit-management/AssignmentStatusTableCell.jsx
+++ b/src/components/learner-credit-management/AssignmentStatusTableCell.jsx
@@ -8,6 +8,7 @@ import FailedReminder from './assignments-status-chips/FailedReminder';
 import FailedSystem from './assignments-status-chips/FailedSystem';
 import NotifyingLearner from './assignments-status-chips/NotifyingLearner';
 import WaitingForLearner from './assignments-status-chips/WaitingForLearner';
+import { capitalizeFirstLetter } from '../../utils';
 
 const AssignmentStatusTableCell = ({ row }) => {
   const { original } = row;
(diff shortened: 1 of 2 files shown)


src/components/learner-credit-management/AssignmentStatusTableCell.jsx
```diff
@@ -16,12 +17,16 @@ const AssignmentStatusTableCell = ({ row }) => {
...
-
   // Learner state is not available for this assignment, so don't display anything.
   if (!learnerState) {
     return null;
   }
 
+  // If learnerState is failed but no error reason is defined, don't display anything.
+  if (learnerState === 'failed' && !errorReason) {
+    return <FailedSystem />;
+  }
```
brobro10000, line 28: Was not sure what to return here, whether a failed system chip or null.

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
```diff
@@ -16,12 +17,16 @@ const AssignmentStatusTableCell = ({ row }) => {
...
   } = original;
-
   // Learner state is not available for this assignment, so don't display anything.
   if (!learnerState) {
     return null;
   }
 
+  // If learnerState is failed but no error reason is defined, return a failed system chip.
+  if (learnerState === 'failed' && !errorReason) {
+    return <FailedSystem />;
```
adamstankiewicz, line 27: Note, there is a `learnerState === 'failed'` condition below where this could go (see line 43).

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
adamstankiewicz: Falling back to failed system chip is reasonable IMO :)

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
pwnage101: +1

adamstankiewicz (approved):

PR #1132, opened 2023-12-11
feat: add additional segment events

brobro10000:
Adds the following segment events

- Chip interactions for opening/closing and any links inside the component.
- New assignment / contact for overview
- Overview dropdown interactions
- On cancel/remind
- On bulk cancel/remind

Pending Tests.

# For all changes


src/components/learner-credit-management/AssignmentTableCancel.jsx
```diff
@@ -17,9 +21,21 @@ const calculateTotalToCancel = ({
...
 
-const AssignmentTableCancelAction = ({ selectedFlatRows, isEntireTableSelected, tableInstance }) => {
-  const assignmentUuids = selectedFlatRows.map(row => row.id);
-  const assignmentConfigurationUuid = selectedFlatRows[0].original.assignmentConfiguration;
+const AssignmentTableCancelAction = ({
+  selectedFlatRows, isEntireTableSelected, learnerStateCounts, tableInstance, enterpriseId,
+}) => {
+  const { subsidyAccessPolicyId } = useBudgetId();
+  const { data: subsidyAccessPolicy } = useSubsidyAccessPolicy(subsidyAccessPolicyId);
+  const { assignmentConfiguration } = subsidyAccessPolicy;
```
brobro10000, line 29: Updates how assignment configuration is retrieved or cancel/remind all. Not dependent on the first value of the array having the assignment configuration. Changed for edge case during remind all where none of the filtered selectedRows are remindable, resulting in an empty array.

macdiesel (approved):

src/components/learner-credit-management/assignments-status-chips/FailedSystem.jsx
```diff
@@ -1,29 +1,50 @@
...
   const [isOpen, open, close] = useToggle(false);
   const [target, setTarget] = useState(null);
+  const {
+    BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_FAILED_SYSTEM,
+    BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_FAILED_SYSTEM_HELP_CENTER,
+  } = EVENT_NAMES.LEARNER_CREDIT_MANAGEMENT;
+
+  const openChipModal = () => {
+    open();
+    trackEvent(BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_FAILED_SYSTEM, { isOpen: !isOpen });
```
pwnage101, line 21: Does this need to be `!!isOpen`?

pwnage101: Is this one of those situations where you can't edit state and read the change in the same event loop cycle?  You called open(), but isOpen remains the old value until the next cycle?

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
```diff
@@ -9,13 +11,49 @@ import NotifyingLearner from './assignments-status-chips/NotifyingLearner';
...
+      }
+      : {
+        errorReason: errorReason.errorReason,
+        actionType: errorReason.actionType,
+      };
+    const errorStateMetadata = {
+      ...sharedTrackEventMetadata,
+      ...errorReasonMetadata,
+      ...eventMetadata,
+    };
```
pwnage101, line 49: I think this would be pretty confusing on the receiving end because the resulting payload would contain an `actionType` key, but only when there's an error, and it's the type for the action that failed.  This is not self-evident.  The key name should be something like "failedActionType" instead.

Also, i'm not sure how the eventName is being passed to this function since it's called without parameters below.

src/components/learner-credit-management/assignments-status-chips/FailedSystem.jsx
brobro10000: That is true, this particular components start is inversed at the time we read it for a segment event. 👍🏽

pwnage101 (approved):

src/components/learner-credit-management/AssignmentTableRemind.jsx
```diff
@@ -35,29 +50,82 @@ const AssignmentTableRemindAction = ({
...
+
+  const trackEvent = (eventName) => {
+    // constructs a learner state object for the select all state to match format of select all on page metadata
+    const learnerStateObject = {};
+    learnerStateCounts.forEach((learnerState) => {
+      learnerStateObject[learnerState.learnerState] = learnerState.count;
+    });
+
+    const selectedRowsMetadata = isEntireTableSelected
+      ? { uniqueLearnerState: learnerStateObject, totalSelectedRows: tableInstance.itemCount }
```
adamstankiewicz, line 75: [clarification] In the event of the entire table being selected for reminders, not all table rows are remindable (unlike cancelations, where all rows are cancelable). Is the `totalSelectedRows` misleading in this case given that we would be logging the total number of rows in the table versus the number of remindable rows?

src/components/learner-credit-management/assignments-status-chips/WaitingForLearner.jsx
```diff
@@ -6,26 +6,45 @@ import { getConfig } from '@edx/frontend-platform/config';
...
-const WaitingForLearner = ({ learnerEmail }) => {
+const WaitingForLearner = ({ learnerEmail, trackEvent }) => {
   const [isOpen, open, close] = useToggle(false);
   const [target, setTarget] = useState(null);
+  const {
+    BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_WAITING_FOR_LEARNER,
+    BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_WAITING_FOR_LEARNER_HELP_CENTER,
+  } = EVENT_NAMES.LEARNER_CREDIT_MANAGEMENT;
+
+  const openChipModal = () => {
```
adamstankiewicz, line 19: nit/suggestion: wondering if its worth DRYing up some of the repeated logic around these functions. Would it make sense to create a custom hook that could be re-used across all chips?

```jsx
const {
  openChipModal,
  closeChipModal,
  isChipModalOpen,
} = useAssignmentStatusChip({
  trackEvent,
  chipInteractionEventName: BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_WAITING_FOR_LEARNER,
  chipHelpCenterEventName: BUDGET_DETAILS_ASSIGNED_DATATABLE_CHIP_WAITING_FOR_LEARNER_HELP_CENTER,
});

const useAssignmentStatusChip = ({ chipInteractionEventName, chipHelpCenterEventName, trackEvent }) => {
  const [isChipModalOpen, open, close] = useToggle(false);
  const openChipModal = () => {
    open();
    // Note: could hardcode true given this function *always* opens
    trackEvent(chipInteractionEventName, { isOpen: true });
  };
  const closeChipModal = () => {
    close();
        // Note: could hardcode false given this function *always* closes
    trackEvent(chipInteractionEventName, { isOpen: false });
  };

  const helpCenterTrackEvent = () => {
    trackEvent(chipHelpCenterEventName);
  };

  return {
    isChipModalOpen,
    openChipModal,
    closeChipModal,
  };
};
```

src/components/learner-credit-management/data/utils.js
```diff
@@ -337,3 +337,57 @@ export async function retrieveBudgetDetailActivityOverview({
...
+ * uniqueContentKeys: [String],
+ * uniqueAssignmentState: [String],
+ * totalSelectedRows: Number,
+ * }}
+ */
+export const transformSelectedRows = (selectedFlatRows) => {
+  const assignmentUuids = selectedFlatRows.map(item => item.id);
+  const totalSelectedRows = selectedFlatRows.length;
+
+  // list of unique courses
```
adamstankiewicz, line 359: nit: if the result of this section below is to generate `uniqueLearnerState` (object), should the comment reflect that its creating a map of how many selected per course key? At first read, based on this comment, I was expecting the logic to generate a list of unique strings as opposed to an object keeping track of how many times each course key was selected.

Same feedback for the other sections (i.e., "list of unique learner states", "list of unique assignment states", etc.).

src/components/learner-credit-management/AssignmentTableRemind.jsx
brobro10000: Went with `selectedRemindedRowCount` to populate the `totalSelectedRows` object. 👍🏽

src/components/learner-credit-management/assignments-status-chips/WaitingForLearner.jsx
brobro10000: The abstractions worked out great 👍🏽  way cleaner implementation for toggle-able events

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
brobro10000: Ended up with a uniqued object with your suggestion, ended up with a cleaner result 👍🏽

PR #1133, opened 2023-12-11
fix: update UI padding

brobro10000:
Updates UI components to match figma output for scheduled, expired, and a centered 'view budget' button.

Figma:
Screenshot 2023-12-11 at 11 32 47 AM

Previous UI:
Screenshot 2023-12-11 at 11 33 49 AM

Updated UI:
Screenshot 2023-12-11 at 12 47 37 PM

# For all changes

### DIFF
--- a/src/components/learner-credit-management/SubBudgetCard.jsx
+++ b/src/components/learner-credit-management/SubBudgetCard.jsx
@@ -2,6 +2,7 @@ import { useContext } from 'react';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import dayjs from 'dayjs';
+import classNames from 'classnames';
 import {
   Card,
   Button,
@@ -42,6 +43,7 @@ const SubBudgetCard = ({
       data-testid="view-budget"
       as={Link}
       to={`/${enterpriseSlug}/admin/${ROUTE_NAMES.learnerCredit}/${budgetId}`}
+      variant={budgetLabel.status === BUDGET_STATUSES.expired ? 'outline-primary' : 'primary'}
     >
       View budget
     </Button>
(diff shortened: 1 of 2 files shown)


src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -49,24 +49,26 @@ const SubBudgetCard = ({
 
   const renderCardHeader = (budgetType, budgetId) => {
     const subtitle = (
-      <Stack direction="horizontal" gap={2.5}>
+      <Stack direction="horizontal" gap={2.5} className="pb-4">
```
adamstankiewicz, line 52: IMO, it's a bit odd to add bottom padding to the subtitle vs. the `Card.Header` component itself, given the spacing we want is more in relation to the containing `Card.Header` vs. the subtitle specifically.

I would recommend moving this to a `Card.Header` class name instead. Also, it looks like the `4` spacer value is not quite matching the default top margin of the `Card.Header`, where there's more whitespace at the top of the `Card` than at the bottom (i.e., `Card.Header` uses `4.5` space value).

I would also recommend conditionally passing this class name for it to return only when necessary (current, non-expired/scheduled budgets):

```jsx

<Card.Header
  className={classNames({
      'mb-4.5': budgetLabel.status !== BUDGET_STATUSES.active,
  })}
  // ...
>
```

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -49,24 +49,26 @@ const SubBudgetCard = ({
...
+      : undefined;
     return (
-      <Card.Header
-        title={<BackgroundFetchingWrapper>{budgetType}</BackgroundFetchingWrapper>}
-        subtitle={<BackgroundFetchingWrapper>{subtitle}</BackgroundFetchingWrapper>}
-        actions={
-          budgetLabel.status !== BUDGET_STATUSES.scheduled
-            ? renderActions(budgetId)
-            : undefined
-        }
```
adamstankiewicz, line 68: I would suggest leaving the CTA within the `actions` prop of `Card.Header` (to take advantage of default styling / sizes of the action CTAs for `Card`). In order to get the action to be vertically centered within the `Card.Header` (instead of the default positioning), you could likely pass the `align-items-center` class name to the `Card.Header` component.

```jsx
<Card.Header
  className={classNames(
    'align-items-center',
    {
      'mb-4.5': budgetLabel.status !== BUDGET_STATUSES.active,
    }
  )}
  // ...
>
```

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -49,24 +49,26 @@ const SubBudgetCard = ({
...
-      />
+      <Stack direction="horizontal" className="justify-content-between">
+        <Card.Header
+          title={<BackgroundFetchingWrapper>{budgetType}</BackgroundFetchingWrapper>}
+          subtitle={<BackgroundFetchingWrapper>{subtitle}</BackgroundFetchingWrapper>}
+        />
+        <div className="pr-3">
+          {action && action}
+        </div>
```
adamstankiewicz, line 70: [inform] This would render the wrapping `div` even when an `action` is not present for the budget (i.e., scheduled).

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -49,24 +49,26 @@ const SubBudgetCard = ({
...
-      <Card.Header
-        title={<BackgroundFetchingWrapper>{budgetType}</BackgroundFetchingWrapper>}
-        subtitle={<BackgroundFetchingWrapper>{subtitle}</BackgroundFetchingWrapper>}
-        actions={
-          budgetLabel.status !== BUDGET_STATUSES.scheduled
-            ? renderActions(budgetId)
-            : undefined
-        }
-      />
+      <Stack direction="horizontal" className="justify-content-between">
```
adamstankiewicz, line 63: This `Stack` likely could go away based on other feedback comments.

src/components/learner-credit-management/SubBudgetCard.jsx
```diff
@@ -106,9 +108,9 @@ const SubBudgetCard = ({
       isLoading={isLoading}
     >
       <Card.Body>
-        <Stack gap={4}>
+        <Stack>
```
adamstankiewicz, line 111: I would recommend leaving the `Stack` component here for the spacing between the `Card.Header` and the below `Card.Section`, but only provide the `gap={4.5}` prop conditionally depending on the budget status. Also note that this `gap` value should be `4.5` to be consistent with the `Card.Header` default top margin.

adamstankiewicz (approved):

PR #1137, opened 2023-12-13
fix: update variant on new assignment button

brobro10000:
Makes the button red for new assignments to match figma. 

Screenshot 2023-12-13 at 4 29 39 PM

# For all changes


iloveagent57 (approved):

adamstankiewicz:
@brobro10000 the button is supposed to change its color depending on the theme set by the enterprise customer in the admin portal.

By using the brand variant as it was before, the button color will dynamically change to the primary color set by the enterprise customer in the "settings" page of the admin portal.

The danger variant does not get the same dynamic theming support and is reserved for destructive actions, which this is not.

We should change this back to brand.

PR #1142, opened 2023-12-15
fix: account for filters in select all bulk actions assigned table

adamstankiewicz:
# Description

Ticket: [ENT-8155](https://2u-internal.atlassian.net/browse/ENT-8155)

If filters are applied to the "Assigned" table, continue to call the standard bulk action API endpoints. Only call the `{remind/cancel}-all` API endpoints when the entire table is selected with NO filters applied.

In the below screenshot, note the 32 rows selected, but only the 25 rows on current page will be applicable (and shown in the count). This avoids unexpectedly applying bulk actions to non-filtered rows, but is still not quite the right behavior (i.e., the expected behavior would be performing the action on only on all _filtered_ rows). It's at least a short term fix to prevent unexpected cancelations and reminders, until something like [ENT-8156](https://2u-internal.atlassian.net/browse/ENT-8156) is implemented where the `{remind|cancel}-all` API endpoints can accept the state of table filters, if it can't already.

image

# For all changes

### DIFF
--- a/src/components/subscriptions/data/utils.js
+++ b/src/components/subscriptions/data/utils.js
@@ -65,15 +65,3 @@ export const transformFiltersForRequest = (filters) => {
     }),
   );
 };
-
-/**
- * Helper to determine which table columns have an active filter applied.
- *
- * @param {object} columns Array of column objects (e.g., { id, filter, filterValue })
- * @returns Array of column objects with an active filter applied.
- */
-export const getActiveFilters = columns => columns.map(column => ({
-  name: column.id,
-  filter: column.filter,
-  filterValue: column.filterValue,
-})).filter(filter => !!filter.filterValue);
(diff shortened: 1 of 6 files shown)


src/components/subscriptions/data/utils.js
```diff
@@ -65,15 +65,3 @@ export const transformFiltersForRequest = (filters) => {
...
   );
 };
-
-/**
- * Helper to determine which table columns have an active filter applied.
- *
- * @param {object} columns Array of column objects (e.g., { id, filter, filterValue })
- * @returns Array of column objects with an active filter applied.
- */
-export const getActiveFilters = columns => columns.map(column => ({
```
adamstankiewicz, line 75: Moved to `src/utils.jsx` to be shared with any DataTable, not just in subscriptions.

brobro10000 (approved):
LGTM 👍🏽  Great catch.

iloveagent57 (approved):

PR #1147, opened 2024-01-04
fix: resolves refetching issue resulting in premature FullscreenModal closure

adamstankiewicz:
Adam:
Implements a fix by removing `isFetching` as `isFetchingBudgets` from the `EnterpriseSubsidiesContext` context layer
Hamzah: 
Removes `budgets` from the `EnterpriseSubsidiesContext` context layer which is derived from the `react-query` cache to be intentional about separating the local context and the `react-query` cache.
Adds `mapStateToProps` to the required components that make the useQuery hook call to retrieve budgets
Updates the tests accordingly.

Includes the ADR which highlights the intentional usage of `react-query` and the separation of concerns between the local state (context/dispatch) and the state created by `react-query` for API data cache.
# For all changes


docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,45 @@
...
+An example where populating context data from ``react-query`` cache data resulted in unintentional behavior is the ``EnterpriseSubsidiesContext`` containing API data from the query key 'learner-credit-management.budgets', ``isFetchingBudgets`` attribute which has since been fixed.
+
+While the user was interacting with a ``FullscreenModal`` component and had not completed their intended activity within the modal before the global ``staleTime`` had expired, the re-fetch behavior would trigger resulting in a full page re-load, thus closing the modal without saving the users state information (form data etc.).
+
+Decisions
+*********
+
+We will take a more intentional approach when implementing API data fetching using the ``react-query`` in the following ways:
+
+* Avoid adding API data fetched using ``react-query`` to a context layer
```
macdiesel, line 31: Could we use a numbered list?

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,45 @@
...
+    * If a context layer does require API data to exist in the context layer, do not use ``react-query`` to retrieve the data
+* When using ``react-query`` make the call as close to the component that uses the returned response data
+    * This may require mapping a component from the parent component which completes the API call
```
macdiesel, line 34: Would it be useful to include a code example of how this would differ?

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+Accepted  (January 2024)
+
+Context
+*******
+
+The ``frontend-app-admin-portal`` currently implements ``@tanstack/react-query`` predominantly within the ``learner-credit-management`` feature. ``@tanstack/react-query`` is a tool used to manage server state in react applications.
+
+See `0006-tanstack-react-query ADR <https://github.com/openedx/frontend-app-admin-portal/blob/master/docs/decisions/0006-tanstack-react-query.rst>`_ for more information
+
+As part of its implementation, a pattern has emerged of API data from the react-query cache is being populated within context providers. This has resulted in the unintended consequence of page refreshes when the top-level component context data is updated with the latest data from the ``react-query`` cache.
```
adamstankiewicz, line 16: nit: in the case of `isFetchingBudgets`, that data isn't really part of the react-query cache; the cache data is the actual API response, not the state of the query (i.e., `isLoading`, `isFetching`, `hasError`, etc.).

```suggestion
As part of its implementation, a pattern has emerged of API data from the react-query output from ``useQuery`` is being populated within context providers. This has resulted in the unintended consequence of component re-renders when the top-level context data is updated with the latest data from the ``useQuery`` output.
```

adamstankiewicz, line 16: nit: `react-query` has backticks around it in some places, but not here. Should they be consistent?

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+
+The re-fetch behavior of ``react-query`` results in page components that aren't dependent on the updated context data to reload.
```
adamstankiewicz, line 18: nit: 

```suggestion
The re-fetch behavior of ``react-query`` results in components that aren't dependent on the updated context data to re-render.
```

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+
+The re-fetch behavior was occurring in the background of a component while the user is interacting with the page or if they had returned focus back onto the browser while interacting with other tabs or applications.
+
+An example where populating context data from ``react-query`` cache data resulted in unintentional behavior is the ``EnterpriseSubsidiesContext`` containing API data from the query key 'learner-credit-management.budgets', ``isFetchingBudgets`` attribute which has since been fixed.
+
+While the user was interacting with a ``FullscreenModal`` component and had not completed their intended activity within the modal before the global ``staleTime`` had expired, the re-fetch behavior would trigger resulting in a full page re-load, thus closing the modal without saving the users state information (form data etc.).
```
adamstankiewicz, line 24: It's not quite a full page re-load, but a re-rendering of all components that subscribe to the ``EnterpriseSubsidiesContext``. To me, "full page re-load" is like a hard refresh where all JS assets need to be re-downloaded, etc. Re-rendering components doesn't involve a full page re-load.

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Decisions
+*********
+
+We will take a more intentional approach when implementing API data fetching using the ``react-query`` in the following ways:
+
+#. Avoid adding API data fetched using ``react-query`` to a context layer
```
adamstankiewicz, line 31: Might expand on this to mention avoiding API data but also the lifecycle state around queries (e.g., `isLoading`, isFetching`, etc.).

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+    * If a context layer does require API data to exist in the context layer, do not use ``react-query`` to retrieve the data
```
adamstankiewicz, line 32: [curious] While I agree with the sentiment of this, I'm wondering if we should extend it to say something along the lines of:

```suggestion
    * If a context layer does require API data to exist in the context layer, though not recommended, do not use ``react-query`` to retrieve the data
```

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+#. When using ``react-query`` make the call as close to the component that uses the returned response data
```
adamstankiewicz, line 33: [clarification] Is this item getting at how we should simply rely on calling the custom hook wrapping `useQuery` where relevant in the component tree as opposed to only calling it once higher up in the component tree and sharing its result with context?

That is, if you need API data, call ``useQuery`` directly (generally, regardless of where in the component tree), as opposed to trying to expose query data via some other mechanism besides the ``useQuery`` cache?

Perhaps this could be re-worded a bit to make this a bit more clear? That is, no need to expose API data/state in context layer because the ``useQuery`` hook can be called numerous times during rendering without risking duplicate API calls.

I also have a similar question as Brian's below around not fully understanding what is meant by "may require mapping a component from the parent component which completes the API call".

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,104 @@
...
+    * This may require mapping a component from the parent component which completes the API call
+#. Leverage the `query key factory <https://github.com/openedx/frontend-app-admin-portal/blob/c67c5e4d8a0328fe75cb9d46791a8b733fad8257/src/components/learner-credit-management/data/constants.js#L67-L77>`_ setup to simplify boilerplate code when making the `useQuery` calls and `invalidateQueries`
```
adamstankiewicz, line 35: super tiny nit: perhaps "pattern" vs. "setup"?

```suggestion
#. Leverage the `query key factory <https://github.com/openedx/frontend-app-admin-portal/blob/c67c5e4d8a0328fe75cb9d46791a8b733fad8257/src/components/learner-credit-management/data/constants.js#L67-L77>`_ pattern to simplify boilerplate code when making the `useQuery` calls and `invalidateQueries`
```

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,177 @@
...
+#. Implement a hook when calling the ``useQuery`` method to avoid duplicate code and easy maintainability. See this `hook <https://github.com/openedx/frontend-app-admin-portal/blob/c67c5e4d8a0328fe75cb9d46791a8b733fad8257/src/components/EnterpriseSubsidiesContext/data/hooks.js#L108-L120>`_ as boilerplate as an example, there are `others to reference in the repo <https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-admin-portal+%2FuseQuery%5C%28%2F&type=code>`_ as additional guides
+#. Invalidate queries when user interaction results in an API data change to avoid stale data
```
adamstankiewicz, line 37: Note, there is an alternative to query invalidation. If the change in server state is known, we can also consider manually update/modify the data stored in the query cache, where relevant, without needing to invalidate or re-fetch from the API, resulting in more performant UIs. Might be worth including that as an option here, too?

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,176 @@
...
+    * If the query is not invalidated, the re-fetching of the data is entirely dependent on the length of time set for the ``staleTime``
+
+Consequences
+************
+
+Based on the call you're making, it may be challenging to coalesce all the required attributes to make the appropriate API call.
```
adamstankiewicz, line 43: [curious] I'm wondering if this bit around `mapStateToProps` is necessary to include? It's not really directly related to keeping server state in query cache vs. server state in context. Regardless of either approach, the coalescing of such metadata from the Redux store would be a consideration, no?

I think the ADR could probably be simplified if these Redux concerns are excluded.

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,176 @@
...
+      enterpriseUUID:PropTypes.string.isRequired,
+      enterpriseFeatures: PropTypes.shape({
+        topDownAssignmentRealTimeLcm: PropTypes.bool,
+      }).isRequired,
+      enablePortalLearnerCreditManagementScreen: PropTypes.bool.isRequired,
+    };
+
+    export default connect(mapStateToProps)(TestComponent);
+
+When writing tests for a hook that leverages useQuery, certain boilerplate code is required. Below is an example of how writing tests with a useQuery directly in the component in the form of a hook,
```
adamstankiewicz, line 100: I think it's useful to callout test implications, but not sure if a complete code example is needed in the ADR? Would simply calling out the need to wrap hook/component with `QueryClientProvider` when using react-query, along with the mention of `queryClient` test  utility function. Perhaps we link out to code snippet as the full example, but keep the textual info around it brief? 🤔

docs/decisions/0008-application_state_differenciation.rst
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+
+
+
+
+Alternatives Considered
+***********************
+
+The alternative
```
adamstankiewicz, line 176: Might mention:
* disabling/modifying of React Query's refetch behavior (i.e., remove refetches altogether, extend `staleTime`).
* Attempting to migrate to context selectors (i.e., `use-context-selectors`) to mitigate re-rendering concerns.

docs/decisions/0008-application_state_differenciation.rst
brobro10000: Was also debating this line as being too harsh as to bar individuals from using this methodology, will move forward with your recommendation

docs/decisions/0008-application_state_differenciation.rst
brobro10000: Noted and removed 👍🏽

docs/decisions/0008-application_state_differenciation.rst
brobro10000: I parred down the example of the bare minimum required to construct the wrapper as opposed to providing the example component and test file.

docs/decisions/0008-application_state_differenciation.rst
brobro10000: I felt like I was over indexing on implementation details of useQuery here. Opted to remove this line to avoid obscurity.

brobro10000 (approved):
LGTM 👍🏽 Good job me!

PR #1153, opened 2024-01-09
feat: auto setting active linked enterprise

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8038

# For all changes

### DIFF
--- a/src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
+++ b/src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
@@ -1,0 +1,42 @@
+import { useEffect } from 'react';
+
+import {
+  useMutation, useQuery,
+} from '@tanstack/react-query';
+import { logError } from '@edx/frontend-platform/logging';
+
+import LmsApiService from '../../../../data/services/LmsApiService';
+
+const useUpdateActiveEnterpriseForUser = ({ enterpriseId, user }) => {
+  // Sets up POST call to update active enterprise.
+  const { mutate, isLoading: isUpdatingActiveEnterprise } = useMutation({
+    mutationFn: () => LmsApiService.updateUserActiveEnterprise(enterpriseId),
+    onError: () => {
+      logError("Failed to update user's active enterprise");
+    },
+  });
+  const { username } = user;
+  const {
+    data,
+    isLoading: isLoadingActiveEnterprise,
+  } = useQuery({
+    queryKey: ['activeLinkedEnterpriseCustomer', username],
+    queryFn: () => LmsApiService.getActiveLinkedEnterprise(username),
+    meta: {
+      errorMessage: "Failed to fetch user's active enterprise",
+    },
+  });
+
+  useEffect(() => {
+    if (!data) { return; }
+    if (data.uuid !== enterpriseId) {
+      mutate(enterpriseId);
+    }
+  }, [data, enterpriseId, mutate]);
+
+  return {
+    isLoading: isLoadingActiveEnterprise || isUpdatingActiveEnterprise,
+  };
+};
+
+export default useUpdateActiveEnterpriseForUser;
--- a/src/data/services/LmsApiService.js
+++ b/src/data/services/LmsApiService.js
@@ -1,5 +1,6 @@
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { camelCaseObject } from '@edx/frontend-platform/utils';
+import { logError } from '@edx/frontend-platform/logging';
 
 import { configuration } from '../../config';
 import generateFormattedStatusUrl from './apiServiceUtils';
@@ -383,6 +384,39 @@ class LmsApiService {
   static generateAIAnalyticsSummary(enterpriseUUID, formData) {
     const url = `${LmsApiService.baseUrl}/enterprise/api/v1/analytics-summary/${enterpriseUUID}`;
     return LmsApiService.apiClient().post(url, formData);
+  }
+
+  static updateUserActiveEnterprise = (enterpriseId) => {
+    const url = `${configuration.LMS_BASE_URL}/enterprise/select/active/`;
+    const formData = new FormData();
+    formData.append('enterprise', enterpriseId);
+
+    return LmsApiService.apiClient().post(
+      url,
+      formData,
+    );
+  };
+
+  static fetchEnterpriseLearnerData(options) {
+    const enterpriseLearnerUrl = `${configuration.LMS_BASE_URL}/enterprise/api/v1/enterprise-learner/`;
+    const queryParams = new URLSearchParams({
+      ...options,
+      page: 1,
+    });
+    const url = `${enterpriseLearnerUrl}?${queryParams.toString()}`;
+    return LmsApiService.apiClient().get(url);
+  }
+
+  static async getActiveLinkedEnterprise(username) {
+    const response = await this.fetchEnterpriseLearnerData({ username });
+    const transformedResponse = camelCaseObject(response.data);
+    const enterprisesForLearner = transformedResponse.results;
+    const activeLinkedEnterprise = enterprisesForLearner.find(enterprise => enterprise.active);
+    if (!activeLinkedEnterprise) {
+      logError(`${username} does not have any active linked enterprise customers`);
+      return null;
+    }
+    return activeLinkedEnterprise.enterpriseCustomer;
   }
 }
 
--- a/src/components/App/index.jsx
+++ b/src/components/App/index.jsx
@@ -4,6 +4,7 @@ import {
 } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
 import {
+  QueryCache,
   QueryClient,
   QueryClientProvider,
 } from '@tanstack/react-query';
@@ -25,9 +26,12 @@ import { SystemWideWarningBanner } from '../system-wide-banner';
 
 import store from '../../data/store';
 import { ROUTE_NAMES } from '../EnterpriseApp/data/constants';
-import { defaultQueryClientRetryHandler } from '../../utils';
+import { defaultQueryClientRetryHandler, queryCacheOnErrorHandler } from '../../utils';
 
 const queryClient = new QueryClient({
+  queryCache: new QueryCache({
+    onError: queryCacheOnErrorHandler,
+  }),
   defaultOptions: {
     queries: {
       retry: defaultQueryClientRetryHandler,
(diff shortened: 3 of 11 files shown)


src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
```diff
@@ -0,0 +1,40 @@
...
+
+import LmsApiService from '../../../../data/services/LmsApiService';
+
+const useUpdateActiveEnterpriseForUser = ({
+  enterpriseId,
+  user,
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+
+  const updateActiveEnterpriseAndRefreshJWT = useCallback(async (username, currentEnterpriseId) => {
```
adamstankiewicz, line 12: nit: likely don't necessarily need to pass `currentEnterpriseId`. This hook function itself gets `enterpriseId` as an argument, and it'd be available within `updateActiveEnterpriseAndRefreshJWT` without it as an explicit function argument.

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
```diff
@@ -0,0 +1,40 @@
+import { useState, useEffect, useCallback } from 'react';
+import { logError } from '@edx/frontend-platform/logging';
+
+import LmsApiService from '../../../../data/services/LmsApiService';
+
+const useUpdateActiveEnterpriseForUser = ({
+  enterpriseId,
+  user,
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
```
adamstankiewicz, line 10: [curious, non-blocking] Might consider what scope is involved in refactoring this hook a bit to utilize `@tanstack/react-query`? What benefits might we get from React Query in relation to the scope? If the benefits are not significant in this particular use case, OK to not utilize React Query here. 

E.g., removes boilerplate around needing to manage our own loading states, etc.

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const updateActiveEnterpriseAndRefreshJWT = useCallback(async (username, currentEnterpriseId) => {
+    setIsLoading(true);
+    try {
+      await LmsApiService.getActiveLinkedEnterprise(username).then(async (linkedEnterpriseId) => {
+        if (linkedEnterpriseId !== currentEnterpriseId) {
+          await LmsApiService.updateUserActiveEnterprise(currentEnterpriseId);
+          await LmsApiService.loginRefresh();
```
adamstankiewicz, line 18: [curious] I'm wondering if we need the `loginRefresh` call here anymore now that we're no longer making assumptions of JWT role ordering. That is, I believe we were calling `loginRefresh` because our prior assumption that the JWT roles would be re-ordered.

Given that this hook no longer cares about the JWT roles, could we nix the call `login_refresh`?

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
     const url = `${LmsApiService.baseUrl}/enterprise/api/v1/analytics-summary/${enterpriseUUID}`;
     return LmsApiService.apiClient().post(url, formData);
   }
+
+  static updateUserActiveEnterprise = (enterpriseId) => {
```
adamstankiewicz, line 389: nit: might co-locate this function with the others related to active enterprise (i.e., `fetchEnterpriseLearnerData`, `getActiveLinkedEnterprise`).

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
...
+        // Clean up the cookie if it exists to eliminate any situation
+        // where the cookie is not expired but the jwt is expired.
+        const cookies = new Cookies();
+        cookies.remove(configuration.ACCESS_TOKEN_COOKIE_NAME);
+      }
+      return Promise.resolve();
+    }
+  };
+
+  static fetchEnterpriseLearnerData(options) {
```
adamstankiewicz, line 417: Might want to have test coverage on these new service methods to ensure the correct API endpoints with expected query params are called.

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
...
+    const queryParams = new URLSearchParams({
+      ...options,
+      page: 1,
+    });
+    const url = `${enterpriseLearnerUrl}?${queryParams.toString()}`;
+    return LmsApiService.apiClient().get(url);
+  }
+
+  static getActiveLinkedEnterprise(username) {
+    return this.fetchEnterpriseLearnerData({ username }).then((response) => {
```
adamstankiewicz, line 428: nit: might recommend using `async/await` over `.then` chaining.

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
...
+      const { data } = response;
+      const results = data?.results;
+      for (let i = 0; i < results.length; i++) {
```
adamstankiewicz, line 431: [suggestion] might be able to trim/simplify this logic down a bit by relying on `.find(...)` instead of explicitly looping through each record to find the `active=true`?

```jsx
static async getActiveLinkedEnterprise(username) {
    const response = await this.fetchEnterpriseLearnerData({ username });
    const transformedResponse = camelCaseObject(response.data);
    const enterprisesForLearner = transformedResponse.results;
    const activeLinkedEnterprise = enterprisesForLearner.find(enterprise => enterprise.active);
    if (!activeLinkedEnterprise) {
      logError(`${username} does not have any active linked enterprise customers`);
    }
}
```

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
...
+      page: 1,
+    });
+    const url = `${enterpriseLearnerUrl}?${queryParams.toString()}`;
+    return LmsApiService.apiClient().get(url);
+  }
+
+  static getActiveLinkedEnterprise(username) {
+    return this.fetchEnterpriseLearnerData({ username }).then((response) => {
+      const { data } = response;
+      const results = data?.results;
```
adamstankiewicz, line 430: [curious] When would `data` be undefined? Do we need to use the optional chaining you have here? If we do, when `data` is undefined, `results` would also be undefined and would subsequently throw a JS error on line 431 when trying to access `results.length`.

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,57 @@ class LmsApiService {
...
+      for (let i = 0; i < results.length; i++) {
+        if (results[i].active) {
+          return results[i].uuid;
+        }
+      }
+      return '';
```
adamstankiewicz, line 436: If there is no active enterprise customer (super edge case, one should always exist), likely should return `undefined` or `null` instead of empty string.

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
alex-sheehan-edx: I didn't realize useCallback's get access to arguments passed to the hook!

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
alex-sheehan-edx: oh yeah, that's a great point! It's all done properly now 😄

src/data/services/LmsApiService.js
alex-sheehan-edx: with the removal of loginRefresh they're already co-located! 👍

src/data/services/LmsApiService.js
alex-sheehan-edx: I think I was mostly just following the pattern in the file- but I agree I can't think of a reason why it would be undefined.

src/data/services/LmsApiService.js
alex-sheehan-edx: added!

src/data/services/LmsApiService.js
```diff
@@ -384,6 +385,38 @@ class LmsApiService {
...
+
+  static async getActiveLinkedEnterprise(username) {
+    const response = await this.fetchEnterpriseLearnerData({ username });
+    const transformedResponse = camelCaseObject(response.data);
+    const enterprisesForLearner = transformedResponse.results;
+    const activeLinkedEnterprise = enterprisesForLearner.find(enterprise => enterprise.active);
+    if (!activeLinkedEnterprise) {
+      logError(`${username} does not have any active linked enterprise customers`);
+    }
+    return activeLinkedEnterprise.enterpriseCustomer;
```
adamstankiewicz, line 418: Ohh, while edge casey, I believe this would throw a JS error if there is no active linked enterprise. If the conditional on line 415 is true, we can't access the `.enterpriseCustomer` property on `undefined` on this line.

Given this, any caller of `getActiveLinkedEnterprise` should likely handle when `undefined` is returned, e.g.:

```jsx
const activeLinkedEnterprise = await LmsApiService.getActiveLinkedEnterprise(username);
if (!activeLinkedEnterprise) { return; }
if (activeLinkedEnterprise.uuid !== enterpriseId) { ... }
```

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
```diff
@@ -0,0 +1,29 @@
...
+import {
+  useQuery,
+} from '@tanstack/react-query';
+
+import LmsApiService from '../../../../data/services/LmsApiService';
+
+const useUpdateActiveEnterpriseForUser = ({ enterpriseId, user }) => {
+  const { username } = user;
+  const { isLoading, error } = useQuery({
+    queryKey: ['updateUsersActiveEnterprise'],
```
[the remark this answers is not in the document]
alex-sheehan-edx: makes sense!

src/components/EnterpriseApp/data/hooks/useUpdateActiveEnterpriseForUser.js
```diff
@@ -0,0 +1,29 @@
...
+    queryFn: async () => {
+      await LmsApiService.getActiveLinkedEnterprise(username).then(async (linkedEnterprise) => {
+        if (linkedEnterprise.uuid !== enterpriseId) {
+          await LmsApiService.updateUserActiveEnterprise(enterpriseId);
```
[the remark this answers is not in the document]
alex-sheehan-edx: ahh I see, on successfully querying the active linked org, we use `mutateAsync`.. ok ok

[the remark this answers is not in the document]
alex-sheehan-edx: we might have to tweak with the deprecation of the callbacks

adamstankiewicz (approved):

adamstankiewicz (approved):

PR #1154, opened 2024-01-17
fix: update path generation to use generatePath

brobro10000:
Fixes regression where given a url path includes the `pathname` from `useLocation`, a trailing forward slash may or may not exist in the url pathname.  Ex.  URL: `portal.edx.org/test-slug/` , pathname: `/test-slug/`
If a pathname is used with a trailing slash to construct a URL without using the `generatePath` method, an edge case arises where the user may navigate to  `portal.edx.org/test-slug/` and be redirected to `portal.edx.org/test-slug//admin/learners` due to the trailing forward slash.

See video for example.

Before fix:
https://github.com/openedx/frontend-app-admin-portal/assets/82611798/c5a1733c-f4df-46dd-b127-41338fa82eed

Post fix:
https://github.com/openedx/frontend-app-admin-portal/assets/82611798/8877fd2d-92b8-4611-8f7d-04323b63ff9a

# For all changes

### DIFF
--- a/src/components/App/index.jsx
+++ b/src/components/App/index.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useMemo } from 'react';
 import {
-  Route, Navigate, Routes, useLocation, generatePath,
+  Route, Navigate, Routes, generatePath, useParams,
 } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
 import {
@@ -42,10 +42,13 @@ const queryClient = new QueryClient({
 });
 
 const RedirectComponent = () => {
-  const location = useLocation();
+  const { enterpriseSlug } = useParams();
+
   const homePage = generatePath(
-    `${location.pathname}/admin/${ROUTE_NAMES.learners}`,
+    `/:enterpriseSlug/admin/${ROUTE_NAMES.learners}`,
+    { enterpriseSlug },
   );
+
   return <Navigate to={homePage} />;
 };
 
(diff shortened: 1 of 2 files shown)


macdiesel (approved):
Taking your word on this one.

src/components/App/index.jsx
```diff
@@ -116,7 +122,7 @@ const AppWrapper = () => {
             element={<PageWrap><UserActivationPage /></PageWrap>}
           />
           <Route
-            path="/:enterpriseSlug"
+            path="/:enterpriseSlug/admin?"
```
brobro10000, line 125: Updated to handle regression introduced where paths at `/:enterpriseSlug/admin` will no longer redirect and throw a 404

The `?` references the `/admin` path as optional to handle 404 on the routes `/:enterpriseSlug/admin` to redirect to `/:enterpriseSlug/admin/learners` as expected

src/components/App/index.jsx
```diff
@@ -42,8 +42,14 @@ const queryClient = new QueryClient({
...
 
 const RedirectComponent = () => {
-  const location = useLocation();
-  return <Navigate to={`${location.pathname}/admin/${ROUTE_NAMES.learners}`} />;
+  const { enterpriseSlug } = useParams();
+
+  const homePage = generatePath(
+    `/:enterpriseSlug/admin/${ROUTE_NAMES.learners}`,
+    { enterpriseSlug },
+  );
```
brobro10000, line 50: By directly referencing the enterprise slug from the route parameters, we are no longer reliant on the `pathname` from `useLocation`

PR #1160, opened 2024-01-22
feat: add cookies for archived courses in highlights

katrinan029:
## Description
Set cookies in the admin portal to indicate whether the admin's created highlights have newly archived courses. The goal is to display a red notification bubble on the 'Highlights' menu item in the sidebar whenever a course in their highlights is archived. This notification will persist until the user opens the Highlights summary page and dismisses the alert.

## UI Changes

https://github.com/openedx/frontend-app-admin-portal/assets/71999631/7599da01-0992-434a-a4ef-2d85f7e9a239

[JIRA](https://2u-internal.atlassian.net/browse/ENT-8239)
# For all changes


src/components/ContentHighlights/data/hooks.js
```diff
@@ -143,3 +149,71 @@ export function useContentHighlightsContext() {
...
+        logError(e);
+      } finally {
+        setIsLoading(false);
+      }
+
+      const highlightedSetContent = highlightSet?.highlightedContent;
+      // checks that the highlight uuid isn't set
+      if (!dismissedArchivedCourses[highlight.uuid]) {
+        highlightedSetContent?.forEach(content => {
+          // if the length is greater than 1 for the course run status e.g. ["archived", "published"]
```
kiram15, line 185: I don't believe this is true, a course could have multiple archived course runs. The main thing is that for a course to be archived all of its course runs have to have that status as well.

src/components/ContentHighlights/ContentHighlightArchivedCoursesAlert.jsx
```diff
@@ -0,0 +1,53 @@
...
+import { enterpriseCurationActions } from '../EnterpriseApp/data/enterpriseCurationReducer';
+
+const ContentHighlightArchivedCoursesAlert = () => {
+  const archivedCoursesCookies = {};
+  const [isOpen, setIsOpen] = useState(true);
+
+  const { enterpriseCuration: { enterpriseHighlightedContents, dispatch } } = useContext(EnterpriseAppContext);
+  enterpriseHighlightedContents.forEach(content => {
+    content.highlightedContent.forEach(course => {
+      if (course.courseRunStatuses?.length === 1 && course.courseRunStatuses.includes('archived')) {
```
marlonkeating, line 15: Should we be checking if all the courseRunStatuses are archived here?

nit: `'archived'` should live in a constant somewhere

src/components/ContentHighlights/ContentHighlightArchivedCoursesAlert.jsx
```diff
@@ -0,0 +1,53 @@
...
+        } else {
+          archivedCoursesCookies[content.uuid] = [course.contentKey];
+        }
+      }
+    });
+  });
+
+  const setArchivedCourseCookies = () => {
+    const currentDate = dayjs();
+    // Chrome sets an cap on expiration limit for cookies to 400 days
```
marlonkeating, line 27: nit
```suggestion
    // Chrome sets a cap on expiration limit for cookies to 400 days
```

src/components/EnterpriseApp/data/hooks/useEnterpriseCuration.js
```diff
@@ -73,6 +85,11 @@ function useEnterpriseCuration({ enterpriseId, curationTitleForCreation }) {
         if (!curation) {
           curation = await createEnterpriseCuration();
         }
+        const curationHighlightedContents = await Promise.all(curation.highlightSets.map(async (highlightSet) => {
+          const data = await getHighlightSet(highlightSet.uuid).then(error => error);
```
marlonkeating, line 89: What does `.then(error => error)` do here?

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -43,18 +48,59 @@ export const enterpriseCurationActions = {
...
 };
 
 function getHighlightSetsFromState(state) {
   return state.enterpriseCuration?.highlightSets || [];
 }
 
+function getIsNewArchivedCourseFromState(payload) {
+  const dismissedCookies = archivedHighlightsCoursesCookies.get(NEW_ARCHIVED_COURSE_ALERT_DISMISSED_COOKIE_NAME);
+  const isDismissedArchivedCoursesSet = dismissedCookies !== undefined;
+  const dismissedArchivedCourses = isDismissedArchivedCoursesSet ? dismissedCookies : {};
```
marlonkeating, line 72: nit
```suggestion
  const dismissedArchivedCourses = dismissedCookies || {};
```

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -43,18 +48,59 @@ export const enterpriseCurationActions = {
...
+  let isNewArchivedCourse = false;
+  // checks that the highlight uuid isn't set
+  payload?.forEach(highlightSet => {
+    if (!dismissedArchivedCourses[highlightSet.uuid]) {
+      highlightSet.highlightedContent.forEach(courseContent => {
+        // if the length is greater than 1 for the course run status e.g. ["archived", "published"]
+        // the course should not be archived
+        if (courseContent.courseRunStatuses?.length === 1 && courseContent.courseRunStatuses?.includes('archived')) {
```
adamstankiewicz, line 80: [curious] What about a situation where there's an archived course run and an unpublished course run:

```js
courseContent.courseRunStatuses = ['archived', 'unpublished']
```

Given that the the unpublished course run is effectively non-existent to learners, would the only relevant course run status remaining be archived, in which `isNewArchivedCourse` might be truthy in this case? 🤔

adamstankiewicz, line 80: [sanity check] Just to verify for my own understanding, this feature intends to only work with course content? That is, this feature isn't intended to work for archived courses that may be part of a program and/or pathway (which can also be highlighted by an admin)?

Related to this, I might recommend renaming `courseContent` variable to something else given its value is not always a course (can either be a course, program, or pathway).

src/components/EnterpriseApp/data/hooks/useEnterpriseCuration.js
katrinan029: good catch. I removed the error handling here since the error should be caught in the catch block instead

src/components/ContentHighlights/ContentHighlightArchivedCoursesAlert.jsx
```diff
@@ -0,0 +1,53 @@
...
+    // https://developer.chrome.com/blog/cookie-max-age-expires
+    const cookieExpiration = currentDate.add(400, 'day').format();
```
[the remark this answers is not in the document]
katrinan029: Thanks for the suggestion. I implemented the `localStorage` approach instead of cookies

src/components/ContentHighlights/data/constants.js
```diff
@@ -179,6 +180,9 @@ export const TEST_HIGHLIGHT_SET = {
   ],
 };
 
+export const NEW_ARCHIVED_COURSE_ALERT_DISMISSED_COOKIE_NAME = 'new-archived-course-alert-dismissed';
+export const archivedHighlightsCoursesCookies = new Cookies();
```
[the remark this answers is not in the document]
katrinan029: I removed the use of cookies and went with `localStorage`

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -43,18 +48,59 @@ export const enterpriseCurationActions = {
...
+          isNewArchivedCourse = true;
+        }
+      });
+    } else {
+      highlightSet.highlightedContent.forEach(courseContent => {
```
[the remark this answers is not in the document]
katrinan029: added the comment!

src/components/EnterpriseApp/data/hooks/useEnterpriseCuration.js
```diff
@@ -73,6 +85,11 @@ function useEnterpriseCuration({ enterpriseId, curationTitleForCreation }) {
         if (!curation) {
           curation = await createEnterpriseCuration();
         }
+        const curationHighlightedContents = await Promise.all(curation.highlightSets.map(async (highlightSet) => {
```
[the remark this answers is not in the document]
katrinan029: Yes, we are including the draft highlight sets as well.

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
katrinan029: In this scenario, we wouldn't want to consider the content as archived because an "unpublished" status could have an upcoming scheduled course run.

katrinan029: This is a good callout. We are intending for this to work with pathways and programs as well. @kiram15 merged in a couple of PR's that added course run statuses to pathways and programs.

https://github.com/openedx/course-discovery/pull/4248
https://github.com/openedx/course-discovery/pull/4261

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -54,12 +71,60 @@ function getHighlightSetsFromState(state) {
...
+          content.courseRunStatuses?.length === 1
+          && content.courseRunStatuses?.includes(COURSE_RUN_STATUSES.archived)
+          && !dismissedCookies.includes(content.contentKey)
+        ) {
+          isNewArchivedContent = true;
+        }
+      });
+    }
+  });
+  return isNewArchivedContent;
```
marlonkeating, line 117: nit: We could condense this logic
```suggestion
    const hasNewArchivedContent = (content) => {
      return content.courseRunStatuses?.length === 1
          && content.courseRunStatuses?.includes(COURSE_RUN_STATUSES.archived)
          && (!dismissedCookies || !dismissedCookies.includes(content.contentKey));
    };
  return highlightSet.highlightedContent.some(hasNewArchivedContent);
```

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -107,6 +172,19 @@ function enterpriseCurationReducer(state, action) {
...
+        ...state,
+        isNewArchivedContent,
+      };
+    }
+    case UPDATE_DISMISSED_ARCHIVED_COURSE: {
+      return {
+        ...state,
+        isNewArchivedContent: action.payload,
+      };
+    }
```
marlonkeating, line 187: I notice these are both triggered [here](https://github.com/openedx/frontend-app-admin-portal/pull/1160/files#diff-353876612b187c97b16c79ddd5eac4b45d88d8df513bb99096a327b23dd8aa06R40) upon update of `enterpriseHighlightedSets`, in a way it would seem they would overwrite each other.  Am I missing something here?

src/components/ContentHighlights/ContentHighlightArchivedContentAlert.jsx
```diff
@@ -0,0 +1,49 @@
...
+import { enterpriseCurationActions } from '../EnterpriseApp/data/enterpriseCurationReducer';
+
+const ContentHighlightArchivedContentAlert = () => {
+  const archivedContentCookies = {};
+  const [isOpen, setIsOpen] = useState(true);
+
+  const { enterpriseCuration: { enterpriseHighlightedSets, dispatch } } = useContext(EnterpriseAppContext);
+  enterpriseHighlightedSets.forEach(set => {
+    set.highlightedContent.forEach(content => {
+      if (content.courseRunStatuses?.length === 1 && content.courseRunStatuses.includes(COURSE_RUN_STATUSES.archived)) {
```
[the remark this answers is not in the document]
katrinan029: Yes, all archived content will only have a single course run status.

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
katrinan029: Only the `setIsNewArchivedContent` is triggered in the file you linked. The `updateDismissedArchivedCourse` is triggered when the alert is dismissed in the `ContentHighlightArchivedAlert` component.

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
katrinan029: thanks for this suggestion. I did some refactoring of this code and for readability I kept the `if-else` statement

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -54,12 +71,66 @@ function getHighlightSetsFromState(state) {
   return state.enterpriseCuration?.highlightSets || [];
 }
 
+/**
+ * Helper function to determine if a content is archived.
+ * If the length > 1 for the course run status e.g. ["archived", "published"]
+ * the content is not considered archived. An "unpublished" course run is not considered "archived" because
+ * it could have a scheduled course run in the future.
```
adamstankiewicz, line 78: [clarification] I suppose the reason I'm concerned about `unpublished` is that a course that might have `courseRunStatuses:  ["archived", "unpublished"]`, the unpublished course run is effectively unusable by learners (hidden). The only _actionable_ course run status in the `courseRunStatuses` array remaining that's relevant / displayed to learners would be `archived`, despite having a `courseRunStatuses` length of 2.

With the logic, as is, in `isArchivedContent`, we would consider the course to be non-archived despite the only available course run displayed to learners being archived.

src/components/ContentHighlights/ContentHighlightArchivedAlert.jsx
```diff
@@ -0,0 +1,49 @@
+import { Alert } from '@edx/paragon';
+import { useContext, useState } from 'react';
+import { NEW_ARCHIVED_CONTENT_ALERT_DISMISSED_COOKIE_NAME } from './data/constants';
+import { EnterpriseAppContext } from '../EnterpriseApp/EnterpriseAppContextProvider';
+import { enterpriseCurationActions, isArchivedContent } from '../EnterpriseApp/data/enterpriseCurationReducer';
+
+const ContentHighlightArchivedAlert = () => {
+  const archivedContentCookies = {};
+  const [isOpen, setIsOpen] = useState(true);
```
adamstankiewicz, line 9: [suggestion, sanity-check] I might recommend treating this `Alert` as a "controlled" component, whereby it accepts an `open` boolean prop to pass to the `Alert` component and instead defines the `const [isArchivedAlertOpen, setIsArchivedAlertOpen] = useState(false);` in `CurrentContentHighlightHeader`. I could imagine seeing a `useEffect` in `CurrentContentHighlightHeader` that executes when `isNewArchivedContent` changes to adjust the `isArchivedAlertOpen` state variable accordingly.

The primary rationale for this is because, as is, the `Alert` component is already open when its rendered in the DOM for the first time, whereby the fade-in transition/animation wouldn't work as expected. Typically, the `Alert` component should still be rendered, even if not opened initially.

The closing of the `Alert` could be done by passing in an `onClose` prop where a callback function is passed from the parent component, with a state update to `isArchivedAlertOpen`.

src/components/ContentHighlights/ContentHighlightArchivedAlert.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const { enterpriseCuration: { enterpriseHighlightedSets, dispatch } } = useContext(EnterpriseAppContext);
+  enterpriseHighlightedSets.forEach(set => {
```
adamstankiewicz, line 12: nit: perhaps `highlightSet` to be a tad more explicit; was initially confused this `set` was related to the JavaScript primitive [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

marlonkeating (approved):

src/components/ContentHighlights/ContentHighlightArchivedAlert.jsx
```diff
@@ -0,0 +1,49 @@
...
+    set.highlightedContent.forEach(content => {
+      if (isArchivedContent(content)) {
+        if (archivedContentCookies[set.uuid]) {
+          archivedContentCookies[set.uuid].push(content.contentKey);
+        } else {
+          archivedContentCookies[set.uuid] = [content.contentKey];
+        }
+      }
+    });
+  });
```
adamstankiewicz, line 22: [nit/suggestion] I believe this logic will be executed on every re-render, which could be a bit expensive given it's iterating through multiple `forEach` loops.

Would it make sense to ensure this logic is only executed when necessary (when data changes) vs. on every re-render?

```jsx
const [archivedContentCookies, setArchivedContentCookies] = useState({});

const { enterpriseCuration: { enterpriseHighlightedSets, dispatch } } = useContext(EnterpriseAppContext);

useEffect(() => {
    enterpriseHighlightedSets.forEach((highlightSet) => {
      highlightSet.highlightedContent.forEach((content) => {
        if (!isArchivedContent(content)) {
          // content not archived; do nothing.
          return;
        }
        // content is archived, so update `archivedContentCookies` state.
        setArchivedContentCookies((prevState) => {
          if (!prevState[highlightSet.uuid]) {
            prevState[highlightSet.uuid] = [content.contentKey];
          } else {
            prevState[highlightSet.uuid].push(content.contentKey);
          }
          return prevState;
        });
      });
    });
}, [enterpriseHighlightedSets]);
```

src/components/ContentHighlights/CurrentContentHighlightHeader.jsx
```diff
@@ -80,6 +83,7 @@ const CurrentContentHighlightHeader = ({ enterpriseId }) => {
       <p>
         {HEADER_TEXT.SUB_TEXT.currentContent}
       </p>
+      {isNewArchivedContent ? <ContentHighlightArchivedAlert /> : null }
```
adamstankiewicz, line 86: See related comment about conditionally rendering `ContentHighlightArchivedAlert` (as is) vs. controlling whether the `Alert` is visible or not via an `open` prop passed here.

Proposed:

```jsx
<ContentHighlightArchivedAlert open={isNewArchivedContent}>
```

src/components/ContentHighlights/ContentHighlightArchivedAlert.jsx
```diff
@@ -0,0 +1,49 @@
...
+    });
+    dispatch(enterpriseCurationActions.updateDismissedArchivedCourse(false));
+    setIsOpen(false);
+  };
+
+  return (
+    <Alert
+      variant="info"
+      dismissible
+      closeLabel="Dismiss"
```
adamstankiewicz, line 39: This is the default string for `Alert`; should be able to remove the prop if opting to use the default?

src/components/ContentHighlights/ContentHighlightArchivedAlert.jsx
```diff
@@ -0,0 +1,49 @@
+import { Alert } from '@edx/paragon';
+import { useContext, useState } from 'react';
+import { NEW_ARCHIVED_CONTENT_ALERT_DISMISSED_COOKIE_NAME } from './data/constants';
+import { EnterpriseAppContext } from '../EnterpriseApp/EnterpriseAppContextProvider';
+import { enterpriseCurationActions, isArchivedContent } from '../EnterpriseApp/data/enterpriseCurationReducer';
+
+const ContentHighlightArchivedAlert = () => {
+  const archivedContentCookies = {};
```
adamstankiewicz, line 8: nit: should this variable name be updated to reflect localStorage instead of cookies?

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
```diff
@@ -54,12 +71,77 @@ function getHighlightSetsFromState(state) {
...
 }
 
+/**
+ * Helper function to determine if a content is archived.
+ *
+ * @param {Object} content
+ * @returns {Boolean}
+ */
+export function isArchivedContent(content) {
+  const hasArchivedCourseRun = content.courseRunStatuses?.includes(COURSE_RUN_STATUSES.archived);
```
adamstankiewicz, line 81: nit: might be an opportunity for an early return if `courseRunStatuses` isnt always set?

```jsx
const { courseRunStatuses } = content;
if (!courseRunStatuses) {
  return false;
}
```

Then, you shouldn't need to include the `?.` when accessing the `courseRunStatuses` variable below.

---

I'm also wondering if the below logic could be a bit more concise but still have the same semantic meaning? Perhaps along the lines of the following, where the check is to see if _all_ elements are either `archived` or `published`:

```jsx
const ARCHIVABLE_STATUSES = [COURSE_RUN_STATUSES.archived, COURSE_RUN_STATUSES.unpublished];
return courseRunStatuses.every(status => ARCHIVABLE_STATUSES.includes(status));
```

Examples:

```jsx
["archived"] // true
["unpublished"] // true
["archived", "unpublished"] // true
["archived", "published", "unpublished] // false
```

adamstankiewicz (approved):
Looks good! I just left one suggestion for a proposed refactor to make logic a bit more concise in the function that determines whether the provided course run statuses are deemed "archived".

src/components/EnterpriseApp/data/enterpriseCurationReducer.js
katrinan029: thanks for this suggestion!

PR #1179, opened 2024-03-05
feat: create zero state for bnr

kiram15:
Creating a zero state in the activity tab for Browse and Request courses. 
[Jira ticket ](https://2u-internal.atlassian.net/browse/ENT-8499)

Screenshot 2024-03-05 at 10 46 03 AM

# For all changes


src/components/learner-credit-management/BudgetDetailActivityTabContents.jsx
```diff
@@ -32,19 +33,22 @@ const BudgetDetailActivityTabContents = ({ enterpriseUUID, enterpriseFeatures })
     );
   }
 
+  const hasSpentTransactions = !!budgetActivityOverview.spentTransactions?.count > 0;
```
marlonkeating, line 36: nit: `!!` isn't really needed here.  Because of javascript operator precedence, this would be converting the count to a boolean, with the end result being `true > 0` (hich still works because javascript would then coerce that to `1 > 0`)  
```suggestion
  const hasSpentTransactions = budgetActivityOverview.spentTransactions?.count > 0;
```
Which would result in a `undefined > 0` check in the event there is no `spentTransactions.count`, but that's fine since `undefined` gets coerced to a `0` in arithmetic operations.

marlonkeating (approved):

PR #1184, opened 2024-03-11
[ENT-8491] Implement Base Link Learners modal

kiram15:
This is the first pass at the invite learner modal for groups. This work includes the validation of the emails (duplicates allowed at this point but warned against), misformatted should surface an error and block. 

[Jira ticket 
](https://2u-internal.atlassian.net/browse/ENT-8491)

https://github.com/openedx/frontend-app-admin-portal/assets/31229189/5e42af8c-02b6-4476-a321-115e0ee2950d

# For all changes


marlonkeating (changes requested):

src/components/learner-credit-management/invite-modal/InviteMembersModalWrapper.jsx
```diff
@@ -0,0 +1,121 @@
...
+              showLaunchIcon
+              target="_blank"
+            >
+              Help Center: Invite Budget Members
+            </Button>
+            <ActionRow.Spacer />
+            <Button variant="tertiary" onClick={handleCloseInviteModal}>Cancel</Button>
+            <StatefulButton
+              labels={{
+                default: 'Invite',
```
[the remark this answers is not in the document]
kiram15: I think personally it would make it less readable, and I wanted to follow the pattern set from this https://github.com/openedx/frontend-app-admin-portal/blob/135cb6ec381b0a36db7b19d214030aaa4dfba4b3/src/components/learner-credit-management/cards/NewAssignmentModalButton.jsx#L241

src/components/learner-credit-management/BudgetDetailActivityTabContents.jsx
```diff
@@ -39,7 +43,10 @@ const BudgetDetailActivityTabContents = ({ enterpriseUUID, enterpriseFeatures })
   if (!isTopDownAssignmentEnabled || !subsidyAccessPolicy?.isAssignable) {
     return (
       <>
-        {!hasSpentTransactions && (<NoBnEBudgetActivity />)}
```
adamstankiewicz, line 42: [inform] Note: I believe you're rendering `NoBnEBudgetActivity` in certain cases where it should not be rendered... as is, `NoBnEBudgetActivity` is rendered / visible in production for Enterprise Offer budgets that have no spend.

As a result, the `NoBnEBudgetActivity` component (non-functional, without modal) _may_ be visible to admins as is today.

Keep in mind this page is intended to be backwards compatible with legacy Enterprise Offers; not everything here deals with Learner Credit per se. In the case of an enterprise offer used as the specific displayed budget, we should _only_ be rendering the `BudgetDetailRedemptions` component ("Spent" table) in this UI for Learner Credit compatible budgets (_not_ legacy enterprise offers).

As is, I believe `NoBnEBudgetActivity` is rendered for production customers who have an auto-applied budget but no spend (e.g., a newly onboarded customer to an auto-applied budget).

Is the intent for this work to be behind a feature flag until it's complete / fully integrated, so it doesn't expose partially functional components in the UI to production customers?

adamstankiewicz: [update] Discussed offline in Slack. The gating of `NoBnEBudgetActivity` behind feature flag will be split out to separate PR to more immediately resolve the production issue where this WIP component is exposed to customers for auto-applied budgets (Learner Credit, or legacy Enterprise Offers) without any spent transactions/redemptions.

src/components/learner-credit-management/BudgetDetailActivityTabContents.jsx
```diff
@@ -39,7 +43,10 @@ const BudgetDetailActivityTabContents = ({ enterpriseUUID, enterpriseFeatures })
...
+        {!hasSpentTransactions && isEnterpriseGroupsEnabled && (
```
adamstankiewicz, line 46: When the feature flag is enabled, the `NoBnEBudgetActivity` component will still render for legacy Enterprise Offers with no spend. Groups UI should only related to new Learner Credit, not legacy Enterprise Offers).

src/components/learner-credit-management/invite-modal/InviteMembersModalWrapper.jsx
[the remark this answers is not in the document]
adamstankiewicz: Also, worth nothing that if/when i18n is introduced into this MFE, any English strings stored in constants will need to be replaced with `intl.formatMessage` or `FormattedMessage`, etc.

src/components/learner-credit-management/invite-modal/InviteMembersModalWrapper.jsx
```diff
@@ -0,0 +1,124 @@
...
+        onClose={() => {
+          handleCloseInviteModal();
+        }}
+        footerNode={(
+          <ActionRow>
+            <Button
+              variant="tertiary"
+              as={Hyperlink}
+              destination={getConfig().ENTERPRISE_SUPPORT_URL}
+              showLaunchIcon
```
adamstankiewicz, line 83: nit: `showLaunchIcon` is `true` by default when `target="_blank"`

src/components/learner-credit-management/invite-modal/InviteModalSummary.jsx
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const InviteModalSummary = ({
+  memberInviteMetadata,
+}) => {
+  const {
+    isValidInput,
+    lowerCasedEmails,
+  } = memberInviteMetadata;
+  const hasLearnerEmails = lowerCasedEmails?.length > 0 && isValidInput;
```
adamstankiewicz, line 35: nit: Is `lowerCasedEmails` ever _not_ an array? It appears it's created as an empty array, which seems to imply it would always exist, and not need the `?.`.

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -442,6 +443,40 @@ describe('<BudgetDetailPage />', () => {
     }
   });
 
+  it('renders bne zero state only when the groups feature flag enabled', async () => {
```
adamstankiewicz, line 446: nit: might be worth negating the test title to indicate more clearly what this is testing; that the bne zero state is NOT rendered with groups feature flag _disabled_, and perhaps also including an assertion for what we _do_ expect to be rendered in the UI for this case.

src/data/services/EnterpriseCatalogApiService.js
```diff
@@ -31,6 +31,10 @@ class EnterpriseCatalogApiService {
     ).get(url);
   }
 
+  static fetchEnterpriseCatalog(catalogUuid) {
+    return EnterpriseCatalogApiService.apiClient().get(`${EnterpriseCatalogApiService.enterpriseCustomerCatalogsUrl}${catalogUuid}`);
```
adamstankiewicz, line 35: [clarification] Is this function intended to be used by anything? Doesn't appear to be used right now.

src/components/learner-credit-management/invite-modal/InviteMembersModalWrapper.jsx
```diff
@@ -0,0 +1,124 @@
...
+  }, []);
+
+  const handleInviteMembers = async () => {
+    setInviteButtonState('pending');
+
+    const payload = snakeCaseObject({
+      learnerEmails: learnerEmails.join(','),
+    });
+    const formData = new FormData();
+    Object.keys(payload).forEach(key => formData.append(key, payload[key]));
```
marlonkeating, line 49: Can use [snakeCaseFormData](https://github.com/openedx/frontend-app-admin-portal/blob/135cb6ec381b0a36db7b19d214030aaa4dfba4b3/src/utils.js#L180) for this

marlonkeating: like so: 
```suggestion
    const formData = snakeCaseFormData({
      learnerEmails: learnerEmails.join(','),
    });
```

marlonkeating: nit: Per discussions offline, it looks like [snakeCaseFormData](https://github.com/openedx/frontend-app-admin-portal/blob/135cb6ec381b0a36db7b19d214030aaa4dfba4b3/src/utils.js#L180) only supports FormData, not plain objects, so in this case I'd create a helper method for converting plain objects to FormData, for something along the lines of
```suggestion
    const formData = toFormData(snakeCaseObject({
      learnerEmails: learnerEmails.join(','),
    }));
```

src/utils.js
```diff
@@ -186,6 +187,14 @@ const snakeCaseFormData = (formData) => {
   return transformedData;
 };
 
+const snakeCaseObjectToForm = (payload) => {
+  // transforms an object to a snake_cased FormData object
+  snakeCaseObject(payload);
+  const formData = new FormData();
+  Object.keys(payload).forEach(key => formData.append(key, payload[key]));
```
marlonkeating, line 194: nit: `snakeCaseObject` mutates the object passed in, which isn't part of the explicit contract for `snakeCaseObjectToForm`.  I would at least make a note that `payload` will be altered, or perform the operations on a copy of the object to leave the original untouched. 
```suggestion
  const snakeCaseData = snakeCaseObject({...payload});
  const formData = new FormData();
  Object.keys(snakeCaseData).forEach(key => formData.append(key, payload[key]));
```

marlonkeating (approved):

PR #1186, opened 2024-03-19
fix: formdata email parsing error fix

kiram15:
Jira issue here - https://2u-internal.atlassian.net/browse/ENT-8657


marlonkeating (approved):
Changes look fine, though it would be good to add unit testing that could detect similar issues in the future.  If that's too heavy a lift given the speed we want to get this fix out, that's fine, but we should at least note this for future Tech Debt work.

PR #1191, opened 2024-03-29
fix: hiding highlights for custom courses

kiram15:
When we have groups feature turned on and they have a group that doesn't apply to all contexts, we do not want to show the highlights feature in the sidebar.

[Jira ticket ](https://2u-internal.atlassian.net/browse/ENT-8694)

### DIFF
--- a/src/components/Sidebar/index.jsx
+++ b/src/components/Sidebar/index.jsx
@@ -1,5 +1,5 @@
 import React, {
-  useRef, useContext, useEffect, useCallback,
+  useCallback, useContext, useEffect, useRef, useState,
 } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
@@ -8,6 +8,7 @@ import {
   BookOpen, CreditCard, Description, InsertChartOutlined, MoneyOutline, Settings, Support, Tag, TrendingUp,
 } from '@edx/paragon/icons';
 
+import { logError } from '@edx/frontend-platform/logging';
 import { getConfig } from '@edx/frontend-platform/config';
 import IconLink from './IconLink';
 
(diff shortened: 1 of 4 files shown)


src/containers/Sidebar/Sidebar.test.jsx
```diff
@@ -421,6 +429,35 @@ describe('<Sidebar />', () => {
...
+      ...initialState,
+      portalConfiguration: {
+        enterpriseFeatures: {
+          enterpriseGroupsV1: true,
+        },
+      },
+    });
+
+    LmsApiService.fetchEnterpriseGroup.mockImplementation(() => Promise.resolve({
+      data: { results: { applies_to_all_contexts: false } },
```
marlonkeating, line 444: From [how you're using the results of fetchEnterpriseGroup](https://github.com/openedx/frontend-app-admin-portal/pull/1191/files#diff-f6bf4f6edd2dac680925dc1421e37fa202dad8b65415291bded6349d4c679ec5R67), should be returning a group array here.
```suggestion
      data: { results: [{ applies_to_all_contexts: false }] },
```

src/components/Sidebar/index.jsx
```diff
@@ -201,6 +224,7 @@ Sidebar.propTypes = {
   enableAnalyticsScreen: PropTypes.bool,
   onWidthChange: PropTypes.func,
   isMobile: PropTypes.bool,
+  enterpriseGroupsV1: PropTypes.bool,
```
marlonkeating, line 227: nit: Would help readability to have a name that's a bit more explicit that it is a feature flag, IMO

marlonkeating (changes requested):

src/components/Sidebar/index.jsx
kiram15: I think it follows the same naming pattern as the other features, and you can see thats its being brought in from `enterpriseFeatures`

marlonkeating (approved):

PR #1192, opened 2024-03-29
feat: replaced packages @edx/paragon & @edx/frontend-build to use openedx namespace

BilalQamar95:
**Description**
- Replaced packages `@edx/paragon` & `@edx/frontend-build` to use `openedx` namespace
- Major version upgrades, Updating `frontend-build` to `v14` & `frontend-platform` to `v8` along with respective edx packages
- Updated `ts-jest` to `v29` along with respective packages
- Fixed failing tests post upgrade by making necessary upgrades


src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
```diff
@@ -922,20 +924,22 @@ describe('<BudgetDetailPage />', () => {
...
-          );
-        });
-      }
+      act(async () => {
+        const filtersDropdown = await screen.findByRole('group', { name: 'Status' });
+        const filtersDropdownContainer = within(filtersDropdown);
+        if (value.includes('waiting')) {
+          const waitingForLearnerOption = filtersDropdownContainer.getByLabelText('Waiting for learner 1', { exact: false });
+          expect(waitingForLearnerOption).toBeInTheDocument();
+          userEvent.click(waitingForLearnerOption);
```
adamstankiewicz, line 933: `userEvent.click`  should not be used within an `act`. See https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily for more details.

package.json
```diff
@@ -105,11 +104,15 @@
...
     "postcss": "8.4.24",
     "react-dev-utils": "11.0.4",
     "react-test-renderer": "^17.0.2",
     "resize-observer-polyfill": "1.5.1",
     "ts-jest": "^26.5.0"
+  },
+  "overrides": {
+    "@edx/frontend-platform": {
+      "@openedx/frontend-build": "13.1.0-alpha.2"
+    }
```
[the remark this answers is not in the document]
abdullahwaheed: i don't this we should add alpha as peer dependency since alpha is supposed to be a pre-release. We should keep peer dependencies for actual releases

src/components/learner-credit-management/tests/BudgetDetailPage.test.jsx
BilalQamar95: Since tests were failing without it `act` was used initially to ensure that all updates to the component (like state changes and DOM updates) were flushed and applied before assertions without these test was failing. However you are right, I have replaced it with `waitFor` which automatically handles the flushing of updates within its callback, waiting for the expected condition to be true which seems to have resolved the issue.

src/components/settings/SettingsLMSTab/LMSSelectorPage.tsx
```diff
@@ -38,6 +38,7 @@ export function LMSSelectorPage(setLms: (string) => void) {
             onChange={handleChange}
             name="colors"
             columns={3}
+            ariaLabelledby="lms-selection-heading"
```
adamstankiewicz, line 41: The `<p>` above with the text "Select the LMS or LXP you want to integrate..." is a more useful label for the `SelectableBox.Set` than "Let's get started"... 

"Let's get started" doesn't really describe the choices in the `SelectableBox.Set`.

package.json
```diff
@@ -91,7 +91,7 @@
   },
   "devDependencies": {
     "@edx/browserslist-config": "1.0.0",
-    "@edx/frontend-build": "^12.9.0-alpha.1",
+    "@openedx/frontend-build": "13.1.0-alpha.2",
```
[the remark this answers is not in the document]
BilalQamar95: This was intentional as v14 contains `jest` v29 upgrade & this PR focuses on `@edx/paragon` & `@edx/frontend-build` to `openedx` namespace switch. There is another [PR (currently drafted)](https://github.com/openedx/frontend-app-admin-portal/pull/1202/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R94) for `jest` v29 upgrade, containing upgrade to `frontend-build` v14 along with respective `edx` packages, which will be updated and ready for review once this PR is merged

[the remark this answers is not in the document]
adamstankiewicz: I'm not really following why this work needs to be split into 2 separate PRs at this point. Why can't the upgrades from the [draft PR](https://github.com/openedx/frontend-app-admin-portal/pull/1202/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R94) just be migrated to this PR since they are so related anyways? IMHO, I would have considered Jest v29 upgrade (and its other Jest-related package upgrades) as part of the original scope of migrating to latest version of `@openedx/frontend-build` in this PR directly; personally don't really see why this work needs to be split across 2 PRs.

Either way, I'm not going to push back further if you feel the incremental upgrade and separate upgrade is truly necessary versus bringing the other upgrades into this PR (given Jest upgrade, etc. actually _is_ related to the scope of migrating to `@openedx/frontend-build`, IMHO).

src/components/settings/SettingsLMSTab/LMSSelectorPage.tsx
adamstankiewicz: @BilalQamar95 Bump on this comment. Can we move the `id="lms-selection-heading"` to the `<p>` instead so it's a more useful/semantically correct message?

```
<p id="lms-selection-heading">Select the LMS or LXP you want to integrate with edX For Business.</p>
```

package.json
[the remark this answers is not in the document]
BilalQamar95: Thank you for your feedback. My rationale behind splitting the work into 2 different PRs was based on 
1. Scope and Complexity Management: The primary focus of the current PR is the namespace switch to `@openedx` for `@edx/paragon` and `@edx/frontend-build`. This was scoped as a standalone epic to ensure a clean transition without introducing additional variables. Also the `frontend-build` `v14` brings in several major package upgrades further adding to the complexity which when handled separately could ensure a more focused and manageable review process. 
2. Project Timeline and Dependencies: When the PR was initially created, `frontend-build `v14 was not yet available. It was intended to address the namespace switch only. The subsequent release of `frontend-build `v14 necessitated additional work, which was scoped into a separate Jest v29 epic
3. Given the significant number of file changes and potential issues arising from multiple major upgrades, splitting the work allows for easier tracking and isolation of issues, also incremental approach reduces the risk of introducing widespread breaking changes and simplifies troubleshooting if things go south.
4. The upgrades correspond to different epics, one focused on the namespace switch and the other on upgrading Jest and related packages. Keeping them separate maintains alignment with the objectives.

I understand the perspective that combining the upgrades might seem more efficient.  I have decided to combine the two considering the review process duration and the fact that the change itself, while widespread, is not overly complicated, it will ultimately expedite the merge process and reduce the overall time required for reviews by having single review for both features.

package.json
```diff
@@ -31,13 +31,13 @@
...
-    "@edx/frontend-enterprise-utils": "4.0.0",
-    "@edx/frontend-platform": "5.6.1",
+    "@edx/frontend-enterprise-catalog-search": "10.3.0",
+    "@edx/frontend-enterprise-hotjar": "7.1.0",
+    "@edx/frontend-enterprise-logistration": "9.1.0",
+    "@edx/frontend-enterprise-utils": "9.1.0",
+    "@edx/frontend-platform": "8.0.0",
     "@edx/openedx-atlas": "^0.6.0",
-    "@edx/paragon": "20.46.3",
+    "@openedx/paragon": "^21.13.1",
```
adamstankiewicz, line 40: nit: pin dependency

adamstankiewicz (approved):
LGTM (with tiny nit around pinning the Paragon dependency). Thanks for addressing the feedback, @BilalQamar95. Happy to help QA/verify on stage once merged.

BilalQamar95:
> LGTM (with tiny nit around pinning the Paragon dependency). Thanks for addressing the feedback, @BilalQamar95. Happy to help QA/verify on stage once merged.

Thank you for the approval & for QA/verification on stage. I have pinned the paragon and it will be merged once CI clears

PR #1214, opened 2024-05-07
fix: consider mixed state of budget expiries when rendering expiry modal

iloveagent57:
When determining budget expiry thresholds/modals, We need to consider the special case where there are a mix of expired and non-expired budgets.  In that case, we only want to determine the expiry threshold from the set of *non-expired* budgets, so that the expiry alert and modal below do not falsely indicate that *all* budgets are expired.

Note: this change leaves in place the current preference that we alert on the *earliest* expiring budget (now drawn from the set of only non-expired budgets).
ENT-8834

## Before change (what we want to fix)
image
image

## After change (the proposed fix)
image
image


brobro10000 (approved):
Some linting errors need to be fixed. LGTM 👍🏽

PR #1236, opened 2024-05-23
feat: retired workflow on budget details page.

zwidekalanga:
# For all changes

Screenshot 2024-06-03 at 1 10 14 AM

Screenshot 2024-06-11 at 1 42 17 PM

Screenshot 2024-06-03 at 1 11 22 AM


src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -112,16 +119,16 @@ const BudgetAssignmentsTable = ({
...
-      additionalColumns={[
+      additionalColumns={(!isRetired) ? [
         {
           Header: '',
           Cell: AssignmentRowActionTableCell,
           id: 'action',
         },
-      ]}
-      tableActions={[
+      ] : []}
```
brobro10000, line 128: [nit] This logic should be resolved outside of the component above the return statement

src/components/learner-credit-management/BudgetAssignmentsTable.jsx
```diff
@@ -112,16 +119,16 @@ const BudgetAssignmentsTable = ({
...
+      tableActions={(!isRetired) ? [
         <AssignmentsTableRefreshAction refresh={fetchTableData} />,
-      ]}
+      ] : []}
```
brobro10000, line 131: [nit/suggestion] similar to comment above, the logic should be resolved outside of the component. Since they are so similar you create a function that returns an object based on the value of `isRetired`

Pseudocode Ex.
```js
const budgetAssignmentsTableData = (isRetired) => {
if(!isRetired) {
return {
    tableActions: [],
    additionalColumns: []
}
return {
    tableActions: [
        <AssignmentsTableRefreshAction refresh={fetchTableData} />,
      ] ,
    additionalColumns: [{
          Header: '',
          Cell: AssignmentRowActionTableCell,
          id: 'action',
        }]
}
}
}

File: src/components/learner-credit-management/BudgetDetailAssignments.jsx [hunk 3]
```diff
@@ -57,32 +58,53 @@ const BudgetDetailAssignments = ({
...
-          id="lcm.budget.detail.page.assignments.heading"
-          defaultMessage="Assigned"
-          description="Heading for the assigned section on the budget detail page"
-        />
-      </h3>
-      <p className="small mb-4">
-        <FormattedMessage
-          id="lcm.budget.detail.page.assignments.description"
-          defaultMessage="Assigned activity earmarks funds in your budget so you can not overspend.
+      {isRetired ? (
```
brobro10000, line 61: [nit/suggestion] Might be worth splitting this out into its own component for a cleaner implementation

Example component name: `<BudgetDetailAssignmentsHeader/>`

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -94,6 +94,46 @@ const BudgetActions = ({
     );
   }
 
+  if (status === BUDGET_STATUSES.retired) {
```
brobro10000, line 97: [nit/suggestion] Similar suggestion as above. Could be worth splitting out into its own component.

src/components/learner-credit-management/data/hooks/useBudgetDetailHeaderData.js
```diff
@@ -30,6 +30,7 @@ const assignBudgetStatus = (policy) => {
     startDateStr: policy.subsidyActiveDatetime,
     endDateStr: policy.subsidyExpirationDatetime,
     isBudgetRetired: policy.retired,
+    retiredDateStr: '20 May 2024',
```
brobro10000, line 33: [nit] is this hard coded data necessary?

src/components/learner-credit-management/AssignmentStatusTableCell.jsx
```diff
@@ -71,6 +72,12 @@ const AssignmentStatusTableCell = ({ enterpriseId, row }) => {
   }
 
   // Display the appropriate status chip based on the learner state.
+  if (isRetired) {
```
adamstankiewicz, line 75: nit: this comment is not quite as relevant to `isRetired`, as `isRetired` does not relate to "learner state" per se. Maybe move this comment above the other existing `learnerState` conditionals, and have a separate comment specific to `isRetired` here, e.g. (or similar):

```jsx
// Always display "Incomplete assignment" status chip for retired budgets
```

src/components/learner-credit-management/BudgetDetailAssignments.jsx
```diff
@@ -57,32 +58,53 @@ const BudgetDetailAssignments = ({
...
+        <>
+          <h3 className="mb-3 text-info-900" ref={assignedHeadingRef}>
```
adamstankiewicz, line 63: [clarification] Where is the `text-info-900` coming from? I'm assuming you're finding an available Paragon color to try to mimic the changed text color in the Figma mock (i.e., the HEX value in Figma is not a valid Paragon color). 

However, note that this color change in Figma is not communicating that the text color itself should change, but rather just calling attention that, "hey, this particular messaging changed" using a lo-fi purple.

We should keep the color styles as they were out-of-the-box.

src/components/learner-credit-management/data/hooks/useBudgetDetailHeaderData.js
adamstankiewicz: @brobro10000 Good catch, I'm guessing this is hardcoded because the backend work to ensure the `retired_at` timestamp is not complete yet (https://2u-internal.atlassian.net/browse/ENT-8985), so the frontend has no date data to pull from as is.

This raises a question around sequencing, though; @zwidekalanga what is your intended plan around this PR and the related ENT-8985? Does ENT-8985 need to complete before this PR merges in, or will this gracefully handle null `retired_at` in the short term? That is, if this line changes to:

```jsx
retiredDateStr: policy.retiredAt,
```

src/components/learner-credit-management/data/hooks/useBudgetDetailTabs.jsx
```diff
@@ -59,7 +57,7 @@ export const useBudgetDetailTabs = ({
           }
           className={TAB_CLASS_NAME}
         >
-          {activeTabKey === BUDGET_DETAIL_CATALOG_TAB && (
+          {(activeTabKey === BUDGET_DETAIL_CATALOG_TAB && !isRetired) && (
```
adamstankiewicz, line 60: I would recommend continuing rendering the `CatalogTabElement` component when `activeTabKey === BUDGET_DETAIL_CATALOG_TAB`. 

`activeTabKey` just never will be equal to `BUDGET_DETAIL_CATALOG_TAB` due to the `disabled={isRetired}` above. The extra check for `!isRetired` isn't really needed at that point, as the component never actually gets rendered until the "Catalog" tab is selected (but can't be due to disabled state).

src/components/learner-credit-management/data/hooks/useIncompleteAssignments.jsx
```diff
@@ -0,0 +1,168 @@
+import {
```
adamstankiewicz, line 1: [clarification] What is the purpose of `useIncompleteAssignments`? It looks to be defined but never used, but largely copied from `useBudgetContentAssignments`. Is this a remnant of a previous implementation introducing a completely distinct/standalone `DataTable` component?

adamstankiewicz (changes requested):

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
brobro10000: Since this ticket is basically in a complete state, you can go ahead and make a new ticket to tackle that bit of work. Its important to make the code as readable as possible, and the long list of return statements with JSX components as each return makes this file difficult to parse. 
You can also add a // TODO: comment as a reminder to handle that optimization in the future. 👍🏽

src/components/learner-credit-management/BudgetDetailAssignments.jsx
```diff
@@ -10,6 +10,81 @@ import BudgetAssignmentsTable from './BudgetAssignmentsTable';
...
+const BudgetDetailAssignmentsHeader = ({
+  isRetired,
+}) => {
+  const assignedHeadingRef = useRef();
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const { state: locationState } = location;
+
+  useEffect(() => {
```
adamstankiewicz, line 22: The original `useEffect` for doing the `.scrollIntoView` for the "Assigned" table is still implemented in `BudgetDetailAssignments`, but is essentially a no-op because the `assignedHeadingRef` in `BudgetDetailAssignments` is not passed to an element anymore.

The new `assignedHeadingRef` within `BudgetDetailAssignmentsHeader` is the only ref still used.

I would recommend removing the `useEffect` that intends to do the same thing from `BudgetDetailAssignments` since it's not really doing anything anymore.

src/components/learner-credit-management/assignments-status-chips/IncompleteAssignment.jsx
```diff
@@ -0,0 +1,88 @@
...
+            <ul className="text-gray pl-3">
+              <li>View an active budget on the budgets page and reassign the course to the learner.</li>
+              <li>
+                Create a new budget by{' '}
+                <Hyperlink
+                  destination={supportUrl}
+                  onClick={helpCenterTrackEvent}
+                  target="_blank"
+                >
+                  contacting your support.
```
adamstankiewicz, line 63: nit: "contacting your support" doesn't really make sense.

```suggestion
                  contacting support.
```

src/components/learner-credit-management/data/hooks/useBudgetContentAssignments.js
```diff
@@ -68,31 +68,31 @@ const useBudgetContentAssignments = ({
...
-        page: args.pageIndex + 1, // `DataTable` uses zeo-indexed array
+        page: args.pageIndex + 1, // `DataTable` uses zero-indexed array
         pageSize: args.pageSize,
       };
       applyFiltersToOptions(args.filters, options);
       applySortByToOptions(args.sortBy, options);
 
-      /* This logic in conjunction with useRef is being used to prevent track events
-      from being called when the page re-renders without the specifically selected
-      arguments (argCopy) being changed */
```
adamstankiewicz, line 87: [nit/curious] What is the intent for removing this comment? Is it no longer relevant? If it is still relevant, seems like useful context to keep in place?

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -95,6 +95,46 @@ const BudgetActions = ({
...
+            <FormattedMessage
+              id="lcm.budget.detail.page.overview.budget.actions.keep.people.learning.active.budget"
+              defaultMessage="Keep people learning with an active budget"
+              description="Title for the budget actions section on the budget detail page overview"
+            />
+          </h3>
+          <p>
+            <FormattedMessage
+              id="lcm.budget.detail.page.overview.budget.actions.budget.retired"
+              defaultMessage="This budget is retired. To create a new budget, please contact your representative."
```
adamstankiewicz, line 112: nit: to be consistent with the other "representative" -> "support" changes...

```suggestion
              defaultMessage="This budget is retired. To create a new budget, please contact support."
```

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -95,6 +95,46 @@ const BudgetActions = ({
...
+              description="Description which tells that budget has retired and to create a new budget contact support"
```
adamstankiewicz, line 113: nit: 

```suggestion
              description="Description which tells that budget has retired and to create a new budget by contacting support"
```

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
adamstankiewicz, line 112: Per the [Figma](https://www.figma.com/design/857ACKyCYW9XfwTf8IvznE/%22Retired%22-state-of-budgets-(ENT---8207)?node-id=223-7353&m=dev), I believe this messaging and subsequent CTA is intended to be conditional based on whether there are any active budgets or not?

image

vs. 

image

"Active" budget here likely refers to any budget is not retired or expired (i.e., current or scheduled).

adamstankiewicz (changes requested):
This is looking good. I think there's still a couple discrepancies with what's depicted in the Figma, though.

1. The "Keep people learning with an active budget" sub-text and CTA has conditional messaging based on whether there are any active budgets available or not. I believe the current implementation in this PR is only taking into account the "no active budgets" case (see [Figma](https://www.figma.com/design/857ACKyCYW9XfwTf8IvznE/%22Retired%22-state-of-budgets-(ENT---8207)?node-id=223-7353&m=dev))? 
2. The "Utilization details" collapsible in the budget detail header (i.e., `BudgetDetailPageOverviewUtilization`) appears that it should NOT be displayed for retired budgets per the [Figma](https://www.figma.com/design/857ACKyCYW9XfwTf8IvznE/%22Retired%22-state-of-budgets-(ENT---8207)?node-id=223-6748&m=dev).
    * The annotation in Figma mentions retired budgets should have parity with expired budgets in this case. If we look at the [Figma](https://www.figma.com/design/gm4yVDt3TixtEi7aD6LETz/Learner-Credit-Expiration?node-id=489-23590&m=dev) for LC expiration, it has an annotation stating, "Utilization details dropdown is removed from the budget header as there are no longer any details to view as the budget is inactive".

adamstankiewicz (commented):
Three remaining items before approval:

1. Bug related to recognizing when on the budget detail page route (not detecting the nested tab routes). Relates to this PR insofar as we should not incorrectly display upcoming expiry messaging for retired budgets on the budget detail page route. Likely would also fix [ENT-9048](https://2u-internal.atlassian.net/browse/ENT-9048) as well.
2. The Figma discrepancy related to the "Utilization details" collapsible mentioned in the last round of feedback. The "Utilization details" section should not be displayed for retired budgets. See (2) in [this previous review comment](https://github.com/openedx/frontend-app-admin-portal/pull/1236#pullrequestreview-2104945842).
3. It appears there are 67 failing tests in CI.

src/config/index.js
```diff
@@ -17,6 +17,7 @@ const configuration = {
   ENTERPRISE_CATALOG_BASE_URL: process.env.ENTERPRISE_CATALOG_BASE_URL,
   ENTERPRISE_ACCESS_BASE_URL: process.env.ENTERPRISE_ACCESS_BASE_URL,
   ENTERPRISE_SUBSIDY_BASE_URL: process.env.ENTERPRISE_SUBSIDY_BASE_URL,
+  ENTERPRISE_SUPPORT_LEARNER_CREDIT_URL: process.env.ENTERPRISE_SUPPORT_LEARNER_CREDIT_URL || null,
```
adamstankiewicz, line 20: Re-posting the feedback from our Slack conversation for posterity:

> the `ENTERPRISE_SUPPORT_LEARNER_CREDIT_URL` config variable is already exposed via `getConfig()` from frontend-platform, but this is adding it to the legacy configuration map defined in this file. While not a huge deal, I'm a bit concerned about having two different config locations defined for `ENTERPRISE_SUPPORT_LEARNER_CREDIT_URL`. 
>
> All the config in `src/config/index.js` really should eventually be moved to the newer mechanism of doing config for the MFE (e.g., [here](https://github.com/openedx/frontend-app-admin-portal/blob/e9b406f86a863e332c906100d547a133061ba481/src/index.jsx#L42) and consumed [here](https://github.com/openedx/frontend-app-admin-portal/blob/e9b406f86a863e332c906100d547a133061ba481/src/components/learner-credit-management/assignments-status-chips/FailedSystem.jsx#L60)).

src/components/learner-credit-management/BudgetOverviewContent.jsx
```diff
@@ -86,12 +86,14 @@ const BudgetOverviewContent = ({
...
-          isAssignable={isAssignable}
-        />
+        {status !== BUDGET_STATUSES.retired && (
+          <BudgetDetailPageOverviewUtilization
+            budgetId={budgetId}
+            budgetTotalSummary={budgetTotalSummary}
+            budgetAggregates={budgetAggregates}
+            isAssignable={isAssignable}
+          />
+        )}
```
adamstankiewicz, line 96: Re-posting the feedback from our Slack conversation for posterity:

> Currently you're conditionally rendering `BudgetDetailPageOverviewUtilization` with a conditional in `BudgetOverviewContent`. However, inside `BudgetDetailPageOverviewUtilization` itself there are conditionals for when the component should not display (`return null`). I'm a bit hesitant about introducing two separate places containing conditional logic for determining when this component renders in the UI. I would recommend moving the `isRetired` check into `BudgetDetailPageOverviewUtilization` similar to `isAssignable` so all that logic is co-located and easier to reason about.

adamstankiewicz (commented):
Re-posted the remaining PR feedback shared in our Slack DM onto the PR itself for posterity. Once those 2 items are addressed, this looks good to go!

Also note it looks like there may be a merge conflict with `BudgetStatusSubtitle` that'll need to be resolved as well.

adamstankiewicz (approved):

PR #1250, opened 2024-06-10
fix: add app content wrapper padding based on page text direction(ltr or rtl)

muhammad-ammar:
**Description:** Add padding(left or right) baesd on the page text direction(ltr or rtl)

# For all changes

### DIFF
--- a/src/components/EnterpriseApp/index.jsx
+++ b/src/components/EnterpriseApp/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Navigate } from 'react-router-dom';
 import { breakpoints, MediaQuery } from '@openedx/paragon';
+import {getLocale, isRtl} from '@edx/frontend-platform/i18n';
 
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import Sidebar from '../../containers/Sidebar';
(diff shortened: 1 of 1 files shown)


src/components/EnterpriseApp/index.jsx
```diff
@@ -141,7 +143,7 @@ class EnterpriseApp extends React.Component {
                   tabIndex="-1"
                   ref={this.contentWrapperRef}
                   style={{
-                    paddingLeft: matchesMediaQ ? sidebarWidth : defaultContentPadding,
+                    [isLocaleRTL ? 'paddingRight' : 'paddingLeft']: matchesMediaQ ? sidebarWidth : defaultContentPadding,
```
adamstankiewicz, line 146: [suggestion] Rather than relying on `getLocale` and `isRtl` here, I would recommend a pure CSS solution with `paddingInlineStart` here instead, the primary reasons being:
1. It's a simpler implementation.
2. Can respond to manual changes of the `dir="ltr"` attribute on the `<html>` element, whereas this solution would not. For example, if you manually change the `<html>` element's `dir="ltr"` attribute in DevTools to `dir="rtl"`, the padding is not properly updated with the current approach.

The `padding-inline-start` CSS property ([docs](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start)) takes into account LTR vs. RTL directionality by default.

Example screenshot, after manually changing to `dir="rtl"` in DevTools and relying on `paddingInlineStart`:

image

PR #1262, opened 2024-07-03
fix: small subtitle fix

kiram15:
# For all changes

### DIFF
--- a/src/components/learner-credit-management/BudgetStatusSubtitle.jsx
+++ b/src/components/learner-credit-management/BudgetStatusSubtitle.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import isEmpty from 'lodash/isEmpty';
 
 import PropTypes from 'prop-types';
 import {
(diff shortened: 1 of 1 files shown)


src/components/learner-credit-management/BudgetStatusSubtitle.jsx
```diff
@@ -8,19 +9,15 @@ import { GroupAdd, Groups, ManageAccounts } from '@openedx/paragon/icons';
...
-  useEnterpriseCustomer,
-  useEnterpriseGroup,
-  isLmsBudget,
+  formatDate, useEnterpriseCustomer, useEnterpriseGroup, isLmsBudget,
 } from './data';
 
 const BudgetStatusSubtitle = ({
   badgeVariant, status, isAssignable, term, date, policy, enterpriseUUID, isRetired,
 }) => {
+  const customGroup = !isEmpty(policy?.groupAssociations);
```
bcitro, line 18: 1. Can we use `enterpriseGroup` from the next line for this check?
2. I think this var will also be true if there is a global group associated with the policy, so we may need to add another check here for the `appliesToAllContexts` value

src/components/learner-credit-management/BudgetStatusSubtitle.jsx
```diff
@@ -8,19 +9,15 @@ import { GroupAdd, Groups, ManageAccounts } from '@openedx/paragon/icons';
...
-  useEnterpriseGroup,
-  isLmsBudget,
+  formatDate, useEnterpriseCustomer, useEnterpriseGroup, isLmsBudget,
 } from './data';
 
 const BudgetStatusSubtitle = ({
   badgeVariant, status, isAssignable, term, date, policy, enterpriseUUID, isRetired,
 }) => {
   const { data: enterpriseGroup } = useEnterpriseGroup(policy);
+  const customGroup = !isEmpty(policy?.groupAssociations) && !enterpriseGroup?.appliesToAllContexts;
```
bcitro, line 19: nit: I think you could say `!isEmpty(enterpriseGroup) && !enterpriseGroup?.appliesToAllContexts` here instead if you wanted to

bcitro (approved):

PR #1272, opened 2024-08-12
fix: reference correct track event name

brobro10000:
Updates track event name used in segment events to an existing track event name. Issue derived from the following DD error. 
Screenshot 2024-08-12 at 9 38 19 AM

The event name has [always existed](https://github.com/openedx/frontend-app-admin-portal/blob/67f42b70c27d561d8c79f09f77654ed46f1ff449/src/eventTracking.js#L176) but was incorrectly referenced. 

# For all changes


macdiesel (approved):
🙇

PR #1279, opened 2024-08-21
feat: adds end timestamp to maintenance banner

brobro10000:
Adds `MAINTENANCE_ALERT_END_TIMESTAMP` as a way to preset an end timestamp for the maintenance banner. 

Effectively handles if an end of start date is not set.

Manually tested this feature locally with permutations including:
* `MAINTENANCE_ALERT_END_TIMESTAMP` and `MAINTENANCE_ALERT_START_TIMESTAMP` set
* `MAINTENANCE_ALERT_END_TIMESTAMP` and not `MAINTENANCE_ALERT_START_TIMESTAMP` set
*  Not `MAINTENANCE_ALERT_END_TIMESTAMP` and `MAINTENANCE_ALERT_START_TIMESTAMP` set

# For all changes


src/components/App/index.jsx
```diff
@@ -101,8 +102,9 @@ const AppWrapper = () => {
       return false;
     }
     const startTimestamp = config.MAINTENANCE_ALERT_START_TIMESTAMP;
-    if (startTimestamp) {
-      return new Date() > new Date(startTimestamp);
+    const stopTimestamp = config.MAINTENANCE_ALERT_STOP_TIMESTAMP;
+    if (startTimestamp && stopTimestamp) {
```
adamstankiewicz, line 106: [suggestion] I think this should still support having a `startTimestamp` with no configured `stopTimestamp`, for scenarios where we might not know the expected duration of a maintenance window. For example:

* Both `startTimestamp` and `stopTimestamp` are defined, then the check for whether today's date is between the two start/stop dates make sense.
* Only `startTimestamp` defined, the date check should only check whether today's date is after the start.
* Only `endTimestamp` defined, the date check should only check whether today's date is before the end.

src/components/App/index.jsx
```diff
@@ -101,8 +102,15 @@ const AppWrapper = () => {
       return false;
     }
     const startTimestamp = config.MAINTENANCE_ALERT_START_TIMESTAMP;
+    const endTimeStamp = config.MAINTENANCE_ALERT_STOP_TIMESTAMP;
```
adamstankiewicz, line 105: nit: similar casing for `Timestamp` as in `startTimestamp`.

nit x2: any concerns around using "end" here but "stop" within the config setting? I might recommend changing the config setting to `MAINTENANCE_ALERT_END_TIMESTAMP` and keeping `endTimestamp` 🤔 (using "end" is more parity with other end-related data fields in other configs/APIs/etc.)

adamstankiewicz (approved):
Just a couple minor nits (tl;dr; a suggestion to possibly rename to `MAINTENANCE_ALERT_END_TIMESTAMP` in the config setting)

PR #1292, opened 2024-09-03
feat: assign by course run implementation

brobro10000:
Ability to assign learner credit assignments by individual course runs. This is a hard cutover from assigning from the top level course key to the individual course run. 
Displays an enroll-by date for each applicable course run that currently verifies that the enroll-by date of the course run is before the `subsidyExpirationDate`- `MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS` to ensure course runs that are far in the future beyond the policies term are not displayed but respect the ability to provide refunds within the subsidy expiration date.

This PR also includes:
  - Updates display of prices to include a range if multiple unique prices are available for runs based on `content_price`
  - Updates course page url on the allocation table to the correct URL (courseRun key --> top level key)
  - Update heuristic that determines displaying start date for self paced courses based whether the user has time to complete the course or within 14 days of today. If not, default the `start_date` to today
  - Supports the late enrollment flow for displaying course runs by determining if the policy allows the user to display late enrollment runs.
  - Updates tests and sample algolia return to include `course_runs` to accurately test the changes. 

Catalog search results card.
Screenshot 2024-09-05 at 12 40 50 PM

Allocation screen:
Screenshot 2024-09-05 at 12 41 09 PM

Zero state was also added for edge cases that the enroll-by date for all course runs is beyond the `subsidyExpirationDate`+`STALE_ENROLLMENT_DROPOFF_DAYS` threshold
Catalog search zero state:
Screenshot 2024-09-05 at 12 49 20 PM

# For all changes


src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +553,59 @@ export const isLmsBudget = (
...
+ * If the start date of the course is before today offset by the START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS
+ * then return today's formatted date. Otherwise, pass-through the start date in ISO format.
+ *
+ * For cases where a start date does not exist, just return today's date.
+ *
+ * @param start
+ * @param format
+ * @returns {string}
+ */
+export const setStaleCourseStartDates = ({ start }) => {
```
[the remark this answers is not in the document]
adamstankiewicz: [update] From our discussion related in scrum, we do indeed want parity between admin+learner.

* learner portal's fallback to original start date should also consider the "old start date" threshold also getting introduced in admin portal.
* admin portal should only consider using today's date for the start date when the course is self-paced.
* admin portal should only consider using today's date for the start date when there's enough time to complete as determined by the existing/similar logic implemented [here](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/bd4087e1370d8c97b32888514c5fa1d0636f3d88/src/components/course/data/utils.jsx#L57-L59).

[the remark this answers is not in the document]
brobro10000: For total agreeement, this would be the goal of implementing the new `start` date threshold logic within the `hasTimeToComplete` conditional as an OR statement 
``` jsx
    if (isCourseSelfPaced(courseRun.pacingType)) {
      if (hasTimeToComplete(courseRun) || OLD_START_DATE_LOGIC) {
        // always today's date (incentives enrollment)
        return intl.formatMessage(messages.courseStartDate, {
          startDate: dayjs().format(DATE_FORMAT),
        });
      }
      return intl.formatMessage(messages.courseStartedDate, {
        startDate: dayjs(courseStartDate).format(DATE_FORMAT),
      });
    }
    return intl.formatMessage(messages.courseStartedDate, {
      startDate: dayjs(courseStartDate).format(DATE_FORMAT),
    });
```

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
```diff
@@ -0,0 +1,45 @@
...
+  SHORT_MONTH_DATE_FORMAT,
+} from '../data';
+
+const NewAssignmentModalDropdown = ({
+  id: courseKey, onClick: openAssignmentModal, courseRuns, children,
+}) => {
+  const [emptyState] = useState(courseRuns.length === 0);
+  return (
+    <Dropdown id={courseKey}>
+      <Dropdown.Toggle variant="primary" id="assign-by-course-runs-dropdown">
```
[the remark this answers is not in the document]
brobro10000: The `NewAssisgnmentModalButton` component passes a top level course `courseKey` value to each dropdown component (each card result). Logging the output of each id on the top level `<Dropdown />` (iamge 1) and `<Dropdown.Item />` componenet show that they are unique 👍🏽  
Screenshot 2024-09-11 at 3 13 56 PM
Screenshot 2024-09-11 at 3 14 38 PM

src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
```diff
@@ -54,6 +62,14 @@ const useCourseCardMetadata = ({
...
   }
 
+  const availableCourseRuns = assignableCourseRuns({
+    courseRuns,
+    subsidyExpirationDatetime: subsidyAccessPolicy.subsidyExpirationDatetime,
+    staleEnrollmentDropOffTime: STALE_ENROLLMENT_DROPOFF_DAYS,
+  });
+  const assignmentImportantDates = <AssignmentModalImportantDates courseRun={courseRun} />;
+  const availableCourseRunsCount = `(${availableCourseRuns.length}) available ${pluralText('date', availableCourseRuns.length)}`;
+  const dateFooterText = displayImportantDates ? assignmentImportantDates : availableCourseRunsCount;
```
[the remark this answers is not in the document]
brobro10000: Refactored some of the logic around here to deal with some typing errors for `enrollBy` expecting a `string` as required from `courseRuns`
Functionally the same, but great suggestion here 👍🏽

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +553,59 @@ export const isLmsBudget = (
...
+ * Determines if the course has already started. Mostly used around text formatting for tense
+ * Introduces 'jitter' in the form of a 30 second offset to take into account any additional
+ * formatting that takes place down stream related to setting values to today's date through dayjs()
+ *
+ * This should also help with reducing 'flaky' tests.
+ *
+ * @param start
+ * @returns {boolean}
+ */
+export const hasCourseStarted = (start) => dayjs(start).isBefore(dayjs().subtract(30, 'seconds'));
```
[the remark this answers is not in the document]
brobro10000: This was mostly added for downstream computations of checks around if the course has started. It was added for cases where once we validated the course has started, we would use the same date to determine which tense the associated text should display (past tense or present)

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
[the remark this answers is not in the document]
adamstankiewicz: What about `id="assign-by-course-runs-dropdown"`, though? Is the `id` on the top-level `Dropdown` necessary? On the [Paragon docs site](https://paragon-openedx.netlify.app/components/dropdown/#advanced-usage), it looks like it's only passed on the `Dropdown.Toggle`. Should the `courseKey` be included as part of this `id` to make it unique?

src/components/learner-credit-management/data/utils.js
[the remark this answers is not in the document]
brobro10000: Removed 👍🏽

src/components/learner-credit-management/BudgetDetailPageOverviewAvailability.jsx
```diff
@@ -18,10 +18,9 @@ import {
   useSubsidyAccessPolicy,
   useEnterpriseCustomer,
   useEnterpriseGroup,
-  isLmsBudget,
+  isLmsBudget, LEARNER_CREDIT_ROUTE,
```
adamstankiewicz, line 21: [tiniest of nits] import on newline

src/components/learner-credit-management/assignment-modal/AssignmentModalmportantDates.jsx
```diff
@@ -0,0 +1,95 @@
...
+  // This is an edge case that the user should never enter but covered nonetheless
+  if (!enrollByDate && !courseStartDate) {
+    return null;
+  }
+
+  return (
+    <section className="assignments-important-dates small">
+      <Stack direction="vertical" gap={1}>
+        {enrollByDate && (
+        <AssignmentModalImportantDate label={intl.formatMessage(messages.enrollByDate)}>
```
adamstankiewicz, line 74: nitty nit: could use extra indents for `AssignmentModalImportantDate` to visually look nested under its related conditional here and below.

src/components/learner-credit-management/assignment-modal/NewAssignmentModalButton.jsx
```diff
@@ -114,11 +133,11 @@ const NewAssignmentModalButton = ({ enterpriseId, course, children }) => {
       trackEventMetadata,
     );
   };
-
   const handleAllocateContentAssignments = () => {
+    // If no assignmentRun key exist, fall back to the top level course key
```
adamstankiewicz, line 137: [nit/clarification] Is this still a valid comment? It looks like we're always passing `assignmentRun.key` without a fallback (which I believe is intentional?).

src/components/learner-credit-management/cards/CourseCardFooterActions.jsx
```diff
@@ -27,32 +26,34 @@ const CourseCardFooterActions = ({ enterpriseId, course }) => {
...
-        defaultMessage="View course"
-        description="Button text to view course"
-      />
-    </Button>,
-    (!catalogGroupView ? (
-      <NewAssignmentModalButton key="assignment-modal-trigger" course={course}>
+  return (
+    <>
+      <Button
+        key="link-to-course"
```
adamstankiewicz, line 32: nit: technically i believe the `key` props here and below could be removed as these components are no longer rendered within an array like before.

src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
```diff
@@ -54,6 +71,15 @@ const useCourseCardMetadata = ({
     linkToCourse = `${ENTERPRISE_LEARNER_PORTAL_URL}/${enterpriseSlug}/executive-education-2u/course/${key}`;
   }
 
+  const assignableCourseRuns = getAssignableCourseRuns({
+    courseRuns,
+    subsidyExpirationDatetime: subsidyAccessPolicy.subsidyExpirationDatetime,
+  });
+  // const footerText = `(${assignableCourseRuns.length}) available ${pluralText('date', assignableCourseRuns.length)}`;
```
adamstankiewicz, line 78: nit: commented out line

src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
```diff
@@ -54,6 +71,15 @@ const useCourseCardMetadata = ({
...
+  const footerText = intl.formatMessage(messages.courseFooterMessage, {
+    courseRuns: assignableCourseRuns.length,
+    pluralText: pluralText('date', assignableCourseRuns.length),
+  });
```
adamstankiewicz, line 82: nice construction of the i18n message here!

src/components/learner-credit-management/data/constants.js
```diff
@@ -22,6 +22,12 @@ export const API_FIELDS_BY_TABLE_COLUMN_ACCESSOR = {
   courseListPrice: 'course_list_price',
 };
 
+// Course pace text
```
adamstankiewicz, line 25: Thank you for making sure we stick to a single `constants` file within this module 🙌

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +553,86 @@ export const isLmsBudget = (
...
+
+export const isCourseSelfPaced = ({ pacingType }) => pacingType === COURSE_PACING_MAP.SELF_PACED;
+
+export const hasTimeToComplete = ({ end, weeksToComplete }) => {
+  const today = dayjs();
+  const differenceInWeeks = dayjs(end).diff(today, 'week');
+  return weeksToComplete <= differenceInWeeks;
+};
+
+const isWithinMinimumStartDateThreshold = ({ start }) => dayjs(start).isBefore(dayjs().subtract(START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS, 'days'));
```
adamstankiewicz, line 599: 👨‍🍳 💋

src/index.scss
```diff
@@ -84,3 +84,9 @@ form {
 .font-size-base {
   font-size: $font-size-base !important;
 }
+
+// TODO: Class override should be a contribution to Paragon if not already
```
adamstankiewicz, line 88: nit: might be worth adding the paragon version your upstream fix was released for future reference (v22.8.1).

src/utils.js
```diff
@@ -546,6 +546,20 @@ function makePlural(num, string) {
...
 
+/**
+ * Pluralizes a word that typically ends with s based on the benchmark passed
+ *
+ * @param textToPlural
+ * @param pluralBenchmark
+ * @param punctuation
+ * @returns {string}
+ */
+const pluralText = (
```
adamstankiewicz, line 557: [inform] I think what you have here is fine, but just as a heads up, it looks like there is some support for defining plural messages within the message format (ICU syntax, see https://formatjs.io/docs/core-concepts/icu-syntax/#plural-format) or perhaps `FormattedPlural` ([docs](https://formatjs.io/docs/react-intl/components/#formattedplural)).

brobro10000: Leaving comment open as reference 👍🏽

src/components/learner-credit-management/data/constants.js
```diff
@@ -71,6 +98,15 @@ export const MEMBERS_TABLE_PAGE_SIZE = 10;
 // Enroll-by date warning message threshold by days
 export const ENROLL_BY_DATE_DAYS_THRESHOLD = 10;
 
+// Allocation assignment expiration dropoff threshold
+export const DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION = 90;
+
+// Maximum days allowed from enrollment for a refund on assignments related to policies
+export const MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS = 14;
+
+// Start date threshold to default to today days, sets start date to today if course start date is beyond this value
```
pwnage101, line 107: Nit:
```suggestion
// When the start date is before this number of days before today, display the alternate start date (fixed to today).
```

src/components/learner-credit-management/data/constants.js
```diff
@@ -85,3 +121,6 @@ export const learnerCreditManagementQueryKeys = {
   budgetGroupLearners: (budgetId) => [...learnerCreditManagementQueryKeys.budget(budgetId), 'group learners'],
   enterpriseCustomer: (enterpriseId) => [...learnerCreditManagementQueryKeys.all, 'enterpriseCustomer', enterpriseId],
 };
+
+// Route to learner credit
+export const LEARNER_CREDIT_ROUTE = '/:enterpriseSlug/admin/:enterpriseAppPage/:budgetId/:activeTabKey';
```
adamstankiewicz, line 126: [nit/sanity check] Is the `:activeTabKey` always present for this route?

```jsx
<Route
  path="/:budgetId/:activeTabKey?"
  element={<BudgetDetailPage />}
/>
```

Based on this, we might to define the route as having parity with the upstream route definition? That said, when we actually _use_ this constant, we are always passing the `activeTabKey` route param, so this is largely a moot point anyways.

Still might be good to have parity with the route definition, though?

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
   activeIntegrationsLength,
   isUniversalGroup,
 ) => activeIntegrationsLength > 0 && isUniversalGroup;
+
+/**
+ * Determines if the course has already started. Mostly used around text formatting for tense
+ *
+ * This should also help with reducing 'flaky' tests.
```
adamstankiewicz, line 562: nit: the flaky tests bit was related to the now-removed jitter, right? If so, this line of the comment docstring likely could be removed.

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
...
+ * @param subsidyExpirationDatetime
+ * @param isLateRedemptionAllowed
+ * @returns {*}
+ */
+export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime, isLateRedemptionAllowed }) => {
+  const clonedCourseRuns = courseRuns.map(a => ({
+    ...a,
+    enrollBy: dayjs.unix(a.enrollBy).toISOString(),
+    upgradeDeadline: dayjs.unix(a.upgradeDeadline).toISOString(),
+  }));
```
adamstankiewicz, line 584: nit: `a` -> `courseRun` or `run`

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
...
+ * @param courseRuns
+ * @param subsidyExpirationDatetime
+ * @param isLateRedemptionAllowed
+ * @returns {*}
+ */
+export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime, isLateRedemptionAllowed }) => {
+  const clonedCourseRuns = courseRuns.map(a => ({
+    ...a,
+    enrollBy: dayjs.unix(a.enrollBy).toISOString(),
+    upgradeDeadline: dayjs.unix(a.upgradeDeadline).toISOString(),
```
adamstankiewicz, line 583: [clarification] Is `upgradeDeadline` needed here if the related `assignableCourseRunsFilter` does not consume the `upgradeDeadline`?

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
...
+const isWithinMinimumStartDateThreshold = ({ start }) => dayjs(start).isBefore(dayjs().subtract(START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS, 'days'));
+
+/**
+ * If the start date of the course is before today offset by the START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS
+ * then return today's formatted date. Otherwise, pass-through the start date in ISO format.
+ *
+ * For cases where a start date does not exist, just return today's date.
+ *
+ * @param start
+ * @param format
```
adamstankiewicz, line 622: nit: these params don't match the actual args

src/components/learner-credit-management/data/utils.js
brobro10000: This change is part of a nice to have for future development where the current `upgradeDeadline` value is also a timestamp value.

src/components/learner-credit-management/assignment-modal/NewAssignmentModalButton.jsx
```diff
@@ -173,19 +199,22 @@ const NewAssignmentModalButton = ({ enterpriseId, course, children }) => {
           EVENT_NAMES.LEARNER_CREDIT_MANAGEMENT.ASSIGNMENT_ALLOCATION_ERROR,
           {
             ...sharedEnterpriseTrackEventMetadata,
+            contentKey: assignmentRun.key,
+            parentContentKey: course.key,
             totalAllocatedLearners: learnerEmails.length,
             errorStatus: httpErrorStatus,
-            errorReason,
+
```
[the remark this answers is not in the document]
brobro10000: Mistakenly removed during refactor 👍🏽

src/components/learner-credit-management/data/constants.js
```diff
@@ -44,6 +53,24 @@ export const BUDGET_DETAIL_TAB_LABELS = {
   [BUDGET_DETAIL_MEMBERS_TAB]: 'Members',
 };
 
+// Card text for used in useCourseCardMetadata
+export const CARD_TEXT = {
```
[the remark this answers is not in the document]
brobro10000: Left a TODO for now but will circle back if its still not addressed in future commits.

src/components/learner-credit-management/data/constants.js
brobro10000: Updated to keep in parity. Great callout 👍🏽

src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
```diff
@@ -22,18 +23,18 @@ const messages = defineMessages({
 });
 
 const getContentPriceDisplay = ({ courseRuns }) => {
-  const flatContentPrice = courseRuns.flatMap(run => run?.contentPrice);
+  const flatContentPrice = courseRuns.flatMap(run => run.contentPrice || EMPTY_CONTENT_PRICE_VALUE);
   // Find the max and min prices
+  if (!flatContentPrice.length) {
+    return formatPrice(0.0);
```
adamstankiewicz, line 29: nit: Should this be using the same `EMPTY_CONTENT_PRICE_VALUE` here as above?

adamstankiewicz (approved):

src/components/learner-credit-management/AssignmentDetailsTableCell.jsx
```diff
@@ -21,6 +21,7 @@ const AssignmentDetailsTableCell = ({ row, enterpriseSlug, enterpriseId }) => {
       state: row.original.state,
     },
   );
+  const contentKey = row.original.isAssignedCourseRun ? row.original.parentContentKey : row.original.contentKey;
```
pwnage101, line 24: Nit: call this variable a `courseKey` now that it is confirmed a course key.

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
...
+  if (!start) {
+    return todayToIso;
+  }
+  const startDateIso = dayjs(start).toISOString();
+  if (isCourseSelfPaced({ pacingType })) {
+    if (hasTimeToComplete({ end, weeksToComplete }) || isWithinMinimumStartDateThreshold({ start })) {
+      // always today's date (incentives enrollment)
+      return todayToIso;
+    }
+    return startDateIso;
```
pwnage101, line 638: Nit: useless line made redundant by the next return

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,98 @@ export const isLmsBudget = (
...
+};
+
+export const getNormalizedEnrollByDate = ({ enrollBy }) => {
+  if (!enrollBy) {
+    return null;
+  }
+  const ninetyDayAllocationOffset = dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days');
+  if (dayjs(enrollBy).isAfter(ninetyDayAllocationOffset)) {
+    return ninetyDayAllocationOffset.toISOString();
+  }
```
pwnage101, line 650: Nit: The variable is suffixed with "Offset" but it is not an offset, it is a date. Maybe "ninetyDaysFromNow" instead?

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,97 @@ export const isLmsBudget = (
...
+ *
+ * @param start
+ * @returns {boolean}
+ */
+export const hasCourseStarted = (start) => dayjs(start).isBefore(dayjs());
+
+/**
+ * Returns assignable course runs within the threshold of within the subsidies expiration date
+ * offset by the DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION constant. It sorts it from the soonest expiring
+ * enroll-by date and the enroll-by date and upgrade deadline has been normalized to ISO format.
```
pwnage101, line 570: Can you restate the goal differently? Based on my reading of the logic, it seems there are roughly two goals:

1. one goal related to giving accepted assignments enough time to refund before subsidy expiry, 
2. another goal related to giving the learner enough time to accept an assignment before a possibly sudden enrollment deadline, causing the learner to have less than 90 days to accept.

If the idea is that we want both of these goals met, then I think the `Math.max` needs to be `Math.min` below, right?

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,97 @@ export const isLmsBudget = (
...
+    enrollBy: dayjs.unix(courseRun.enrollBy).toISOString(),
+    upgradeDeadline: dayjs.unix(courseRun.upgradeDeadline).toISOString(),
+  }));
+  const assignableCourseRunsFilter = ({ enrollBy, isActive }) => {
+    const enrollByDateThreshold = dayjs(enrollBy).isBefore(
+      Math.max(
+        dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate(),
+        dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days').toDate(),
+      ),
+    );
```
pwnage101, line 589: Logic changes to conform to my assumed goals above:
```suggestion
    const enrollByDateThreshold = dayjs(enrollBy).isBefore(
      dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate()
    ) && dayjs(enrollBy).isAfter(
      dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days').toDate(),
    );
```

src/components/learner-credit-management/data/utils.js
pwnage101: Proposed logic changes would be more extensive, actually: https://github.com/openedx/frontend-app-admin-portal/pull/1292/files#r1762034441

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +555,97 @@ export const isLmsBudget = (
...
+    if (isLateRedemptionAllowed) {
+      return enrollByDateThreshold;
+    }
```
pwnage101, line 592: Unfortunately, I think if we just completely ignore isActive, it would potentially expose runs that are unenrollable due to not being marketable, having an "Archived" availability, or not having a "published" status. Unfortunately any of these could be the case regardless of proximity to the enrollment deadline. We might have to bypass isActive and pull in those 3 checks into the frontend, or extract out a new field in the algolia indexing logic, like `is_marketable_published_unarchived" alongside the existing `is_active` property: https://github.com/openedx/enterprise-catalog/blob/555e018ef7a1e5c701642e729bbd2fd5c8a086bc/enterprise_catalog/apps/catalog/algolia_utils.py#L1111

pwnage101: Furthermore, we might just have to completely drop returning `enrollByDateThreshold` in the late redemption case. The entire point of late redemption is to be able to assign/redeem a course *after* the enrollment deadline, let alone after 90 days before the enrollment deadline 😛

src/components/learner-credit-management/data/utils.js
brobro10000: Great callout. By looking at the min value as opposed to the max value, we are taking into account the suggested changes 👍🏽

src/components/learner-credit-management/data/utils.js
brobro10000: Given the upcoming PR, we are looking at an explicit `isLateEnrollmentEligibile` field which mimics the derived value from the learner portal [here](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/f1887e42b7fc39623b56184c17258ef9a3f578fa/src/components/app/data/utils.js#L523-L531).

src/components/learner-credit-management/data/constants.js
```diff
@@ -111,6 +111,14 @@ export const START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS = 14;
 // Default empty content_price value
 export const EMPTY_CONTENT_PRICE_VALUE = 0;
 
+// Course Run availability map
+export const COURSE_AVAILABILITY_MAP = {
```
adamstankiewicz, line 115: [nit/clarification] Could we re-use the existing `COURSE_AVAILABILITY_MAP` defined [here](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/f1887e42b7fc39623b56184c17258ef9a3f578fa/src/components/app/data/constants.js#L21-L26)?

src/components/learner-credit-management/data/utils.js
```diff
@@ -577,18 +585,23 @@ export const hasCourseStarted = (start) => dayjs(start).isBefore(dayjs());
...
-    const enrollByDateThreshold = dayjs(enrollBy).isBefore(
-      Math.max(
-        dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate(),
-        dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days').toDate(),
-      ),
-    );
+  const assignableCourseRunsFilter = ({
+    enrollBy, isActive, hasEnrollBy = false, isLateEnrollmentEligible = false,
+  }) => {
+    let enrollByDateThreshold = true;
```
adamstankiewicz, line 594: nit: perhaps `isEligibleForEnrollment` to indicate this a boolean in its usage.

src/components/learner-credit-management/data/utils.js
```diff
@@ -564,6 +565,13 @@ export const isLmsBudget = (
  */
 export const hasCourseStarted = (start) => dayjs(start).isBefore(dayjs());
 
+export function isArchived(courseRun) {
```
adamstankiewicz, line 568: nit: (discussed already, but posting for posterity) no longer used with `isLateEnrollmentEligible`

adamstankiewicz (approved):

src/components/learner-credit-management/data/constants.js
brobro10000: Moot point since I ended up removing it for the `isLateEnrollmentEligible`

src/components/learner-credit-management/data/utils.js
adamstankiewicz: [inform/context] Here is the upcoming PR @brobro10000 mentioned: https://github.com/openedx/enterprise-catalog/pull/945 😄 

> Unfortunately, I think if we just completely ignore isActive, it would potentially expose runs that are unenrollable due to not being marketable, having an "Archived" availability, or not having a "published" status.

Given the Learner Portal filtering we in place today for the late enrollment use case ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/f1887e42b7fc39623b56184c17258ef9a3f578fa/src/components/app/data/utils.js#L524-L531)), I believe the only check we're missing here is around the `availability="Archived"` case; great catch.

In the above linked enterprise-catalog PR, we're now going to expose a `is_late_enrollment_eligible` field to take into account the same checks the Learner Portal filter is doing as described above.

> For an MVP and testing purposes, this might be enough, but to be completely accurate we'd still need to loop back and add a add checks to mirror the whole "late enrollment buffer" logic from the learner portal in order to prevent showing really old runs.

The Algolia indexing of the `course_runs` list on top-level course objects includes the "late enrollment buffer" logic to not index really old runs (see above linked enterprise-catalog PR) that are well beyond the late enrollment buffer window. That said, there is the off-chance we could have a stale Algolia index whereby already-indexed courses that were eligible for late enrollment at the time of last index are _no longer_ eligible.

@brobro10000 is going to push another commit to this PR to ensure a similar check as done during Algolia indexing around "Is the `enroll_by` within the late enrollment buffer window?" to account for this edge case of stale Algolia index.

__mocks__/react-instantsearch-dom.jsx
```diff
@@ -31,6 +31,7 @@ const fakeHits = [
       start: mockCurrentStartDate,
       end: dayjs(mockCurrentStartDate).add(1, 'year').toISOString(),
       enroll_by: dayjs(mockCurrentStartDate).unix(),
```
adamstankiewicz, line 33: nit: curious whether we can either rename this mock date to be more generic or have separate variables to use for each related field.

As is, the `mockCurrentStartDate` is being used for `end`/`enroll_by` dates, too which could be a bit misleading/confusing.

src/components/learner-credit-management/data/constants.js
```diff
@@ -111,6 +111,9 @@ export const START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS = 14;
 // Default empty content_price value
 export const EMPTY_CONTENT_PRICE_VALUE = 0;
 
+//
```
adamstankiewicz, line 114: nit: empty comment

src/components/learner-credit-management/data/utils.js
```diff
@@ -577,20 +585,27 @@ export const hasCourseStarted = (start) => dayjs(start).isBefore(dayjs());
...
+    let isEligibleForEnrollment = true;
+    if (hasEnrollBy) {
+      isEligibleForEnrollment = dayjs(enrollBy).isBefore(
+        Math.min(
+          dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate(),
+          dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days').toDate(),
+        ),
+      );
+    }
+    if (hasCourseStarted(enrollBy) && isLateRedemptionAllowed) {
```
adamstankiewicz, line 603: nit: Should we be using a function `hasCourseStarted` with an `enrollBy` date? Perhaps this util function should be renamed to be more generic, so it's not tied explicitly to "start" dates based on its name.

adamstankiewicz (approved):

src/components/learner-credit-management/data/utils.js
```diff
@@ -551,3 +556,111 @@ export const isLmsBudget = (
...
+    enrollBy, isActive, hasEnrollBy = false, isLateEnrollmentEligible = false,
+  }) => {
+    let isEligibleForEnrollment = true;
+    if (hasEnrollBy) {
+      isEligibleForEnrollment = dayjs(enrollBy).isBefore(
+        Math.min(
+          dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate(),
+          dayjs().add(DAYS_UNTIL_ASSIGNMENT_ALLOCATION_EXPIRATION, 'days').toDate(),
+        ),
+      );
```
pwnage101, line 601: You sure just changing this to min is sufficient? I ask again, what's the goal? The docstring was updated to match this logic, but still the desired outcome is not explicitly stated.  If the goal is as I described it previously, then I think this is still incorrect because it guarantees that learners ALWAYS have less than 90 days to accept an assignment.

adamstankiewicz (approved):

src/components/learner-credit-management/data/utils.js
```diff
@@ -664,3 +622,77 @@ export const getNormalizedEnrollByDate = ({ enrollBy }) => {
   }
   return enrollBy;
 };
+
+/**
+ * Filters assignable course runs based on the following criteria:
+ *  - If hasEnrollBy, we return assignments with enroll before the soonest of the two date: The subsidy expiration
+ *    date - refund threshold OR today offset by the 90-day allocation threshold for an assignment denoted as
+ *    isEligibleForEnrollment
```
pwnage101, line 630: Nit: clarify that "the purpose of making run visibility sensitive to the subsidy expiration date is to incentivize renewal, and the purpose of making run visibility sensitive to today's date is to ..." (fill in the ellipsis).

pwnage101 (approved):

src/components/learner-credit-management/data/utils.js
brobro10000: Updated the docstring language to be specific of the intentions 👍🏽

src/components/learner-credit-management/data/utils.js
```diff
@@ -628,15 +628,20 @@ export const getNormalizedEnrollByDate = (enrollBy) => {
...
  *    each run.
  *
  *  Based on the above criteria, if isLateRedemptionAllowed is false, filter on if the course run isActive AND
  *  isEligibleForEnrollment
  *
- *  We transform the assignedCourseRuns data to normalize the start and enrollby dates based on the functions
+ *  The main purpose of the filter is to ensure that course runs for a
+ *  course are within the enterprises LC subsidy duration
+ *  The inclusion of the increased sensitivity reduces the chance of a specific run
+ *  (which may be allocated but not accepted) falling outside the date range of the subsidy expiration date.
```
pwnage101, line 640: IMO this doesn't directly speak to the reason for making the run visibility sensitive to the assignment expiration date. Maybe mention the specifics about how it's a product requirement or contractual requirement?

src/components/learner-credit-management/data/utils.js
```diff
@@ -625,9 +624,8 @@ export const getNormalizedEnrollByDate = (enrollBy) => {
 
 /**
  * Filters assignable course runs based on the following criteria:
- *  - If hasEnrollBy, we return assignments with enroll before the soonest of the two date: The subsidy expiration
- *    date - refund threshold OR today offset by the 90-day allocation threshold for an assignment denoted as
- *    isEligibleForEnrollment
+ *  - If hasEnrollBy, we return assignments with enroll before the soonest date: The subsidy expiration
```
adamstankiewicz, line 627: nit: `enroll` -> `enrollBy` (deferred)

src/components/learner-credit-management/data/utils.js
adamstankiewicz: The previously implemented 90-day assignment expiration date is now removed, after clarifying with product/Joe.

adamstankiewicz (approved):

PR #1304, opened 2024-09-16
fix: Updated courseKey with courseTitle for charts

irfanuddinahmad:
# For all changes


saleem-latif (approved):

mahamakifdar19 (changes requested):
I believe we should truncate long titles. We can set a character limit and ensure that titles are truncated at word boundaries and an ellipsis is added at the end to indicate truncation.

mahamakifdar19 (approved):

PR #1313, opened 2024-09-18
fix: don't consider elapsed enrollBy in assignable course runs

adamstankiewicz:
# For all changes

### DIFF
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -660,13 +660,16 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
   }) => {
     let isEligibleForEnrollment = true;
     if (hasEnrollBy) {
-      isEligibleForEnrollment = dayjs(enrollBy).isBefore(
-        minimumEnrollByDateFromToday({ subsidyExpirationDatetime }),
+      const enrollByDate = dayjs(enrollBy);
+      // Determine eligibility based on the provided enrollBy is and the subsidy expiration date - refund threshold
+      isEligibleForEnrollment = (
+        !isDateBeforeToday(enrollByDate)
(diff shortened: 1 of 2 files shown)


brobro10000 (approved):

src/components/learner-credit-management/data/utils.js
```diff
@@ -660,13 +660,16 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
   }) => {
     let isEligibleForEnrollment = true;
     if (hasEnrollBy) {
-      isEligibleForEnrollment = dayjs(enrollBy).isBefore(
-        minimumEnrollByDateFromToday({ subsidyExpirationDatetime }),
+      const enrollByDate = dayjs(enrollBy);
+      // Determine eligibility based on the provided enrollBy is and the subsidy expiration date - refund threshold
+      isEligibleForEnrollment = (
+        !isDateBeforeToday(enrollByDate)
```
pwnage101, line 666: Maybe add a comment that we'll want to eventually loop back to this and allow a certain buffer to allow enrollBy to be a few days before today?

pwnage101: hm, the code below seems to consider late enrollment buffer days, so I just might be misunderstanding something...

pwnage101: oh okay that overrides this, LGTM

pwnage101 (approved):

src/components/learner-credit-management/data/utils.js
adamstankiewicz: Isn't this the use case for late enrollment, handled below, though? If late enrollment is not enabled for a budget, I'm not sure we'd want to expose runs that are otherwise not enrollable without late enrollment enabled? 🤔

adamstankiewicz: Ha, sorry didn't see your updated comments until after submitting the above comment :) 

Agreed this should be adequate.

PR #1317, opened 2024-09-24
fix: Update threshold dates for displaying course runs

brobro10000:
Updates the logic which determines the course runs displayed in the learner credit assignment dropdown.
Ensures the following logic:
  - Given an existing `enrollBy` date for a course run, the `enrollBy` date is after today's date. If late enrollment is allowed, the `enrollBy` date is after today - `LATE_ENROLLMENTS_BUFFER_DAYS`
  - Given an existing `enrollmentStart` and `start` date for a course run, the minimum date between an enrollmentStart and start date must be before the `subsidyExpirationDate` - `MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS`

# For all changes

### DIFF
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -587,11 +587,11 @@ const isStartDateWithinThreshold = ({ enrollStart, start, subsidyExpirationDatet
 
 const isEnrollByDateWithinThreshold = ({ enrollBy, isLateRedemptionAllowed = false }) => {
   if (!enrollBy) { return true; }
-  let enrollByCutoffDate = dayjs();
+  let enrollmentEffectiveDate = dayjs();
   if (isLateRedemptionAllowed) {
-    enrollByCutoffDate = enrollByCutoffDate.subtract(LATE_ENROLLMENTS_BUFFER_DAYS, 'days');
+    enrollmentEffectiveDate = enrollmentEffectiveDate.subtract(LATE_ENROLLMENTS_BUFFER_DAYS, 'days');
   }
-  return dayjs(enrollBy).isAfter(enrollByCutoffDate, 'seconds');
+  return dayjs(enrollBy).isAfter(enrollmentEffectiveDate, 'seconds');
 };
 
 export const isCourseSelfPaced = ({ pacingType }) => pacingType === COURSE_PACING_MAP.SELF_PACED;
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -15,6 +15,7 @@ import {
   LATE_ENROLLMENTS_BUFFER_DAYS,
   LOW_REMAINING_BALANCE_PERCENT_THRESHOLD,
   MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS,
+  MAX_MILLISECONDS,
   NO_BALANCE_REMAINING_DOLLAR_THRESHOLD,
   START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS,
 } from './constants';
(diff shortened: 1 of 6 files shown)


src/components/learner-credit-management/data/utils.js
```diff
@@ -656,20 +676,18 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
     upgradeDeadline: dayjs.unix(courseRun.upgradeDeadline).toISOString(),
   }));
   const assignableCourseRunsFilter = ({
-    enrollBy, isActive, hasEnrollBy, isLateEnrollmentEligible,
+    enrollBy, start, isActive, hasEnrollBy, isLateEnrollmentEligible,
   }) => {
     let isEligibleForEnrollment = true;
+    // Determine eligibility based on the provided enrollBy, start, and enrollmentStart date
+    isEligibleForEnrollment = enrollByDateThreshold({ enrollBy, isLateRedemptionAllowed }) && startDateThreshold({
+      enrollmentStart: dayjs().toDate(),
```
adamstankiewicz, line 684: [inform] I have the field currently named as `enroll_start` within the `course_runs` metadata in Algolia (i.e., `enrollStart` here). Not sure if we want to align on one or the other. Do you have a preference? I kept it as `enroll_start` so its somewhat parity with the existing `enrollBy` (e.g., we don't say `enrollmentBy`).

brobro10000: Synced asynchrounously about this, but for posterity, we are falling back to the fields you set in your [enterprise catalog PR](https://github.com/openedx/enterprise-catalog/pull/949) of `enrollStart` and `hasEnrollStart`

src/components/learner-credit-management/data/utils.js
```diff
@@ -570,6 +570,29 @@ export const minimumEnrollByDateFromToday = ({ subsidyExpirationDatetime }) => M
   dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate(),
 );
 
+const subsidyExpirationRefundCutoffDate = ({ subsidyExpirationDatetime }) => dayjs(subsidyExpirationDatetime).subtract(MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS, 'days').toDate();
```
iloveagent57, line 573: nit: you could re-use this directly above in `minimumEnrollByDateFromToday`. Actually now that I'm looking at that function, is the `min()` doing anything?

adamstankiewicz (approved):

src/components/learner-credit-management/data/utils.js
```diff
@@ -570,6 +570,29 @@ export const minimumEnrollByDateFromToday = ({ subsidyExpirationDatetime }) => M
...
+  const timeStampEnrollStartDate = dayjs(enrollStart).unix();
+  const earliestStartDate = Math.min(timeStampStartDate, timeStampEnrollStartDate);
+  return dayjs(
+    earliestStartDate,
+  ).isBefore(subsidyExpirationRefundCutoffDate({ subsidyExpirationDatetime }), 'seconds');
+};
+
+const isEnrollByDateWithinThreshold = ({ enrollBy, isLateRedemptionAllowed = false }) => {
+  if (!enrollBy) { return true; }
+  let enrollByCutoffDate = dayjs();
```
iloveagent57, line 589: nit: this might read more naturally as `enrollmentEffectiveDate`.
Then the return statement would read as `return dayjs(enrollBy).isAfter(enrollmentEffectiveDate, 'seconds');`, which (IMHO) is slightly easier to parse the intention of.

iloveagent57 (approved):
Don't let my nits stop you from merging when you're comfortable.

src/components/learner-credit-management/data/utils.js
brobro10000: I believe the `min` portion was left in incase additional criteria would need to be added, for right now its determining a fallback enrollby date to display within `getNormalizedEnrollByDate`.

src/components/learner-credit-management/data/utils.js
```diff
@@ -570,6 +571,29 @@ export const minimumEnrollByDateFromToday = ({ subsidyExpirationDatetime }) => M
...
+  if (isLateRedemptionAllowed) {
+    enrollByCutoffDate = enrollByCutoffDate.subtract(LATE_ENROLLMENTS_BUFFER_DAYS, 'days');
```
pwnage101, line 592: LATE_ENROLLMENTS_BUFFER_DAYS looks like a typo (extra S)

src/components/learner-credit-management/data/utils.js
brobro10000: Updated 👍🏽

src/components/learner-credit-management/data/utils.js
```diff
@@ -653,25 +677,27 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
...
+        enrollStart,
+        start,
+        subsidyExpirationDatetime,
+      },
+    );
+    if (hasEnrollBy && isLateRedemptionAllowed && isDateBeforeToday(enrollBy)) {
+      return isLateEnrollmentEligible && isEligibleForEnrollment;
     }
     // General courseware filter
     return isActive && isEligibleForEnrollment;
```
pwnage101, line 703: Nit: isEligibleForEnrollment is a common consideration, maybe return early if that's false? That way we can end the entire function with a sequence of checks and comments for ultimate readability.  Furthermore, you don't need to check hasEnrollBy or isDateBeforeToday(enrollBy) anymore because that's already rolled into isEligibleForEnrollment.
```suggestion
    if (!isEligibleForEnrollment) {
        // Basic checks against this content's critical dates and their relation to
        // the current date and subsidy expiration date have failed.
        return false;
    }
    if (isLateRedemptionAllowed) {
      // Special case: late enrollment has been enabled by ECS for this budget, and 
      // isEligibleForEnrollment already succeeded, so we know that late enrollment
      // would be happy given enrollment deadline of the course.  Now all we need
      // to do is make sure the run itself is generally eligible for late enrollment
      // (e.g. it is published and not archived, etc.)
      return isLateEnrollmentEligible;
    }
    // General courseware filter.  Content that's active is generally eligible for
    // enrollment (e.g. it is published and not archived and generally enrollable,
    // etc.).
    return isActive;
```

pwnage101 (approved):
My comment is also non-blocking, but for the sake of readability I'd like my Nit addressed eventually.

src/components/learner-credit-management/data/utils.js
brobro10000: Definitely leaving this open to put into consideration for a refined implementation. 👍🏽

PR #1320, opened 2024-09-25
fix: switch to valueof when doing unix conversion

brobro10000:
When doing comparisons of date using dates and timestamp, use the milliseconds conversion to increase precision. With this change the start date logic comparison of "Is the earliest of 2 start dates (enrollStart OR  start) before the subsidy expiration date minus 14 day threshold.", the resolved logic is now correct.

# For all changes

### DIFF
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -577,8 +577,8 @@ const isStartDateWithinThreshold = ({ enrollStart, start, subsidyExpirationDatet
   if (!start && !enrollStart) {
     return true;
   }
-  const timeStampStartDate = dayjs(start).unix() || MAX_MILLISECONDS;
-  const timeStampEnrollStartDate = dayjs(enrollStart).unix() || MAX_MILLISECONDS;
+  const timeStampStartDate = dayjs(start).valueOf() || MAX_MILLISECONDS;
+  const timeStampEnrollStartDate = dayjs(enrollStart).valueOf() || MAX_MILLISECONDS;
   const earliestStartDate = Math.min(timeStampStartDate, timeStampEnrollStartDate);
   return dayjs(
     earliestStartDate,
--- a/src/components/learner-credit-management/data/utils.js
+++ b/src/components/learner-credit-management/data/utils.js
@@ -15,7 +15,6 @@ import {
   LATE_ENROLLMENTS_BUFFER_DAYS,
   LOW_REMAINING_BALANCE_PERCENT_THRESHOLD,
   MAX_ALLOWABLE_REFUND_THRESHOLD_DAYS,
-  MAX_MILLISECONDS,
   NO_BALANCE_REMAINING_DOLLAR_THRESHOLD,
   START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS,
 } from './constants';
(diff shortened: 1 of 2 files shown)


macdiesel (approved):

src/components/learner-credit-management/data/utils.js
```diff
@@ -647,23 +621,51 @@ export const getNormalizedEnrollByDate = (enrollBy) => {
...
  *
  *  The main purpose of the filter is to ensure that course runs for a
  *  course are within the enterprises LC subsidy duration
  *  The inclusion of the increased sensitivity reduces the chance of a specific run
  *  (which may be allocated but not accepted) falling outside the date range of the subsidy expiration date
  *  refund threshold.
  *
- *  We transform the assignedCourseRuns data to normalize the start and enrollBy dates based on the functions
+ *  We transform the assignedCourseRuns data to normalize the start and
+ *  enrollBy and enrollStart dates based on the functions
```
adamstankiewicz, line 668: nit: can remove an "and"; also not sure the "based on the functions" is needed? "Functions" is fairly ambiguous. Is there another way to phrase "based on the functions" to be more explicit and/or should "based on the functions" be removed altogether?

src/components/learner-credit-management/data/utils.js
```diff
@@ -680,36 +682,50 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
...
   const assignableCourseRunsFilter = ({
-    enrollBy, enrollStart, start, isActive, hasEnrollBy, isLateEnrollmentEligible,
+    enrollBy, enrollStart, start, hasEnrollBy, hasEnrollStart, isActive, isLateEnrollmentEligible,
   }) => {
+    const isWithinEnrollByDateThreshold = isEnrollByDateWithinThreshold({
+      hasEnrollBy,
+      enrollBy,
+      isLateRedemptionAllowed,
+    });
+    const isWithinStartDateThreshold = isStartDateWithinThreshold({
```
adamstankiewicz, line 694: nit: perhaps could be simplified as `isEnrollByDateValid` and `isStartDateValid`?

src/components/learner-credit-management/data/utils.js
```diff
@@ -680,36 +682,50 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
...
-    if (hasEnrollBy && isLateRedemptionAllowed && isDateBeforeToday(enrollBy)) {
-      return isLateEnrollmentEligible && isEligibleForEnrollment;
+    const isEligibleForEnrollment = isWithinEnrollByDateThreshold && isWithinStartDateThreshold;
+
+    if (!isEligibleForEnrollment) {
+      // Basic checks against this content's critical dates and their relation to
+      // the current date and subsidy expiration date have failed.
+      return false;
+    }
+    if (isLateRedemptionAllowed) {
```
adamstankiewicz, line 709: [clarification] Is this intentionally no longer relying on `hasEnrollBy && isDateBeforeToday(enrollBy)`?

brobro10000: Updated to reinclude `if (hasEnrollBy && isLateRedemptionAllowed && isDateBeforeToday(enrollBy))`
I had missed that if `hasEnrollBy` is false, we return true regardless, so this check is required to return late enrollment runs 👍🏽

src/components/learner-credit-management/data/utils.js
brobro10000: Removed 👍🏽

src/components/learner-credit-management/data/utils.js
```diff
@@ -680,36 +681,50 @@ export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime,
...
+      // the current date and subsidy expiration date have failed.
+      return false;
+    }
     if (hasEnrollBy && isLateRedemptionAllowed && isDateBeforeToday(enrollBy)) {
-      return isLateEnrollmentEligible && isEligibleForEnrollment;
+      // Special case: late enrollment has been enabled by ECS for this budget, and
+      // isEligibleForEnrollment already succeeded, so we know that late enrollment
+      // would be happy given enrollment deadline of the course.  Now all we need
+      // to do is make sure the run itself is generally eligible for late enrollment
+      // (e.g. it is published and not archived, etc.)
```
adamstankiewicz, line 713: nit: `isLateEnrollmentEligible` does not take into account whether a course run is published. Cross-check the logic abstracted within `isLateEnrollmentEligible` ([source](https://github.com/openedx/enterprise-catalog/blob/4db5daa4c5f7b8172e99174e3e036683c8886da1/enterprise_catalog/apps/catalog/algolia_utils.py#L1255-L1267)). 

```python
def _get_is_late_enrollment_eligible(course_run):
    """
    Determines if the course run is eligible for late enrollment:
      * Must not be archived
      * Must have a marketing URL
      * Must have seats
    """
    is_archived = course_run.get('availability') == 'Archived'
    has_marketing_url = bool(course_run.get('marketing_url'))
    has_seats = bool(course_run.get('seats'))
    if is_archived or not has_marketing_url or not has_seats:
        return False
    return True
```

adamstankiewicz (approved):
LGTM, with a suggestion to update a comment related to late enrollment.

PR #1329, opened 2024-10-05
feat: Create group base detail page 

kiram15:
# Description
Creates the card for the groups detail page. Includes functionality for editing the group name and deleting the group. 
https://2u-internal.atlassian.net/browse/ENT-9506

# Test plan
- npm run start:stage
- because we don't actually have the button that will open it, just navigate straight to https://localhost.stage.edx.org:1991/exec-ed-2u-integration-qa/admin/people-management/89e6451c-b187-44b2-b546-ad4511fc34ea
- open up edit and delete modals

Screenshot 2024-10-04 at 1 01 50 PM

Screenshot 2024-10-04 at 1 01 56 PM
Screenshot 2024-10-05 at 2 16 12 AM

# For all changes

### DIFF
--- a/src/components/PeopleManagement/DeleteGroupModal.jsx
+++ b/src/components/PeopleManagement/DeleteGroupModal.jsx
@@ -2,10 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useParams } from 'react-router-dom';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { logError } from '@edx/frontend-platform/logging';
 import {
   ActionRow, Button, ModalDialog, useToggle,
 } from '@openedx/paragon';
 import { RemoveCircleOutline } from '@openedx/paragon/icons';
+
 import GeneralErrorModal from './GeneralErrorModal';
 import { ROUTE_NAMES } from '../EnterpriseApp/data/constants';
 
@@ -17,12 +19,15 @@ const DeleteGroupModal = ({
   const { enterpriseSlug } = useParams();
   const [isErrorOpen, openError, closeError] = useToggle(false);
   const removeEnterpriseGroup = async () => {
-    const response = await LmsApiService.removeEnterpriseGroup(group?.uuid);
-    if (response.status === 204) {
-      close();
-      // redirect back to the people management page
-      window.location.href = `/${enterpriseSlug}/admin/${ROUTE_NAMES.peopleManagement}`;
-    } else {
+    try {
+      const response = await LmsApiService.removeEnterpriseGroup(group?.uuid);
+      if (response.status === 204) {
+        close();
+        // redirect back to the people management page
+        window.location.href = `/${enterpriseSlug}/admin/${ROUTE_NAMES.peopleManagement}`;
+      }
+    } catch (error) {
+      logError(error);
       openError();
     }
   };
--- a/src/components/PeopleManagement/DeleteGroupModal.jsx
+++ b/src/components/PeopleManagement/DeleteGroupModal.jsx
@@ -57,7 +57,7 @@ const DeleteGroupModal = ({
             id="people.management.delete.group.modal.body.2"
             defaultMessage={
             'By deleting this group you will no longer be able to track analytics associated'
-            + 'with this group and the group will be removed from your People Management page.'
+            + ' with this group and the group will be removed from your People Management page.'
           }
             description="Warning shown when deleting a group part 2."
           />
--- a/src/components/PeopleManagement/DeleteGroupModal.jsx
+++ b/src/components/PeopleManagement/DeleteGroupModal.jsx
@@ -1,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useParams } from 'react-router-dom';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import {
+  ActionRow, Button, ModalDialog, useToggle,
+} from '@openedx/paragon';
+import { RemoveCircleOutline } from '@openedx/paragon/icons';
+import GeneralErrorModal from './GeneralErrorModal';
+import { ROUTE_NAMES } from '../EnterpriseApp/data/constants';
+
+import LmsApiService from '../../data/services/LmsApiService';
+
+const DeleteGroupModal = ({
+  group, isOpen, close,
+}) => {
+  const { enterpriseSlug } = useParams();
+  const [isErrorOpen, openError, closeError] = useToggle(false);
+  const removeEnterpriseGroup = async () => {
+    const response = await LmsApiService.removeEnterpriseGroup(group?.uuid);
+    if (response.status === 204) {
+      close();
+      // redirect back to the people management page
+      window.location.href = `/${enterpriseSlug}/admin/${ROUTE_NAMES.peopleManagement}`;
+    } else {
+      openError();
+    }
+  };
+  return (
+    <>
+      <GeneralErrorModal
+        isOpen={isErrorOpen}
+        close={closeError}
+      />
+      <ModalDialog
+        title="Delete group"
+        isOpen={isOpen}
+        onClose={close}
+        hasCloseButton
+        isFullscreenOnMobile
+      >
+        <ModalDialog.Header>
+          <ModalDialog.Title>Delete group?</ModalDialog.Title>
+        </ModalDialog.Header>
+        <ModalDialog.Body>
+          <p className="pb-3">
+            <FormattedMessage
+              id="people.management.delete.group.modal.body.1"
+              defaultMessage="This action cannot be undone."
+              description="Warning shown when deleting a group."
+            />
+          </p>
+          <FormattedMessage
+            id="people.management.delete.group.modal.body.2"
+            defaultMessage={
+            'By deleting this group you will no longer be able to track analytics associated'
+            + 'with this group and the group will be removed from your People Management page.'
+          }
+            description="Warning shown when deleting a group part 2."
+          />
+        </ModalDialog.Body>
+        <ModalDialog.Footer>
+          <ActionRow>
+            <ModalDialog.CloseButton variant="tertiary">
+              Go back
+            </ModalDialog.CloseButton>
+            <Button variant="danger" data-testid="delete-group-button" onClick={removeEnterpriseGroup} iconBefore={RemoveCircleOutline}>
+              Delete group
+            </Button>
+          </ActionRow>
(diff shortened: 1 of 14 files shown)


src/components/PeopleManagement/GroupDetailPage.jsx
```diff
@@ -0,0 +1,115 @@
...
+        activeLabel={enterpriseGroup?.name}
+      />
+
+      <Card orientation="horizontal">
+        <Card.Body>
+          <Card.Header
+            title={budgetNameAndEdit}
+            subtitle={`${enterpriseGroup?.acceptedMembersCount} accepted members`}
+          />
+          <Card.Section className="pt-1 x-small">Created on</Card.Section>
```
kiram15, line 89: Need to add in created to the serializer, will add this value in with my next ticket

src/components/PeopleManagement/EditGroupNameModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+import LmsApiService from '../../data/services/LmsApiService';
+
+const EditGroupNameModal = ({
+  group, isOpen, close, setShowToast, setToastMessage,
+}) => {
+  const [name, setName] = useState(group?.name);
+  const editEnterpriseGroup = async () => {
+    try {
+      const formData = { name };
+      const response = await LmsApiService.updateEnterpriseGroup(group?.uuid, formData);
```
katrinan029, line 17: I believe the url needs to be updated since the patch request is not going through. I think you just need to add a forward slash at the end of the url. we might also want to consider how we want to handle the patch request error. what kind of error do we want to display?

https://github.com/user-attachments/assets/79af3692-6be6-4884-a99d-8e22a1dc0428

src/components/PeopleManagement/EditGroupNameModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+      if (response.status === 200) {
+        setShowToast(true);
+        setToastMessage('Group name updated');
```
katrinan029, line 20: once the request succeeds, i think we should close the modal.

src/components/PeopleManagement/EditGroupNameModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+          />
+        </Form.Group>
+      </ModalDialog.Body>
+
+      <ModalDialog.Footer>
+        <ActionRow>
+          <ModalDialog.CloseButton variant="tertiary">
+            Cancel
+          </ModalDialog.CloseButton>
+          <Button variant="primary" onClick={editEnterpriseGroup}>Done</Button>
```
katrinan029, line 55: let's consider making this a stateful button in case of an error or a slow request, we'd want to show some sort of feedback to the user.

katrinan029 (changes requested):
looks good so far! I have some feedback regarding handling errors

src/components/PeopleManagement/GroupDetailPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+          },
+        ]}
+        activeLabel={enterpriseGroup?.name}
+      />
+
+      <Card orientation="horizontal">
+        <Card.Body>
+          <Card.Header
+            title={budgetNameAndEdit}
+            subtitle={`${enterpriseGroup?.acceptedMembersCount} accepted members`}
```
katrinan029, line 85: I think we should avoid this conditional check by adding in a loading state and `useEffect` hook. i'm seeing this first say `undefined` in the UI and then the count populates (see screen recording). 

To guarantee that the `enterpriseGroup` data exists before rendering the card, we can do something like the sample code below. When the loading state is true, we could render the `Skeleton`.

>  const [isLoading, setIsLoading] = useState(true);
> 
>   useEffect(() => {
>     if (enterpriseGroup !== undefined) {
>       setIsLoading(false);
>     }
>   }, [enterpriseGroup]);

https://github.com/user-attachments/assets/5118fb1a-f4d2-4805-8937-ab0be94548ac

src/components/PeopleManagement/EditGroupNameModal.jsx
```diff
@@ -0,0 +1,109 @@
...
+  const [buttonState, setButtonState] = useState('default');
+  const editEnterpriseGroup = async () => {
+    setButtonState('pending');
+    const formData = { name };
+    const response = await LmsApiService.updateEnterpriseGroup(group?.uuid, formData);
+    if (response.status === 200) {
+      setButtonState('complete');
+      close();
+      setShowToast(true);
+      setToastMessage('Group name updated');
```
katrinan029, line 26: in addition to these states, I think we should also trigger a re-render for the group name card. Currently the group name stays the same unless the user triggers a refresh.

https://github.com/user-attachments/assets/008be0a2-d055-4f6b-a3e4-4339e92378f7

src/components/PeopleManagement/EditGroupNameModal.jsx
```diff
@@ -0,0 +1,109 @@
...
+  group, isOpen, close, setShowToast, setToastMessage,
+}) => {
+  const intl = useIntl();
+  const [isErrorOpen, openError, closeError] = useToggle(false);
+  const [name, setName] = useState(group?.name);
+  const [buttonState, setButtonState] = useState('default');
+  const editEnterpriseGroup = async () => {
+    setButtonState('pending');
+    const formData = { name };
+    const response = await LmsApiService.updateEnterpriseGroup(group?.uuid, formData);
```
katrinan029, line 21: we should wrap this in a `try...catch` in order to handle the errors instead of the conditional `if...else` check, otherwise we wouldn't be able to catch errors to open up the error modal. 

Uploading Screen Recording 2024-10-09 at 1.49.51 PM.mov…

src/components/PeopleManagement/GroupDetailPage.jsx
```diff
@@ -0,0 +1,125 @@
...
+    <FormattedMessage
+      id="adminPortal.peopleManagement.groupDetail.deleteGroup.icon"
+      defaultMessage="Delete group"
+      description="Tooltip to confirm meaning of trash icon."
+    />
+  );
+
+  return (
+    <div className="pt-4 pl-4">
+      {isLoading && <Skeleton className="mt-3" height={200} count={1} />}
```
katrinan029, line 44: [nit]: we can consolidate this logic check in the line below:

```
 {!isLoading ? (
        <>
          <DeleteGroupModal
            group={enterpriseGroup}
            isOpen={isDeleteModalOpen}
            close={closeDeleteModal}
          />
          <EditGroupNameModal
            group={enterpriseGroup}
            isOpen={isEditModalOpen}
            close={closeEditModal}
            handleNameUpdate={handleNameUpdate}
          />
          <Breadcrumb
            aria-label="people management breadcrumb navigation"
            links={[
              {
                label: intl.formatMessage({
                  id: 'adminPortal.peopleManagement.breadcrumb',
                  defaultMessage: 'People Management',
                  description:
                    'Breadcrumb label for the people management page',
                }),
                href: `/${enterpriseSlug}/admin/${ROUTE_NAMES.peopleManagement}`,
              },
            ]}
            activeLabel={groupName}
          />
          <Card orientation="horizontal">
            <Card.Body>
              <Card.Header
                title={(
                  <>
                    <span className="pr-1">{groupName}</span>
                    <IconButton
                      key="editGroupTooltip"
                      src={Edit}
                      iconAs={Icon}
                      alt="Edit group"
                      className="mr-2"
                      onClick={openEditModal}
                      size="sm"
                      data-testid="edit-modal-icon"
                    />
                  </>
              )}
                subtitle={`${enterpriseGroup.acceptedMembersCount} accepted members`}
              />
              <Card.Section className="pt-1 x-small">Created on</Card.Section>
            </Card.Body>
            <Card.Footer
              className="justify-content-end"
              orientation="horizontal"
            >
              <IconButtonWithTooltip
                alt="icon to trash group"
                key="trashGroupTooltip"
                tooltipPlacement="top"
                tooltipContent={tooltipContent}
                src={Delete}
                iconAs={Icon}
                data-testid="delete-group-icon"
                className="mr-2"
                onClick={openDeleteModal}
              />
              <Hyperlink
                className="btn btn-primary"
                target="_blank"
                destination={`/${enterpriseSlug}/admin/${ROUTE_NAMES.learners}`}
              >
                View group progress
              </Hyperlink>
            </Card.Footer>
          </Card>
        </>
      ) : <Skeleton className="mt-3" height={200} count={1} /> }
```

src/components/PeopleManagement/DeleteGroupModal.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+const DeleteGroupModal = ({
+  group, isOpen, close,
+}) => {
+  const { enterpriseSlug } = useParams();
+  const [isErrorOpen, openError, closeError] = useToggle(false);
+  const removeEnterpriseGroup = async () => {
+    try {
+      const response = await LmsApiService.removeEnterpriseGroup(group?.uuid);
+      if (response.status === 204) {
```
katrinan029, line 24: we can remove this check since the promise is already wrapped in a `try...catch`

src/components/PeopleManagement/DeleteGroupModal.jsx
```diff
@@ -0,0 +1,88 @@
...
+              id="people.management.delete.group.modal.body.1"
+              defaultMessage="This action cannot be undone."
+              description="Warning shown when deleting a group."
+            />
+          </p>
+          <FormattedMessage
+            id="people.management.delete.group.modal.body.2"
+            defaultMessage={
+            'By deleting this group you will no longer be able to track analytics associated'
+            + 'with this group and the group will be removed from your People Management page.'
```
katrinan029, line 60: minor change here to add a space in front of `with` or after `associated` so that it doesn't look like one word.
Screenshot 2024-10-15 at 8 31 51 AM

src/components/PeopleManagement/GroupDetailPage.jsx
```diff
@@ -0,0 +1,124 @@
...
+                      className="mr-2"
+                      onClick={openEditModal}
+                      size="sm"
+                      data-testid="edit-modal-icon"
+                    />
+                  </>
+              )}
+                subtitle={`${enterpriseGroup.acceptedMembersCount} accepted members`}
+              />
+              <Card.Section className="pt-1 x-small">Created on</Card.Section>
```
katrinan029, line 92: can we add in the created date since that's being returned by the response now?

katrinan029 (approved):
just a couple of minor changes but it's good to go once they get addressed!

PR #1359, opened 2024-11-26
feat: enable allocation of restricted runs

pwnage101:
ENT-9411

Two skeletons appear briefly while simulating a slow network connection.  Note that the dropdown contains only one skeleton item despite ultimately displaying two runs in the following screenshot:
Screenshot 2024-12-04 at 11 37 11 AM copy

After the page fully loads, the skeleton states all disappear, and the number of courses (left side) jumps up from "(1) available date" to "(2) available dates".
Screenshot 2024-12-04 at 11 38 32 AM copy

Selecting the restricted run opens the assignment modal and displays the correct fixed price for the restricted run:
Screenshot 2024-12-04 at 11 50 38 AM

After assigning it to myself, I see the assignment appear in the assignments table:
image


src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
```diff
@@ -54,10 +55,20 @@ const useCourseCardMetadata = ({
     title,
     courseRuns,
   } = course;
+  const {
+    dataByContentKey: catalogContainsRestrictedRunsData,
+    isLoading: isLoadingcatalogContainsRestrictedRuns,
+  } = useCatalogContainsContentItemsMultipleQueries(
+    subsidyAccessPolicy.catalogUuid,
+    // Pass only restricted runs.
+    courseRuns?.filter(run => run.restrictionType === 'custom-e2e-enterprise'), // TODO: replace with constant
```
adamstankiewicz, line 64: Should this be passing the course run **keys** vs. the entire course run objects? It appears the entire course run object is added in the query key vs. just the course run `key`.

adamstankiewicz: nit: I also wonder if it's worth abstracting a utility function for this filtering for restricted runs, given the same filtering logic is used in 2 places?

src/components/learner-credit-management/data/hooks/useCatalogContainsContentItemsMultipleQueries.js
```diff
@@ -0,0 +1,52 @@
...
+  const multipleResults = useQueries({
+    queries: contentKeys.map((contentKey) => ({
+      queryKey: learnerCreditManagementQueryKeys.catalogContainsContentItem(catalogUuid, contentKey),
+      queryFn: getCatalogContainsContentItem,
+      enabled: !!catalogUuid,
+      ...queryOptions,
+    })),
+  });
+  return {
+    data: multipleResults,
```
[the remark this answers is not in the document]
pwnage101: oh yea, that seems reasonable

src/components/learner-credit-management/data/utils.js
```diff
@@ -688,8 +688,20 @@ export const startAndEnrollBySortLogic = (prev, next) => {
...
-export const getAssignableCourseRuns = ({ courseRuns, subsidyExpirationDatetime, isLateRedemptionAllowed }) => {
-  const clonedCourseRuns = courseRuns.map(courseRun => ({
+export const getAssignableCourseRuns = ({
+  courseRuns, subsidyExpirationDatetime,
+  isLateRedemptionAllowed,
+  catalogContainsRestrictedRunsData,
+}) => {
+  // First, whittle down the runs to just the ones that are either completely
+  // unrestricted, or restricted but allowed for the current subsidy.
+  const unrestrictedCourseRuns = courseRuns.filter(courseRun => {
```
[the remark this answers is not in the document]
pwnage101: Yea, I constantly face this naming issue.  I hesitated to use "allowed" in this case because the purpose of the surrounding function is to determine which runs are "assignable" which could be conflated with "allowed".  Open to suggestions, or if you think "allowed" works for sufficient code comments.

src/components/learner-credit-management/data/hooks/useCatalogContainsContentItemsMultipleQueries.js
```diff
@@ -0,0 +1,52 @@
...
+    // This whole hook is considered to be still loading if either:
+    //   1. The input catalogUuid is falsey, implying the upstream
+    //      waterfall query to fetch the policy has not yet returned, or
+    //   2. The input contentKeys is undefined/null, implying the upstream
+    //      waterfall query to algolia has not yet returned, or
+    //   3. At least one query is still loading.
+    isLoading: (
+      !catalogUuid
+      || contentKeys === undefined
+      || contentKeys === null
```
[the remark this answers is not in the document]
pwnage101: Yes, actually I think this all makes sense.  I'm still in draft code writing phase so I'm not 100% sure if it was ever possible for the algolia response to construct a card with a null/undefined value for [coureRuns](https://github.com/openedx/frontend-app-admin-portal/blob/2c0b8685069c7d7d182dbbde3bbe446113d74bb1/src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx#L56), but that maybe seems unlikely.  In which case your suggestion for a simpler isLoading looks almost right, with one minor tweak:

```javascript
isLoading: multipleResults.length !== 0 && multipleResults.some(result => result.isLoading)
```

src/components/learner-credit-management/cards/BaseCourseCard.jsx
```diff
@@ -35,7 +35,11 @@ const BaseCourseCard = ({
     formattedPrice,
     isExecEdCourseType,
     footerText,
+    isLoadingRestrictedRuns,
   } = courseCardMetadata;
+  const runPrice = formatPrice(courseRun.contentPrice);
+  const coursePrice = isLoadingRestrictedRuns ? <Skeleton /> : formattedPrice;
+  const cardPrice = courseRun ? runPrice : coursePrice;
```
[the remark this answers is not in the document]
pwnage101: Indeed, I ran into this yesterday and fixed in the same way, but forgot to push.

src/components/learner-credit-management/cards/data/useCourseCardMetadata.jsx
pwnage101: leaving open to remind myself to implement more DRYness on filtering restricted runs.

src/components/learner-credit-management/data/utils.js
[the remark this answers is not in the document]
pwnage101: Made irrelevant after refactor that removed the variable declaration entirely.

src/components/learner-credit-management/cards/tests/CourseCard.test.jsx
```diff
@@ -20,13 +21,20 @@ import {
...
 import { getButtonElement, queryClient } from '../../../test/testUtils';
 
 import EnterpriseAccessApiService from '../../../../data/services/EnterpriseAccessApiService';
 import { BudgetDetailPageContext } from '../../BudgetDetailPageWrapper';
 import { EMAIL_ADDRESSES_INPUT_VALUE_DEBOUNCE_DELAY } from '../data';
 import { getGroupMemberEmails } from '../../data/hooks/useEnterpriseFlexGroups';
+import { ENTERPRISE_RESTRICTION_TYPE } from '../../data/constants';
+
+jest.mock('@edx/frontend-platform', () => ({
+  ensureConfig: jest.fn(),
```
adamstankiewicz, line 35: [nit/curious] Is mocking `ensureConfig` needed?

src/components/learner-credit-management/cards/tests/CourseCard.test.jsx
```diff
@@ -20,13 +21,20 @@ import {
...
+  getConfig: jest.fn(),
```
adamstankiewicz, line 36: nit: Might recommend defaulting to returning an empty object vs. undefined, should a test case in this test suite _not_ mock `getConfig` explicitly. That way, accessing properties on `getConfig` in implementation won't throw a JS error if an explicit mock happens to be forgotten.

Something like:

```jsx
jest.fn(() => ({}))
```

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
```diff
@@ -59,26 +60,31 @@ const NewAssignmentModalDropdown = ({
...
-            <Stack>
-              {intl.formatMessage(messages.startDate, {
-                startLabel: startLabel(courseRun),
-                startDate: dayjs(courseRun.start).format(SHORT_MONTH_DATE_FORMAT),
-              })}
-              <span className={classNames('small', { 'text-muted': getDropdownItemClassName(courseRun) })}>
-                {intl.formatMessage(messages.enrollBy, {
-                  enrollByDate: dayjs(courseRun.enrollBy).format(SHORT_MONTH_DATE_FORMAT),
+        {courseRuns.length > 0 && courseRuns.map(courseRun => {
+          if (isLoading) {
```
adamstankiewicz, line 64: [clarification] When the async loading is resolved, will we still have the same number of `courseRuns` we're iterating over in this code path? E.g., if the loading completes, and we determine a restricted run is NOT contained by the catalog, does the `courseRuns` list end up having 1 less run to iterate through?

If so, I wonder if it's worth keeping the `Skeleton` loading dropdown items separate from the iteration of `courseRuns.map(...)`?

For example, when `isLoading: true`, it could make sense to just default to showing ~3 skeleton dropdown items before showing all available runs once loading is complete so as not to show a loading state for a course run that would ultimately end up getting removed anyway?

adamstankiewicz (commented):
Looking good; largely just have one clarifying/curiosity question around the `Skeleton` loading states rendering before coming back to leave a 👍

brobro10000 (approved):
Approved to unblock 👍🏽 , minor nits

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
pwnage101: The way I designed it: the default number of skeleton options to show in the dropdown is always just the number of confirmed assignable runs at any given time. Before any API calls to contains_content_items resolves, the number of skeletons should be equal to the number of unrestricted runs, then as each API call for each restricted run resolves, the number of skeletons will grow (or not) until finally isLoading flips to false.  At that point, a final re-render of the component should reveal the assignable runs.

While the explanation above is more complex than just "default to 3 skeletons", the nice thing is that there's no extra code to handle the extra case.

adamstankiewicz (approved):

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
adamstankiewicz: Thanks for the explanation/clarification. Approved with the nit considerations :)

src/components/learner-credit-management/assignment-modal/NewAssignmentModalDropdown.jsx
```diff
@@ -59,26 +60,31 @@ const NewAssignmentModalDropdown = ({
...
+            return (<span className="dropdown-item" data-testid="assignment-dropdown-item"><Skeleton /></span>);
```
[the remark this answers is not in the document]
pwnage101: I originally tried that without the `className="d-block"`, but it didn't work so I gave up on Dropdown.Item.  How did you know `d-block` fixes it?

PR #1364, opened 2024-12-04
Adding people management data table 

kiram15:
Adding a card view members data table to the people management page 
Screenshot 2024-12-04 at 10 00 23 AM

[Figma](https://www.figma.com/design/eMWRd2jVa4Jqc6y7qrUfaa/Groups-Phase-2?node-id=917-50547&t=lp3VlGYjzvCBvmXC-4) | [Jira ticket 
](https://2u-internal.atlassian.net/browse/ENT-9466)

# Testing
https://localhost.stage.edx.org:1991/exec-ed-2u-integration-qa/admin/people-management
Search by name and email, look through pagination 

# For all changes


src/components/PeopleManagement/data/hooks/useEnterpriseMembersTableData.js
```diff
@@ -0,0 +1,58 @@
...
+          pageCount: data.numPages ?? Math.floor(data.count / options.pageSize),
+          results: data.results,
+        });
+      } catch (error) {
+        logError(error);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+    fetch();
```
katrinan029, line 43: Because we are displaying cards in the table, every keystroke causes the table to rerender and shifts the layout which can disrupt the user experience.
https://github.com/user-attachments/assets/470b5ca5-4118-40dd-87ff-1ad103788717

since we are already debouncing, I would suggest to also add in a minimum query string limit, let's say more than 2 characters so that we're only going to make requests when the user types in more than 2 characters. I think we can do something like this: 

```
    if (args.filters.length && args.filters[0].value.length > 2) {
      fetch()
    } else if(!args.filters.length) {
      fetch()
    }
```

katrinan029 (changes requested):
Looks good! I just have one comment about the card rendering

katrinan029 (approved):

PR #1367, opened 2024-12-12
Adding in remove capability for group members + csv download 

kiram15:
Because I'm going on Christmas break, this isn't a fully finished PR. It doesn't contain tests, nor does it have the filter functionality needed in the CSV download when a user searches a string or filters on something else. 

# For all changes


src/components/PeopleManagement/GroupDetailPage/DownloadCsvButton.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const getCsvFileName = () => {
+    const currentDate = new Date();
+    const year = currentDate.getUTCFullYear();
+    const month = currentDate.getUTCMonth() + 1;
+    const day = currentDate.getUTCDate();
+    return `${year}-${month}-${day}-group-detail-report.csv`;
+  };
+
+  const createCsvData = (jsonData) => jsonToCsv(jsonData.map(row => ({
```
marlonkeating, line 33: The issue with generating the csv file on the client right now is that it can only generate a file with the current page of group member results.  We'll want to generate the file on the backend, where we have full access to the entire data set.

marlonkeating (approved):

PR #1374, opened 2025-01-08
[ENT-9879] Validate 'ordering' parameter before sorting the data

mahamakifdar19:
**Ticket:**
https://2u-internal.atlassian.net/browse/ENT-9879

**Description:**
Added a truthy check for the ordering parameter to ensure the application does not throw runtime errors when the ordering query parameter is temporarily missing or null. 
# For all changes

### DIFF
--- a/src/data/actions/table.js
+++ b/src/data/actions/table.js
@@ -119,14 +119,17 @@ const sortTable = (tableId, fetchMethod, ordering) => (
 
     // If we can sort client-side because we have all of the data, do that
     if (tableState.data && tableState.data.num_pages === 1) {
-      const isDesc = ordering.startsWith('-');
+      const isDesc = ordering && ordering.startsWith('-');
       const orderField = isDesc ? ordering.substring(1) : ordering;
-      const result = sortBy(tableState.data.results, orderField);
 
-      return dispatch(sortSuccess(tableId, ordering, {
-        ...tableState.data,
-        results: isDesc ? result.reverse() : result,
-      }));
+      if (orderField) {
+        const result = sortBy(tableState.data.results, orderField);
+
+        return dispatch(sortSuccess(tableId, ordering, {
+          ...tableState.data,
+          results: isDesc ? result.reverse() : result,
+        }));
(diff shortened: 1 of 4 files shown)


src/data/actions/table.js
```diff
@@ -119,14 +119,17 @@ const sortTable = (tableId, fetchMethod, ordering) => (
 
     // If we can sort client-side because we have all of the data, do that
     if (tableState.data && tableState.data.num_pages === 1) {
-      const isDesc = ordering.startsWith('-');
+      const isDesc = ordering && ordering.startsWith('-');
```
jajjibhai008, line 122: We could use optional chaining (?.) while accessing the ordering property to handle potential undefined or null values safely. This will help avoid runtime errors in cases where ordering might be missing or undefined.

`const isDesc = ordering?.startsWith('-');`

mahamakifdar19: @jajjibhai008 Optional chaining `ordering?.startsWith('-')` handles `undefined` or `null` values safely, it will return `false` for an empty string, which might not be the desired behavior. 

By using `&&`, `isDesc` will only return `true` or `false` when there is a valid value in `ordering`, and `undefined` when `ordering` is `null`, `undefined`, or an empty string

jajjibhai008: shouldn't `isDesc` only be `true` or `false`? Is there any benefit in considering the `undefined` here?

mahamakifdar19: @jajjibhai008 No. The value should be true or false only when there is a valid string. If the string is empty or undefined, having `isDesc = false` can be misleading and may disrupt the overall logic. 
Please look at the code snippet below. In such cases, `orderField` would also be an empty string. Sorting should not occur if `ordering` is empty:

```javascript
const orderField = isDesc ? ordering.substring(1) : ordering;
```

jajjibhai008 (approved):
A patch coverage check is failing. Other than that LGTM

muhammad-ammar (approved):

PR #1422, opened 2025-02-10
fix: budget invite duplicate text entry bug

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-9846)

This change fixes an issue around duplicate emails being input on the Learner Credit Browse and Enroll Add Members modal.

## Testing
- Navigate to https://localhost.stage.edx.org:1991/timtam/admin/learner-credit/ecdfe8e7-94cd-461b-bd9d-205094f04f46/members
- Click 'New members'
- Verify emails can be typed in without triggering duplicate warning

# For all changes


jesperhodge (changes requested):
Great work. Works very smoothly now.

Could we add a test for this? I feel this could have ideally been caught by a test in the first place, and if we don't add a test now, we risk this breaking again in the future.

Maybe it would be a good time to have a look if there's any other aspect of this feature that's not sufficiently tested while doing that?

src/components/learner-credit-management/invite-modal/tests/InviteMemberModal.test.jsx
```diff
@@ -169,6 +169,20 @@ describe('<InviteMemberModal />', () => {
       expect(screen.getByText(`Summary (${mockLearnerEmails.length})`)).toBeInTheDocument();
     }, { timeout: EMAIL_ADDRESSES_INPUT_VALUE_DEBOUNCE_DELAY + 1000 });
   });
+  it('shows no errors for valid manual input', async () => {
```
jesperhodge, line 172: Hey as far as I can tell, this test checks that when you enter a valid email, it doesn't display an error. I guess it's a good test to have if it isn't there yet, but not quite sure how it tests the bug that was fixed here?

I noticed that when I change the code back to the previous broken state, the test still succeeds. I am feeling that that could be a bit misleading and instill false confidence.

What do you think about adding a test `should not produce a duplicate email when editing an email`? The test would enter an email address, wait for the debounce, and then edit the same line. It would then not just look for the warning but actually check if a duplicate was in fact produced. And it would be great to make sure that the test fails  until the change you made here is active, so it doesn't create false positives.

jesperhodge (changes requested):
If you don't mind, could you just have another look at the test? I left a comment there.

jesperhodge (approved):

PR #1426, opened 2025-02-13
docs: Add TypeScript typings to LmsApiService group methods

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-9742)

This change converts `LmsApiService` into a .ts file, and adds TypeScript typings to methods based around Enterprise Groups.  This change is intended as a first step towards annotating all of our api contracts with TypeScript.

# For all changes


src/types.d.ts
```diff
@@ -0,0 +1,33 @@
...
+  current_page: number,
+  start: number,
+  results: ResultItem[],
+};
+
+export type EnterpriseGroupType = 'budget' | 'flex';
+
+export type EnterpriseGroup = {
+  /* uuid of enterprise customer */
+  enterprise_customer: string,
```
adamstankiewicz, line 20: [curious/consideration] Given we typically camelCase API response data properties in service functions, e.g.:

```js
const response = await getAuthenticatedHttpClient().get(url);
return camelCaseObject(response.data);
```

Should the interface represent the API response (snake_case), or how we typically consume it within hooks/components/etc. (camelCase)? For instance, when the interface is defined with snake_case, there's a greater likelihood it could discourage the pattern of camelCasing API data to make working with the data in the frontend feel more like JavaScript (camelCase is the norm in JavaScript/TypeScript).

Worth noting that looking at the resulting types within `getEnterpriseGroup` on your current branch, the types are lost after the `camelCaseObject` (i.e., `const enterpriseGroup = camelCaseObject(response.data);` resolves to type `any`).

Would it make sense to move the `camelCaseObject` into the service function, and define the type with camelCase properties?

src/data/services/LmsApiService.ts
```diff
@@ -3,6 +3,22 @@ import { camelCaseObject } from '@edx/frontend-platform/utils';
 
 import { configuration } from '../../config';
 import generateFormattedStatusUrl from './apiServiceUtils';
+import { EnterpriseGroup, HttpResponse, Paginated } from '../../types';
+
+export type CreateEnterpriseGroupArgs = {
```
adamstankiewicz, line 8: [curious] I'm interested in your thoughts on when to use `type`  vs. `interface` for object type definitions.

src/types.d.ts
```diff
@@ -0,0 +1,33 @@
+export type HttpResponse<Payload = {}> = {
+  data: Payload,
```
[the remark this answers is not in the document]
marlonkeating: Good call out! I changed it to '{}?' to reflect that possibility (before deleting the type altogether since we can just use AxiosResponse)

src/types.d.ts
```diff
@@ -0,0 +1,33 @@
+export type HttpResponse<Payload = {}> = {
```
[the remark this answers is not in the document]
marlonkeating: Nice, I knew this had to be defined somewhere!  Thanks for pointing it out.

src/types.d.ts
adamstankiewicz: [food for thought] FWIW, while I don't think it's being used, `frontend-platform` technically supports MFEs to use custom middleware with `axios` ([ADR](https://github.com/openedx/frontend-platform/blob/master/docs/decisions/0006-middleware-support-for-http-clients.rst)), e.g. integrating the http client with `axios-case-converter`([docs](https://www.npmjs.com/package/axios-case-converter)) within the MFE's `initialize` call:

```ts
initialize({
    messages: [appMessages],
    requireAuthenticatedUser: true,
    hydrateAuthenticatedUser: true,
    authMiddleware: [axiosCaseConverter],
});
```

See frontend-platform's implementation where the middleware function (`axiosCaseConverter` in example above) is called for the `client` [here](https://github.com/openedx/frontend-platform/blob/15ef507e41127b4fd4ace5d31f7e527381678572/src/auth/AxiosJwtAuthService.js#L92-L94).

Maybe this might be an interesting topic to discuss in an upcoming Enterprise Solution Review? E.g., if we want to adopt the `axios-case-converter` auth middleware to be consistent about camelCasing API responses within our Enterprise MFEs 🤔

src/data/services/LmsApiService.ts
marlonkeating: IMO, types are for data specifications, while interfaces are for class/function contracts.  In the case of a TypeScript project that is a web app front-end (as opposed to a library or back-end api), I don't see what I consider the ideal use cases for interfaces coming up all that often.
				
Also, I find the operations available to types more generally useful, especially given the heavily composition-oriented (rather than inheritance-oriented/OOP) paradigm coming from our Javascript codebase.

src/types.d.ts
marlonkeating: Overall I would agree that camelCase/snake_case conversions ideally happen in the api libraries, and so I've made that change to the group methods and adjusted their typings accordingly.

In regards to adopting an axios plugin that automatically converts snake_case to camelCase values, I would agree it's a more elegant solution, but don't know if there's a graceful way to integrate it into our code base at this point.  The only way I can think to do it without modifying every point in the code that does camelCase conversion, is to carry around two axios client objects during the transition period: where one client does the automatic conversion and the legacy client does not.

adamstankiewicz: Agreed, adopting it gradually would be ideal, though not the worst to tackle all existing usages in one go (N = ~65 usages). Especially given doing so would probably be less overall effort than carrying around two axios clients.

That said, if the axios plugin was enabled without any changes, the `camelCaseObject` usages we have today would largely be camelCasing data that's already been camelCased, aka largely a no-op operation that doesn't necessarily break anything either (where usages could be removed incrementally, even while the axios plugin is enabled). The primary concern would be if we ever access snake_case properties prior to passing the response data to `camelCaseObject`, but I don't believe we do on a cursory search through usages.

src/data/services/LmsApiService.ts
```diff
@@ -1,8 +1,25 @@
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { camelCaseObject } from '@edx/frontend-platform/utils';
+import type { AxiosResponse } from 'axios';
 
 import { configuration } from '../../config';
 import generateFormattedStatusUrl from './apiServiceUtils';
+import { EnterpriseGroup, Paginated } from '../../types';
```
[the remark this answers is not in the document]
marlonkeating: I started out using `Types.` instead of importing, but then saw that some of the deeply nested template types were losing readability.  For instance: 
`export type EnterpriseGroupListResponse = Promise<Types.AxiosResponse<Types.Paginated<Types.EnterpriseGroup>>>;`

package.json
```diff
@@ -38,6 +38,7 @@
     "@edx/frontend-enterprise-utils": "9.1.0",
     "@edx/frontend-platform": "8.1.5",
     "@edx/openedx-atlas": "^0.6.0",
+    "@edx/typescript-config": "1.1.0",
```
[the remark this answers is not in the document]
marlonkeating: Good catch, I meant to put it there!

src/types.d.ts
```diff
@@ -0,0 +1,27 @@
+export type Paginated<ResultItem> = {
+  next: string?,
+  previous: string?,
+  count: number,
+  current_page: number,
```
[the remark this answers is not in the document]
marlonkeating: Good call-out!  I was not aware.

tsconfig.json
```diff
@@ -1,26 +1,8 @@
...
-    "noImplicitThis": true,
-    "noUnusedParameters": true,
-    "skipLibCheck": true,
-    "sourceMap": true,
-    "strict": true,
-    "strictFunctionTypes": false,
-    "target": "ES6",
     "rootDir": ".",
-    "outDir": "dist"
+    "outDir": "dist",
```
adamstankiewicz, line 5: nit: does JSON support trailing commas on the final property defined in object? it doesn't seem my IDE is complaining about the trailing comma, but I don't think it's technically valid JSON 🤷 

image

adamstankiewicz (approved):

PR #1441, opened 2025-03-11
fix: block custom theme creation with blank colors

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-9675)

This change fixes an issue where a custom enterprise theme could be created with blank colors, which would lead to an error when selecting the blank theme.  Now, the modal will require all colors to be filled out before it will allow the theme to be created.

As part of this change, the `CustomThemeModal` component was refactored to reduce code duplication.

## Testing
- Navigate to https://localhost.stage.edx.org:1991/alc-general/admin/settings/appearance
- Click Create a custom theme
- Verify that all fields must be filled with valid hex values before 'Add theme' is clickable

# For all changes


jesperhodge (approved):
Nit: the ColorEntryField would be nice to have as separate reusable component that might be used elsewhere in the future. That might be a good opportunity to turn that component with its props into TypeScript.

PR #1442, opened 2025-03-11
feat: introduce `withAlgoliaSearch` and `useAlgoliaSearch`

adamstankiewicz:
[ENT-10080](https://2u-internal.atlassian.net/browse/ENT-10080)

Introduces `withAlgoliaSearch` Higher-Order-Component (HOC) for consuming components to get access to the Algolia search client to serve search results.

The `withAlgoliaSearch` component is connected to the Redux store to get access to the feature flag and the enterprise customer UUID. The rest of the logic is encapsulated within `useAlgoliaSearch`.

Based on the `enterpriseFeatures.catalogQuerySearchFiltersEnabled`, `useAlgoliaSearch` will conditionally make an API call to `<enterprise-catalog>/api/v1/enterprise-customer/<enterprise_id>/secured-algolia-api-key/` to generate a secured Algolia API key and get the mappings of catalog uuids <> catalog query UUIDs for further filtering.

If the feature flag is enabled and a secured Algolia API key was returned, use it to initialize the Algolia `searchClient`. Otherwise, fallback to continuing to use the generic (unscoped) `searchClient`, requiring customer-specific attributes to be explicitly passed as filters (undesired).

See example reference for its usage: https://github.com/openedx/frontend-app-admin-portal/pull/1443

# For all changes

- [ ] Ensure adequate tests are in place (or reviewed existing tests cover changes)

# Only if submitting a visual change

- [ ] Ensure to attach screenshots
- [ ] Ensure to have UX team confirm screenshots

PR #1443, opened 2025-03-11
refactor: migrate subscriptions bulk enrollment and highlights to `withAlgoliaSearch`

adamstankiewicz:
Note: Depends on https://github.com/openedx/frontend-app-admin-portal/pull/1442

Refactors two features to utilize the `withAlgoliaSearch` HOC introduced in the above linked, related PR instead of generic Algolia search API keys; instead, when the Waffle-based feature flag is enabled for a learner, an API request will be made to enterprise-catalog's `secured-algolia-api-key` API endpoint to retrieve a scoped /secured Algolia key that bakes in filters for the customer's `enterprise_catalog_query_uuids` and exposes a `catalogUuidsToCatalogQueryUuids` mapping.

The two refactored features in this PR to use the new Algolia filtering approach include:
* Subscriptions Bulk Enrollment
* Content Highlights

The same number of search results are returned in each case regardless of using the old filtering with customer-specific attributes vs. the new approach via secured Algolia API key with catalog query identifiers.

## Subscriptions Bulk Enrollment

Using the secured Algolia API key, filtering by the subscription's catalog query returns stage search results as expected.

image

## Content Highlights

Using the secured Algolia API key, no additional filters (including the existing videos exclusion) needed beyond the scoped API key, which itself filters on catalog query uuids. The highlights "Add" modal correctly shows stage search results as expected.

image

# For all changes


src/components/BulkEnrollmentPage/stepper/AddCoursesStep.tsx
```diff
@@ -0,0 +1,171 @@
...
+  return (
+    <BaseAddCoursesStepContents>
+      <SearchEnabled
+        selectedCourses={selectedCourses}
+        algolia={algolia}
+        subscription={subscription}
+      />
+    </BaseAddCoursesStepContents>
+  );
+};
```
brobro10000, line 169: [nit] This could be simplified a bit

``` jsx
  if (algolia.isCatalogQueryFiltersEnabled && algolia.isLoading) {
      return (
        <BaseAddCoursesStepContents>
          <Skeleton height={360} />
          <div data-testid="skeleton-algolia-loading-courses" className="sr-only">Loading courses...</div>
        </BaseAddCoursesStepContents>
      );
  }

  return (
    <BaseAddCoursesStepContents>
      <SearchEnabled
        selectedCourses={selectedCourses}
        algolia={algolia}
        subscription={subscription}
      />
    </BaseAddCoursesStepContents>
  );
```

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.tsx
```diff
@@ -85,15 +95,29 @@ export const SelectedContent = ({ enterpriseId }) => {
...
+    if (!hasSecuredAlgoliaApiKey) {
+      filterString = `enterprise_customer_uuids:${ENABLE_TESTING(enterpriseId)}`;
+    }
     // import testEnterpriseId from the existing ../data/constants folder and replace with enterpriseId to test locally
-    let filterString = `enterprise_customer_uuids:${ENABLE_TESTING(enterpriseId)}`;
     if (currentSelectedRowIds.length > 0) {
-      filterString += ' AND (';
+      if (filterString) {
+        filterString += ' AND ';
+      }
```
brobro10000, line 119: [curious] Should this filter string check be within the first conditional that encompasses the query were building here as opposed to just the ` AND  ` string? 
``` js
if (filterString && currentSelectedRowIds.length > 0) {
...
}
```

If `!!filterString` is false, the query returned could be `( selectedId1 OR selectedId2)` 

In that case should we just return the empty filter string?

brobro10000 (approved):
Approved to unblock. Great job tackling the TS side of things here. 2 conditional related questions in this PR.

src/components/BulkEnrollmentPage/stepper/AddCoursesStep.tsx
adamstankiewicz: good catch! will update

src/components/ContentHighlights/HighlightStepper/HighlightStepperConfirmContent.tsx
brobro10000: After async conversation, the return is what we expect with `( selectedId1 OR selectedId2 )` where the base filter is encoded into the algolia search key itself. The updated logic is in support of both an encoded algolia search key format and legacy.

adamstankiewicz:
Closing and re-opening in an attempt to get the `openedx/cla` CI check to run.

PR #1453, opened 2025-03-20
chore: migrate deprecated Table to DataTable for LearnerActivityTable

jajjibhai008:
# For all changes


src/components/LearnerActivityTable/index.jsx
```diff
@@ -1,152 +1,200 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [id, hasData]);
+
+  // Wrap fetchCourseEnrollments to update the URL when pagination changes
+  const fetchTableData = useCallback((tableState) => {
+    const newPageIndex = tableState.pageIndex;
+    const newPageForUrl = newPageIndex + 1; // Convert zero-based index to one-based for URL
+
+    // Update URL with new page number
+    const newQueryParams = new URLSearchParams(location.search);
```
mahamakifdar19, line 62: Consider handling the URL query parameter extraction into a utility function for reusability?

src/components/LearnerActivityTable/index.jsx
```diff
@@ -1,152 +1,200 @@
...
+      FilterStatusComponent={FilterStatus}
+      defaultColumnValues={{ Filter: TextFilter }}
+      columns={columns}
+      initialState={{
+        pageIndex: currentPageFromUrl, // Use page from URL
+        pageSize: PAGE_SIZE,
+        sortBy: [{ id: 'lastActivityDate', desc: true }],
+        selectedRowsOrdered: [],
+      }}
+      fetchData={fetchTableData}
```
mahamakifdar19, line 182: What is the difference between the `courseEnrollments.results` and `fetchTableData`? Do they both represent different data?

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }, [enterpriseId, tableId]);
+
+  const debouncedFetchCourseEnrollments = useMemo(
+    () => debounce(fetchCourseEnrollments, 300),
+    [fetchCourseEnrollments],
+  );
+
+  // Add a computed property to check if data exists
+  const hasData = useMemo(
```
mahamakifdar19, line 93: `hasData` could be extracted into a utility method for reusability across different tables?

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,106 @@
...
+      applySortByToOptions(args.sortBy, options);
+
+      const response = await EnterpriseDataApiService.fetchCourseEnrollments(enterpriseId, options);
+      const data = camelCaseObject(response.data);
+      setCourseEnrollments({
+        itemCount: data.count,
+        pageCount: data.numPages ?? Math.floor(data.count / options.pageSize),
+        results: data.results,
+      });
+      if (shouldTrackFetchEvents.current) {
```
mahamakifdar19, line 64: We might need this logic to check whether tracking events should be sent after an API query in other tables as well?
We can move the logic into a helper function.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,106 @@
...
+    courseListPrice: { key: 'course_list_price' },
+    courseStartDate: { key: 'course_start_date' },
+    courseEndDate: { key: 'course_end_date' },
+    passedDate: { key: 'passed_date' },
+    currentGrade: { key: 'current_grade' },
+    progressStatus: { key: 'progress_status' },
+    lastActivityDate: { key: 'last_activity_date' },
+  };
+  const orderingStrings = sortBy.map(({ id, desc }) => {
+    const apiFieldForColumnAccessor = apiFieldsForColumnAccessor[id];
```
mahamakifdar19, line 27: We could rename the constant to avoid duplication i.e:

`const apiFieldKey = apiFieldsForColumnAccessor[id];`

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,106 @@
...
+import { sendEnterpriseTrackEvent } from '@edx/frontend-enterprise-utils';
+import EnterpriseDataApiService from '../../../../data/services/EnterpriseDataApiService';
+import EVENT_NAMES from '../../../../eventTracking';
+
+const applySortByToOptions = (sortBy, options) => {
+  if (!sortBy || sortBy.length === 0) {
+    return;
+  }
+  const apiFieldsForColumnAccessor = {
+    userEmail: { key: 'user_email' },
```
mahamakifdar19, line 16: We can store the key directly to simply it
 `userEmail: 'user_email',`

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,106 @@
...
+    courseTitle: { key: 'course_title' },
+    courseListPrice: { key: 'course_list_price' },
+    courseStartDate: { key: 'course_start_date' },
+    courseEndDate: { key: 'course_end_date' },
+    passedDate: { key: 'passed_date' },
+    currentGrade: { key: 'current_grade' },
+    progressStatus: { key: 'progress_status' },
+    lastActivityDate: { key: 'last_activity_date' },
+  };
+  const orderingStrings = sortBy.map(({ id, desc }) => {
```
mahamakifdar19, line 26: Instead of using `map()` to create an array and then filtering out undefined values, we simply use `forEach()` to directly populate the `orderingStrings` array, skipping any invalid entries.

```
 sortBy.forEach(({ id, desc }) => {
    const apiFieldKey = apiFieldsForColumnAccessor[id];
    if (apiFieldKey) {
      orderingStrings.push(desc ? `-${apiFieldKey}` : apiFieldKey);
    }
  });
```


```
if (orderingStrings.length > 0) {
    options.ordering = orderingStrings.join(',');
  }
```

src/components/LearnerActivityTable/index.jsx
```diff
@@ -1,152 +1,200 @@
...
+import { formatPrice } from '../learner-credit-management/data';
+import { useTableData } from '../Admin/TableDataContext';
+
+const FilterStatus = (rest) => <DataTable.FilterStatus showFilteredFields={false} {...rest} />;
+
+const UserEmail = ({ row }) => (
+  <span data-hj-suppress>{row.original.userEmail}</span>
+);
+
+UserEmail.propTypes = {
```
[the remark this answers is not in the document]
jajjibhai008: I think it will depend use case to use case. If someone is supposed to use it again then he/she can convert it into an abstract component. But for now it has single usage here.

src/components/LearnerActivityTable/index.jsx
jajjibhai008: For now, I am not sure if the same thing is supposed to be used in other places or not that's why I haven't abstracted the code too much. I try to make it simple and more readable.

src/components/LearnerActivityTable/index.jsx
jajjibhai008: `courseEnrollments.results` is the actual formatted data that we are supposed to pass the DataTable Component and `fetchTableData` is the reference of the data fetching function that we are passing to our hook.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: Maybe in the future, we may need to add other fields like labels, etc, that's why I keep it so.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: The map/filter pattern is more functional and aligns with modern JS practices. That's why I prefer map over ForEach.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: This was specifically to resolve the Downloading button rendering issue. No need to make it abstract.

mahamakifdar19: @jajjibhai008 The function is primarily checking if the given array contains data and this check could be a common requirement for other tables as well.

mahamakifdar19 (approved):

src/components/LearnerActivityTable/LearnerActivityTable.test.jsx
```diff
@@ -107,33 +36,54 @@ const LearnerActivityTableWrapper = props => (
...
+          pageCount: 0,
+          results: [],
+        },
+        fetchCourseEnrollments: jest.fn(),
+      },
+    );
     const tree = renderer
       .create((
-        <LearnerActivityEmptyTableWrapper id="active-week" activity="active_past_week" />
+        <LearnerActivityTableWrapper id="active-week" activity="active_past_week" />
```
brobro10000, line 86: [suggestion] We want to try and move away from using snapshots as a form of testing. These should be refactored to no longer generate snapshots.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,88 @@
...
+        eventName: EVENT_NAMES.PROGRESS_REPORT.DATATABLE_SORT_BY_OR_FILTER,
+        tableId,
+        options,
+      });
+    } catch (error) {
+      logError(error);
+    } finally {
+      setIsLoading(false);
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
brobro10000, line 66: [nit] What dependencies are being left out that requires overriding the linter?

src/components/LearnerActivityTable/data/utils.js
```diff
@@ -0,0 +1,43 @@
+import { sendEnterpriseTrackEvent } from '@edx/frontend-enterprise-utils';
+
+export const trackDataTableEvent = ({
```
brobro10000, line 3: [nit] Coverage is a bit low for this function. A individual unit test, or integration test with the hook using it would help increase the coverage and validate behavior.

Including it in the integration test in the learner activity tables a second benefit of producing a test that can be validated across other Datatable integration tests as well.

src/components/LearnerActivityTable/data/utils.js
```diff
@@ -0,0 +1,43 @@
...
+  } else {
+    // set to true to enable tracking events on future API queries
+    // eslint-disable-next-line no-param-reassign
+    shouldTrackRef.current = true;
+  }
+
+  return shouldTrackRef.current;
+};
+
+export const updateUrlWithPageNumber = (pageNumber, location, navigate, replace = true) => {
```
brobro10000, line 30: [nit] Similar to previous comment, create a seperate unit test with a preference for integration test to create a reproducible test across all Datatable components using this function

src/components/LearnerActivityTable/index.jsx
```diff
@@ -1,152 +1,200 @@
...
+  row: PropTypes.shape({
+    original: PropTypes.shape({
+      userEmail: PropTypes.string,
+    }).isRequired,
+  }).isRequired,
+};
+
+const LearnerActivityTable = ({ id, enterpriseId, activity }) => {
```
brobro10000, line 34: [suggestion] Given a unique id is available via the top level component rendering each individual table, (along with the custom search filter component, `AdminSearchForm`) in we can utilize this value to start prepending ur relevant filters and page information with the identifier with this value.

This will allow the URL navigation to display its corresponding table, and update the filter and page number with the correct information from the API.

src/components/LearnerActivityTable/index.jsx
```diff
@@ -1,152 +1,200 @@
...
+  }, [id, hasData]);
+
+  // Wrap fetchCourseEnrollments to update the URL when pagination changes
+  const fetchTableData = useCallback((tableState) => {
+    const newPageForUrl = tableState.pageIndex + 1; // Convert zero-based index to one-based for URL
+
+    updateUrlWithPageNumber(newPageForUrl, location, navigate);
+
+    return fetchCourseEnrollments(tableState);
+  }, [fetchCourseEnrollments]);
```
brobro10000, line 76: [curious] Could this be collocated within `useCouresEnrollments`? The learner allocation table uses a similarly constructed [useBudgetRedemptions](https://github.com/openedx/frontend-app-admin-portal/blob/master/src/components/learner-credit-management/data/hooks/useBudgetRedemptions.js#L49) hook to populate data and retrieve access to the table metadata in the top level hook. 

This pattern allows us to parse fields from the datatable (filters and sortBy) directly within the hook and use it as needed (track event metadata, logging errors/infos, possibly url generation).

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,88 @@
...
+
+      trackDataTableEvent({
+        shouldTrackRef: shouldTrackFetchEvents,
+        enterpriseId,
+        eventName: EVENT_NAMES.PROGRESS_REPORT.DATATABLE_SORT_BY_OR_FILTER,
+        tableId,
+        options,
+      });
+    } catch (error) {
+      logError(error);
```
brobro10000, line 62: [suggestion] We have an opportunity to capture much more actionable information here such as the state of the filters, page number, table id etc. I would update the log error to try and capture as much of that as possible in a way that if something were to go wrong with the response, we can understand the table state directly from the log error, vs parsing Datadogs RUM section which tends to obfuscates in a way that debugging becomes cumbersome.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
```diff
@@ -0,0 +1,88 @@
...
+  useCallback, useMemo, useRef, useState,
+} from 'react';
+import { camelCaseObject } from '@edx/frontend-platform/utils';
+import debounce from 'lodash.debounce';
+import { logError } from '@edx/frontend-platform/logging';
+import EnterpriseDataApiService from '../../../../data/services/EnterpriseDataApiService';
+import EVENT_NAMES from '../../../../eventTracking';
+import { trackDataTableEvent } from '../utils';
+
+const applySortByToOptions = (sortBy, apiFieldsForColumnAccessor, options) => {
```
brobro10000, line 11: 🥳  Nice job bringing this in.  Could this be the beginnings of a `useDatatable` hook with this simple function?

brobro10000 (commented):
Great job with implementation. I'd implore you to give the `useDatatable` hook an attempt with at least generating the necessary datastructures for sorting an filtering to be parsed by an API outside of `useDatatable`.

Some nits related to testing and implementation.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: `apiFieldsForColumnAccessor` is being used inside that hook, but we don't need to make it a dependency because it's going to be static, and when I added it to the dependency array, it created an unnecessary rendering issue.

src/components/LearnerActivityTable/data/utils.js
jajjibhai008: `trackDataTableEvent` and `updateUrlWithPageNumber` are only being called (without using their return values) within the hook and component, respectively. Therefore, I will write a separate test for them in the utils.test.js file.

src/components/LearnerActivityTable/data/utils.js
jajjibhai008: `trackDataTableEvent` and `updateUrlWithPageNumber` are only being called (without using their return values) within the hook and component, respectively. Therefore, I will write a separate test for them in the utils.test.js file.

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: It's a good idea for better debugging. I have added this necessary info inside the logError.
```javascript
logError(error, {
        tableState: {
          tableId,
          enterpriseId,
          filters: args.filters || 'none',
          sortBy: JSON.stringify(args.sortBy || []),
        },
        message: `Error fetching course enrollments for table ${tableId}`,
      });
```

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: Thanks.
Yes, I initially implemented the necessary functionalities, and this can be updated over time as more requirements arise.

src/components/LearnerActivityTable/index.jsx
jajjibhai008: Are you suggesting adding tableId + pageNumber so that we can use a consistent structure for identifying table states across different tables while navigating?

brobro10000: Yes exactly 👍🏽

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
brobro10000: We should still be attempting to get to the root cause of the re-rendering issue and resolve it by memoizing the values.

src/components/LearnerActivityTable/index.jsx
brobro10000: I can consider this a non-blocking nit but something to hopefully strive for as an in future iterations down the road.

src/components/LearnerActivityTable/index.jsx
jajjibhai008: I have implemented this change and now URL will look like 
http://localhost:1991/test-enterprise/admin/learners/learners-inactive-week?learners-inactive-week_page=2

src/components/LearnerActivityTable/data/hooks/useCourseEnrollments.js
jajjibhai008: fixed.

brobro10000 (approved):

PR #1455, opened 2025-03-20
fix: Email Matching in CSV upload should be case-insensitive

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-10195)

This fixes the issue when creating/adding users to a group via csv upload and getting an error when the email addresses differ from the org registered email in casing alone.

## Testing Instructions
- Navigate to https://localhost.stage.edx.org:1991/alc-general/admin/people-management
- Click 'Create group'
- Create a csv file containing the email of an org member with casing that differs from the People table
- Verify the uploaded user can be added, without any error messages indicating they aren't part of the org

# For all changes

### DIFF
--- a/src/components/learner-credit-management/cards/data/utils.ts
+++ b/src/components/learner-credit-management/cards/data/utils.ts
@@ -147,7 +147,7 @@ export const isInviteEmailAddressesInputValueValid = ({
 }: LearnerEmailsValidityArgs): LearnerEmailsValidityReport => {
   let validationError;
   const learnerEmailsCount = learnerEmails.length;
-  const lowerCasedEmails: string[] = [];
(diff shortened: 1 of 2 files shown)


src/components/learner-credit-management/cards/data/utils.ts
```diff
@@ -162,7 +162,7 @@ export const isInviteEmailAddressesInputValueValid = ({
       // Check for duplicates (case-insensitive)
```
kiram15, line 162: nit: It shouldn't be called lowerCasedEmails, because it seems like we're manually setting them to lowercase from that list

kiram15 (approved):
A single nit but looks good otherwise!

PR #1457, opened 2025-03-24
fix: upgrade paragon to support datatable fix

brobro10000:
In the current state of Datatable components with the `isSelectable` attribute enabled along with attribute `SelectionStatusComponent={DataTable.ControlledSelectionStatus}` there is a bug that when a table is filtered, all the items of the filtered component is selected, and the filter removed, the table selects all components from the unfiltered state despite only being selected in the filtered state.

This is due to the field `isEntireTableSelected` being set to true when a datatable is filtered and all fields are selected in the filtered state, but retains that value after removing the filter. 

This PR updates Paragon to `v22.16.1` which contains the [fix for this bug](https://github.com/openedx/paragon/pull/3477). 

https://github.com/user-attachments/assets/844324e1-56ec-411d-9274-5914c95b0879

# For all changes


src/components/ForbiddenPage/ForbiddenPage.test.jsx
```diff
@@ -1,15 +1,18 @@
...
+    render(
+      <IntlProvider locale="en">
         <ForbiddenPage />
-      ))
-      .toJSON();
-    expect(tree).toMatchSnapshot();
+      </IntlProvider>,
+    );
+    expect(screen.getByText('403')).toBeInTheDocument();
+    expect(screen.getByText('You do not have access to this page.')).toBeInTheDocument();
```
adamstankiewicz, line 16: [kudos] thanks for migrating to explicit assertions vs. snapshot

src/components/settings/SettingsAppearanceTab/index.tsx
```diff
@@ -168,11 +181,7 @@ export const SettingsAppearanceTab = ({
         onClose={() => setConfigChangeSuccess(null)}
         show={configChangeSuccess || false}
       >
-        <FormattedMessage
-          id="adminPortal.settings.portalAppearanceTab.successMessage"
-          defaultMessage="Portal appearance updated successfully."
-          description="Success message for updating portal appearance."
-        />
+        {updatedAppearanceToastText}
```
adamstankiewicz, line 184: [kudos] thanks for addressing this!

src/components/settings/SettingsSSOTab/steps/SSOConfigConfigureStep.test.jsx
```diff
@@ -16,26 +17,34 @@ const INITIAL_SSO_STATE = {
   },
 };
 
+const SSOConfigConfigureStepWraper = ({ store = defaultStore, initalState = INITIAL_SSO_STATE, children }) => (
```
adamstankiewicz, line 20: nit: typos

```suggestion
const SSOConfigConfigureStepWrapper = ({ store = defaultStore, initialState = INITIAL_SSO_STATE, children }) => (
```

adamstankiewicz (approved):
LGTM, with a typo nit!

PR #1464, opened 2025-03-26
fix: add/remove group member with non-lowercase email

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-10244)

Fix for issue where attempting to click and add a member whose email address does not have all lowercase letters will result in adding the user to the summary, but:
- The checkbox is not checked
- Clicking checkbox again will add the user multiple times, creating duplicate user errors instead of removing the user

## Testing Instructions
- Navigate to https://localhost.stage.edx.org:1991/aj-test-co/admin/people-management
- Click 'Create group'
- Navigate to second page and you will find user that has a capitalized letter in their email address
- Click this user's checkbox and verify their (lowercased) email is added to the summary
- Click user's checkbox again and verify their email is removed from the summary

# For all changes

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -690,7 +690,7 @@ function removeStringsFromListCaseInsensitive(list, stringsToRemove) {
   // Create lookup of lowercased -> original strings
   const lowercaseLookup = fromPairs(lowercasePairs);
   // Use lowercased list with lowercased removal strings to perform an efficient set difference operation
-  const lowercaseList = list.map(pair => pair[0]);
+  const lowercaseList = lowercasePairs.map(pair => pair[0]);
   const lowercaseToRemove = stringsToRemove.map(str => str.toLowerCase());
   const removedLowercase = removeStringsFromList(lowercaseList, lowercaseToRemove);
   // Return set difference mapped back to original strings
--- a/src/utils.js
+++ b/src/utils.js
@@ -6,6 +6,7 @@ import isArray from 'lodash/isArray';
 import mergeWith from 'lodash/mergeWith';
 import isNumber from 'lodash/isNumber';
 import isString from 'lodash/isString';
+import fromPairs from 'lodash/fromPairs';
 import isEmail from 'validator/lib/isEmail';
 import isEmpty from 'validator/lib/isEmpty';
 import isNumeric from 'validator/lib/isNumeric';
@@ -663,7 +664,7 @@ function splitAndTrim(separator, str) {
 }
 
 /**
- *
+ * Remove strings from a list, matching in a case-sensitive manner
  * @param {Array<string>} list
  *  List of strings to remove from
  * @param {Array<string>} stringsToRemove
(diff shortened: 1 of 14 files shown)


src/utils.js
```diff
@@ -675,6 +676,22 @@ function removeStringsFromList(list, stringsToRemove) {
...
+ * @param {Array<string>} stringsToRemove
+ *  Strings that should be removed from list
+ * @returns {Array<string>}
+ */
+function removeStringsFromListCaseInsensitive(list, stringsToRemove) {
+  const lowercaseMap = fromPairs(list.map(str => [str.toLowerCase(), str]));
+  const lowercaseList = list.map(str => str.toLowerCase());
+  const lowercaseToRemove = stringsToRemove.map(str => str.toLowerCase());
+  const removedLowercase = removeStringsFromList(lowercaseList, lowercaseToRemove);
+  return removedLowercase.map(str => lowercaseMap[str]);
```
jesperhodge, line 692: Good solution. Do you think we could simplify it even more for readability?
Maybe something similar to this?
```
const lowerCaseRemoveList = stringsToRemove.map(str => str.toLowerCase());
const result = list.filter(str => !lowerCaseRemoveList.includes(str.toLowerCase());
return result;
```

jesperhodge (approved):
Nice work! I suggested a little simplification in one case, but aside from that I'll approve it. As long as the tests run after adjusting that, you can merge it.

src/utils.js
marlonkeating: That is a more concise and readable implementation.  However, its performance isn't as good for large datasets, being approximately O(n^2) as opposed to the current O(n).

I did go ahead and add comments on the current implementation, which should help future readers understand what's going on.

jesperhodge: Understanding where this is used, that makes sense.

src/utils.js
```diff
@@ -675,6 +676,27 @@ function removeStringsFromList(list, stringsToRemove) {
...
+  // Create lookup of lowercased -> original strings
+  const lowercaseLookup = fromPairs(lowercasePairs);
+  // Use lowercased list with lowercased removal strings to perform an efficient set difference operation
+  const lowercaseList = lowercasePairs.map(pair => pair[0]);
+  const lowercaseToRemove = stringsToRemove.map(str => str.toLowerCase());
+  const removedLowercase = removeStringsFromList(lowercaseList, lowercaseToRemove);
+  // Return set difference mapped back to original strings
+  return removedLowercase.map(str => lowercaseLookup[str]);
+}
+
```
jesperhodge, line 699: Following up from our conversation in slack:
Since performance is a concern, to get the best result, it would be great to have a look if some existing options like lodash's `_.without()` is more performance-optimized than we are able to do here.

Then, it's hard to figure out every detail of this code, maybe I'm missing something. But it seems that it would be more readable and performant if we simply lowercase the `stringsToRemove` here once like this:

```
function removeStringsFromListCaseInsensitive(list, stringsToRemove) {
  return removeStringsFromList(
    list,
    stringsToRemove.map(str => str.toLowerCase()),
    { caseInsensitive: true }
  )
}
```

That would get rid of a number of iterations over the list.
Then, assuming we are not using a lodash method or anything, in the called function, you could do:

```
function removeStringsFromList(list, stringsToRemove, options = { caseInsensitive: false }) {
  const removalSet = new Set([...stringsToRemove]);
  if (options.caseInsensitive) {
    return return list.filter((item) => !removalSet.has(item.toLowerCase()));
  }

  return list.filter((item) => !removalSet.has(item));
}
```

jesperhodge (changes requested):
What do you think about the followup comment I left?

src/utils.js
marlonkeating: I went ahead and switched to lodash `_.without()` after looking at its [implementation](https://github.com/lodash/lodash/blob/8a26eb42adb303f4adc7ef56e300f14c5992aa68/dist/lodash.js#L2806) to confirm it's doing much the same thing that `removeStringsFromList` originally did (just using object lookups instead of `Set` membership tests, both being hash-type O(1) lookups).

But as long as we are wanting to do case-insensitive matching between two arbitrary string lists, there's no getting around lowercasing both sets of strings.

jesperhodge (approved):

PR #1470, opened 2025-04-09
[ENT-10167] Added DataTable for EnrolledLearnersForInactiveCoursesTable

mahamakifdar19:
**Ticket:** 
https://2u-internal.atlassian.net/browse/ENT-10167

**Description**
Migrated to `DataTable` for `EnrolledLearnersForInactiveCoursesTable`

# For all changes


src/components/EnrolledLearnersForInactiveCoursesTable/data/hooks/useCourseUsers.js
```diff
@@ -0,0 +1,95 @@
...
+  useCallback, useMemo, useRef, useState,
+} from 'react';
+import { camelCaseObject } from '@edx/frontend-platform/utils';
+import debounce from 'lodash.debounce';
+import { logError } from '@edx/frontend-platform/logging';
+import EnterpriseDataApiService from '../../../../data/services/EnterpriseDataApiService';
+import EVENT_NAMES from '../../../../eventTracking';
+import { trackDataTableEvent } from '../../../LearnerActivityTable/data/utils';
+
+const applySortByToOptions = (sortBy, apiFieldsForColumnAccessor, options) => {
```
jajjibhai008, line 11: This code is being used in two hooks, so I think we can move it to some common place by following the DRY pattern.

src/components/EnrolledLearnersForInactiveCoursesTable/data/hooks/useCourseUsers.js
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      setIsLoading(true);
+      const options = {
+        page: args.pageIndex + 1,
+        pageSize: args.pageSize,
+      };
+      applySortByToOptions(args.sortBy, apiFieldsForColumnAccessor, options);
+
+      const response = await EnterpriseDataApiService.fetchEnrolledLearnersForInactiveCourses(enterpriseId, options);
+      const data = camelCaseObject(response.data);
```
jajjibhai008, line 47: We only have difference of this endpoint between useCoursesUsers and useCourseEnrollments. So I think we can merge these two hooks into a single one by adding an extra parameter of the endpoint. This will reduce the code and put common functionality at one place.

src/components/EnrolledLearnersForInactiveCoursesTable/data/tests/constants.js
```diff
@@ -0,0 +1,38 @@
+const mockUseCourseEnrollments = {
```
jajjibhai008, line 1: I already created these mock enrollments at src/components/LearnerActivityTable/data/tests/constants.js. As these are common so I think we can move it to some parent common place. Currently, the same code is being duplicated.

jajjibhai008 (changes requested):
By this way, Download CSV button will be disable. So, we need to use the saved context value hasData to enable it. You need to do necessary changes in this file src/components/Admin/index.jsx.

jajjibhai008 (approved):
LGTM 👍

PR #1471, opened 2025-04-09
Added new LPR page with recent design changes behind a feature flag

omar-sarfraz:
**Ticket:** [ENT-10240](https://2u-internal.atlassian.net/browse/ENT-10240)
### Description: 
Modified the design and layout according to new Figma design. Added drag and drop functionality but disabled it in current PR. Here is the list of all the changes.
- Created a new page behind feature flag. 
- Changed design and added gray background.
- Minor language changes in LPR.
- Changed placement of AI summary card.
- Changed button in module activity to outline primary.
- Implemented drag n drop and disabled it.
- Saved the components' order in local storage.
- Removed the Embedded Manage Learner section

### Review Guidelines:
I've duplicated the existing Admin folder to create a new folder AdminV2. This means that majority of the files are duplicates while few are new and some files are modified. Ignore the test files for now. I've not updated those, I'm still working on test cases. Here is a list of files on which the review should be focused on:
- AdminV2/AIAnalyticsSummary.jsx: Changed the layout and placed overview heading inside the component to match the design.
- AdminV2/AIAnalyticsSummarySkeleton.jsx: Rendered overview heading so that we can see it while skeleton is loading.
- AdminV2/AnalyticsOverview.jsx: New component which contains AI summary and overview cards.
- AdminV2/LearnerReport.jsx: New component which contains the full report table and module activity.
- AdminV2/SortableItem.jsx: New component which is used for drag n drop. It also applies background color to draggable items.
- AdminV2/index.jsx: Contains logic for drag n drop.
- AdminV2/tabs/DownloadCSVButton.jsx: Changed variant of button to outline-primary.
- EnterpriseApp/EnterpriseAppRoutes.jsx: Added a new route for AdminV2.
- EnterpriseApp/data/constants.js: Added a constant for new route.
- config/index.js: Added feature flag.
- containers/AdminPageV2/index.jsx: Used AdminV2 in AdminContainer instead of previous page.
- src/utils.js: Added 2 utility functions to save and retrieve a values from local storage.

image

# For all changes


adamstankiewicz (commented):
Left a few initial comments, but a couple additional things worth mentioning:

1. I think it may be worth moving away from snapshots for the `Admin` component (the file with 25,000+ line additions), and possibly others, as large snapshot files are not all that useful in practice; it's difficult to parse and understand when regressions occur. For larger components like `Admin`, it's generally preferable these days (especially with React Testing Library) to rely on explicit assertions of what you expect to see in the UI vs. ensuring a snapshot didn't change incorrectly). We've generally moved away from snapshot testing in the Enterprise MFEs for net-new stuff.
2. It appears that the LPR feature has recently doubled down on Redux usage, which is something this MFE (and [other Enterprise MFEs](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/master/docs/decisions/0005-moving-away-redux-in-learner-portal.rst)) is trying to move away from long-term in favor of other means of client state management and using React Query for server state. See the following related ADRs in the repo [here](https://github.com/openedx/frontend-app-admin-portal/blob/master/docs/decisions/0006-tanstack-react-query.rst) and [here](https://github.com/openedx/frontend-app-admin-portal/blob/master/docs/decisions/0008-application_state_differenciation.rst), as well as this Open edX OEP [here](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0067-bp-tools-and-technology.html) and [here](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0067/decisions/frontend/0010-react-query.html). We'd like to eventually remove Redux entirely from this MFE, so **I might recommend starting to think how LPR could be replaced with React Query for managing API-derived data and vanilla React state management for client-only data**, where possible; if state management for LPR is too complex for vanilla React statement, there are better, lighter weight alternatives beyond Redux these days (e.g., [`zustand`](https://zustand.docs.pmnd.rs/getting-started/introduction)).
  * That said, this higher level feedback around pulling back out Redux usage is not blocking for this specific PR's LPR redesign, as the additional LPR Redux usage was already added via earlier, recent contributions. Just want to call out that we'll be moving away from Redux in this code area long-term, so I would recommend considering to start making incremental efforts towards that, where appropriate.

src/components/AdminV2/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,140 @@
...
+        >
+          <>
+            <AutoFixHigh className="mr-2" />
+            <FormattedMessage id="adminPortal.v2.summarizeAnalytics" defaultMessage="Summarize analytics" />
+          </>
+        </Button>
+        {/* Track Progress is currently hidden due to data inconsistency. It will be addressed as part of ENT-7812 */}
+        <Button
+          variant="outline-primary"
+          className="d-none"
```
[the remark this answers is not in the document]
omar-sarfraz: I believe someone added `d-none` class as a temporary solution and didn't get back to it. I'll follow up with the team to inquire about what needs to be done about this button.

adamstankiewicz:
> Just curious if we are already using `zustand` anywhere? I understand that our preference is to use `react-query` as much as possible but if we need to use some client side state management library, has the decision to use `zustand` been finalized or are there any other state management libraries still under consideration?

For client side state management, the preference is to rely on vanilla React as much as possible (e.g., `useState`, `useContext`, `useReducer`, etc.). Given much of the data we have in Redux / React Context throughout this MFE today is actually API-derived server state and should be in React Query (including most LPR data), the majority of remaining, true client-only state management remaining is fairly simple and largely doable with vanilla React state management alone (i.e., no addtl libraries).

That said, a React context-only based approach to share client state across multiple components without prop drilling can have its limitations in certain use cases due to component re-rendering concerns (e.g., multi-step forms, steppers, etc.); in those specific scenarios where re-rendering is a concern using a React context-only approach, alternative state management libraries may be considered, though we haven't really found a use case yet where an alternative state management library is needed beyond relying on vanilla React state management/hooks/context for client-only state in this MFE or the Enterprise Learner Portal MFE.

The one spot we have started to experiment with `zustand` within Enterprise is our new self-service checkout MFE ([GitHub](https://github.com/edx/frontend-app-enterprise-checkout)), though it's very much still WIP/prototype ([PR](https://github.com/edx/frontend-app-enterprise-checkout/pull/1)).

There's been no official decision made around `zustand` specifically outside of that one MFE, beyond the high-level decision to rely on client-only state management with vanilla React (e.g., state + context) vs. libraries, where possible. Here's how I'm thinking about it:
* server state (API-derived data) = React Query
* simple client state (client-only data) = native React (hooks, context)
* complex client state (client-only data), shared across multiple components (e.g., multi-step form) = `zustand` (_not_ a global store like Redux typically is, and without the boilerplate of Redux or even Redux Toolkit)

If Redux is still preferred by folks within Enterprise for these specific cases for client-only state where vanilla React does not suffice (e.g., re-rendering concerns), I'd recommend we discuss/align on a direction in the recurring Enterprise Solution Review meeting. If we were to continue using Redux for these more complex use cases (which shouldn't be all that common, in practice), we definitetly should avoid continuing the current paradigm of legacy Redux and instead migrate to using Redux Toolkit ([docs](https://redux-toolkit.js.org/)). That said, in my early discovery thus far, `zustand` has significantly less boilerplate than even Redux Toolkit, making it easier to understand & develop with.

src/components/AdminV2/_Admin.scss
```diff
@@ -0,0 +1,34 @@
+.learner-progress-report {
+  .table-title {
+    display: inline-block;
+    vertical-align: middle;
+  }
+
+  .reset {
+    margin-top: -5px;
```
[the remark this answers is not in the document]
omar-sarfraz: I believe this class is not being used. I'll verify and remove it.

src/components/AdminV2/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,144 @@
...
+  onClose,
+  isLoading,
+  error,
+  data,
+}) => {
+  const messages = defineMessages({
+    errorMessage: {
+      id: 'adminPortal.analyticsCardText.errorMessage',
+      defaultMessage: 'We encountered an issue while fetching analytics data. Kindly try again later or contact support for assistance. (Error: {error_message})',
+      description: 'Message shown to the user in case of error returned byt analytics API.',
```
jajjibhai008, line 25: Typo here `byt`

src/components/AdminV2/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,144 @@
...
+      values: { error_message: error?.message },
+    },
+    noContentErrorMessage: {
+      id: 'adminPortal.analyticsCardText.noContentErrorMessage',
+      defaultMessage: 'Analytics not found.',
+      description: 'Message shown to the user in case of empty response returned byt analytics API.',
```
jajjibhai008, line 31: Typo here `byt`

src/components/AdminV2/AIAnalyticsSummary.jsx
```diff
@@ -0,0 +1,144 @@
...
+          onClick={() => {
+            showSummarizeCard(true);
+            hideTrackProgressCard(true);
+            sendEnterpriseTrackEvent(enterpriseId, SUMMARIZE_ANALYTICS_CLICK_SEGMENT_EVENT_NAME);
+          }}
+          data-testid="summarize-analytics"
+        >
+          <>
+            <AutoFixHigh className="mr-2" />
+            <FormattedMessage id="adminPortal.v2.summarizeAnalytics" defaultMessage="Summarize analytics" />
```
jajjibhai008, line 93: You should addd some description as well.

src/components/AdminV2/AdminCards.jsx
```diff
@@ -0,0 +1,156 @@
...
+import PropTypes from 'prop-types';
+
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import {
+  Award, Check, Groups, RemoveRedEye,
+} from '@openedx/paragon/icons';
+
+import NumberCard from '../NumberCard';
+
+class AdminCards extends React.Component {
```
jajjibhai008, line 11: [Suggestion] Can't we change this class-based component to a functional component here?

src/components/AdminV2/AdminCardsSkeleton.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Skeleton } from '@openedx/paragon';
+
+const AdminCardsSkeleton = () => (
```
jajjibhai008, line 4: As per my knowledge, whenever you use a Card component from Paragon, you don't need to create a separate Skeleton for that, because when you provide the isLoading property to that card component, it will handle the skeleton accordingly. You need to confirm this behavior.

src/components/AdminV2/AdminCards.jsx
omar-sarfraz: I tried to keep the changes to minimum and only changed the components which were required to work with `dnd-kit`. We could change this into a functional component but I believe we should write another improvement ticket for this because there are other components like `AdminSearchForm` which needs to be converted to a functional component for consistency.

src/components/AdminV2/AdminCardsSkeleton.jsx
omar-sarfraz: In this implementation, we are not using the card component from paragon. This skeleton is used because we are creating a custom `AdminCards` component. @saleem-latif is working on a [ticket](https://2u-internal.atlassian.net/browse/ENT-10281) to replace `AdminCards` with the paragon's card component. We can remove this skeleton after that work is complete.

src/components/AdminV2/index.jsx
```diff
@@ -0,0 +1,567 @@
...
+import EnterpriseDataApiService from '../../data/services/EnterpriseDataApiService';
+import { withLocation, withParams } from '../../hoc';
+import BudgetExpiryAlertAndModal from '../BudgetExpiryAlertAndModal';
+import LearnerReport from './LearnerReport';
+import SortableItem from './SortableItem';
+import { getFromLocalStorage, saveToLocalStorage } from '../../utils';
+import SubscriptionModal from './SubscriptionModal';
+import { SubscriptionData } from '../subscriptions';
+
+const DISABLE_DRAG_AND_DROP = true;
```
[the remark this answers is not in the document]
omar-sarfraz: We need to disable drag and drop temporarily because the current ticket mainly focuses on design changes and we don't want to include drag and drop functionality just yet. I believe in the near future, the drag and drop will be enabled for all customers as part of Analytics revamp. That's why i just added a temporary variable which can be removed in the future. Do you believe adding a feature flag would be appropriate for this scenario?

jajjibhai008 (approved):
LGTM 👍

mahamakifdar19 (approved):

src/components/AdminV2/index.jsx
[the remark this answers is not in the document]
omar-sarfraz: @mahamakifdar19 You're right. It would definitely make things easier. Thanks for pointing it out. I'll implement this.

PR #1485, opened 2025-04-17
fix: remove unicode characters from license assignment

brobro10000:
Parses emails for any Unicode characters, and removes them from the email. This typically occurs for customers who may have RTL formatting when copy pasting emails assign. 

Parses out the unicode using the built in regular expression library.

# For all changes


src/data/validation/email.js
```diff
@@ -68,7 +78,7 @@ const validateEmailAddresses = (emails) => {
     return result;
   }
   emails.forEach((email, index) => {
-    const sanitizedEmail = email.trim();
+    const sanitizedEmail = sanitizeEmail(email);
```
adamstankiewicz, line 81: [consideration] I wonder if it's worth also lowercasing the emails?

src/data/validation/email.js
```diff
@@ -50,6 +55,11 @@ const validateEmailTemplateFields = (formData, templateKey, isSubjectRequired =
   return errorsDict;
 };
 
+const sanitizeEmail = (email) => {
+  // Strip bidirectional characters. This is mostly to make RTL emails compatible
```
adamstankiewicz, line 59: nit: perhaps amend the comment to include a mention that this only strips _invisible_ characters.

src/data/validation/email.test.js
```diff
@@ -37,6 +38,45 @@ describe('email validation', () => {
...
+        '\u202Aalice@example.org', // LRE at start of address
+        'kate.jones@domain\u202B.com', // RLE before '.' in domain
+        'roger+\u202Csales@company.com', // PDF after '+' in local part
+        'chris@exa\u202Dmple.com', // LRO in domain name
+        'mary@contoso.com\u202E', // RLO at end of address
+        'anna\u2066.smith@mysite.com', // LRI before '.' in local part
+        'info@support.\u2067example.com', // RLI after subdomain '.'
+        'mark@example.c\u2068om', // FSI inside TLD
+        'tom.\u2069adams@example.com', // PDI after '.' in local part
+      ];
```
pwnage101, line 56: Maybe add a test case for multiple unicode format characters in one email. And another test with just unicode format characters and nothing else.

src/data/validation/email.test.js
```diff
@@ -37,6 +38,45 @@ describe('email validation', () => {
...
+
+      // Verify that the Unicode does exist in the email
+      const formatCharRegex = XRegExp('\\p{Cf}', 'g');
+      unicodeEmails.forEach((unicodeEmail) => {
+        const hasUnicodeFormat = XRegExp.test(unicodeEmail, formatCharRegex);
+        expect(hasUnicodeFormat).toBe(true);
+      });
+
+      const expectedResult = { ...baseResults };
+      expectedResult.validEmails = unicodeEmails.map(email => XRegExp.replace(email, formatCharRegex, '').trim());
```
pwnage101, line 66: I just wanted to call out this general test pattern which is self-fulfilling. IMO, the code you are trying to test should be off-limits from being in the test.

In this case I'd define `unicodeEmails` above as a list of lists, one input and one expected output.

src/data/validation/email.test.js
```diff
@@ -1,14 +1,15 @@
 import _ from 'lodash';
+import XRegExp from 'xregexp';
```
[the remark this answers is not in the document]
adamstankiewicz: [related consideration] I wonder if it's worth code splitting the `regex` import with `lazy`, so it only gets downloaded by the user just-in-time when it's needed.

src/data/validation/email.test.js
brobro10000: Corrected with hard coded output and an individualized check at a char code level whether the unicode value exist 👍🏽

src/data/validation/email.test.js
brobro10000: Updated 👍🏽

src/data/validation/email.test.js
```diff
@@ -1,16 +1,25 @@
...
   validateEmailTemplateFields,
   validateEmailTemplateForm,
-  validateEmailAddrTemplateForm,
-  returnValidatedEmails,
 } from './email';
-import { EMAIL_ADDRESS_TEXT_FORM_DATA, EMAIL_ADDRESS_CSV_FORM_DATA } from '../constants/addUsers';
+import { EMAIL_ADDRESS_CSV_FORM_DATA, EMAIL_ADDRESS_TEXT_FORM_DATA } from '../constants/addUsers';
 import { EMAIL_TEMPLATE_SUBJECT_KEY, OFFER_ASSIGNMENT_EMAIL_SUBJECT_LIMIT } from '../constants/emailTemplate';
 
+const hasFormatChar = (str) => [...str].some(char => {
```
brobro10000, line 14: This function was added to prevent "self fulfilling" test cases where using the same regular expression to validate the regular expression is not actually testing the regular expression

src/data/validation/email.test.js
```diff
@@ -1,16 +1,25 @@
 import _ from 'lodash';
```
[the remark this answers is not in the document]
brobro10000: A quick search indicates 18 instances of the import in non-test files as well. ([src](https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-admin-portal%20%2Fimport.*%3Flodash%27%2F&type=code))

I can file a ticket to backlog to resolve this in the future.

PR #1502, opened 2025-05-05
feat: Enabled group functionality for groups.

saleem-latif:
__Jira Ticket:__ [ENT-10130](https://2u-internal.atlassian.net/browse/ENT-10130)

__Description:__
Add a new single select drop down menu to display a list of groups for admin analytics, this will be enabled for enrollments, engagements and completions data only and will not have any affect on skills and leaderboard data. The list can be fetched from enterprise API endpoint [edx-enterprise/enterprise/api/v1/urls.py at master · openedx/edx-enterprise](https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/urls.py#L85) When a user make a selection then we need to pass in the selected group’s uuid via group_uuid named parameter in API calls to the backend.

# Screenshot
Screenshot 2025-05-05 at 12 09 34 PM

### DIFF
--- a/src/components/AdvanceAnalyticsV2/AnalyticsV2Page.jsx
+++ b/src/components/AdvanceAnalyticsV2/AnalyticsV2Page.jsx
@@ -15,15 +15,19 @@ import Leaderboard from './tabs/Leaderboard';
 import Skills from './tabs/Skills';
 import { useEnterpriseAnalyticsAggregatesData } from './data/hooks';
 import { GRANULARITY, CALCULATION } from './data/constants';
+import { DEFAULT_GROUP } from './constants';
 import { formatTimestamp } from '../../utils';
+import { useAllFlexEnterpriseGroups } from '../learner-credit-management/data';
 
 const AnalyticsV2Page = ({ enterpriseId }) => {
   const [activeTab, setActiveTab] = useState('enrollments');
   const [granularity, setGranularity] = useState(GRANULARITY.WEEKLY);
   const [calculation, setCalculation] = useState('Total');
+  const [groupUUID, setGroupUUID] = useState('');
   const [startDate, setStartDate] = useState();
   const [endDate, setEndDate] = useState();
   const intl = useIntl();
+  const { data: groups, isLoading: isGroupsLoading } = useAllFlexEnterpriseGroups(enterpriseId);
 
   const PAGE_TITLE = intl.formatMessage({
     id: 'analytics.page.title',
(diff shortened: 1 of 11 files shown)


src/components/AdvanceAnalyticsV2/AnalyticsV2Page.jsx
```diff
@@ -15,15 +15,22 @@ import Leaderboard from './tabs/Leaderboard';
...
   const [activeTab, setActiveTab] = useState('enrollments');
   const [granularity, setGranularity] = useState(GRANULARITY.WEEKLY);
   const [calculation, setCalculation] = useState('Total');
+  const [groupUUID, setGroupUUID] = useState('');
   const [startDate, setStartDate] = useState();
   const [endDate, setEndDate] = useState();
   const intl = useIntl();
+  const { data: groups, isLoading: isGroupsLoading } = useAllFlexEnterpriseGroups(enterpriseId);
+
+  console.log('Groups Data: ', groups, isGroupsLoading);
```
mahamakifdar19, line 32: Is this console statement intentional?

src/components/AdvanceAnalyticsV2/constants.js
```diff
@@ -0,0 +1,2 @@
+// Default group for the analytics. For now, it will be an empty string.
+export const DEFAULT_GROUP = '';
```
mahamakifdar19, line 2: If it's meant to be an empty string, perhaps we should explicitly set value='' on line 118?

mahamakifdar19 (approved):

src/components/AdvanceAnalyticsV2/AnalyticsV2Page.jsx
```diff
@@ -15,15 +15,22 @@ import Leaderboard from './tabs/Leaderboard';
...
 import { GRANULARITY, CALCULATION } from './data/constants';
+import { DEFAULT_GROUP } from './constants';
 import { formatTimestamp } from '../../utils';
+import { useAllFlexEnterpriseGroups } from '../learner-credit-management/data';
 
 const AnalyticsV2Page = ({ enterpriseId }) => {
   const [activeTab, setActiveTab] = useState('enrollments');
   const [granularity, setGranularity] = useState(GRANULARITY.WEEKLY);
   const [calculation, setCalculation] = useState('Total');
+  const [groupUUID, setGroupUUID] = useState('');
```
muhammad-ammar, line 26: Initially we will fetch data regardless of groups, correct?

saleem-latif: Yes, that is correct. empty value "" means no group selected.

PR #1509, opened 2025-05-19
feat: implement ESM friendly lodash, lodash-es

brobro10000:
Uses the `lodash-es` to replace `lodash` with a es-module friendly version of the package. 

# For all changes


__mocks__/lodash-es.js
```diff
@@ -0,0 +1,25 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
adamstankiewicz, line 1: I think this file can be removed. See comment in `jest.config.js`.

jest.config.js
```diff
@@ -1,8 +1,12 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 const { createConfig } = require('@openedx/frontend-build');
+const path = require('path');
 
 module.exports = createConfig('jest', {
   setupFiles: [
     '<rootDir>/src/setupTest.js',
   ],
+  moduleNameMapper: {
```
adamstankiewicz, line 9: In `src/data/validation/email.js` as an isolated example locally, on `master` after installing `lodash-es`, I replaced the original `lodash` usage with `lodash-es` to observe the test failure around the `SyntaxError` with `export`. Then, modifying the `jest.config.js` made the test pass (without any other changes, like the `__mocks__` file):

```js
const config = createConfig('jest', {
  setupFiles: [
    '<rootDir>/src/setupTest.js',
  ],
});

config.transformIgnorePatterns = ['/node_modules/(?!(lodash-es|@(open)?edx)/)'];

module.exports = config;
```

babel.config.js
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { createConfig } = require('@openedx/frontend-build');
+
+const babelConfig = createConfig('babel-typescript', {});
+
+babelConfig.presets[0] = ['@babel/preset-env', { targets: { node: 'current' }, modules: 'auto' }]
```
adamstankiewicz, line 6: I believe we can remove this file as:
1. The tests seem to pass fine by just including `lodash-es` in the existing `transformIgnorePatterns` regex in `jest.config.js`.
2. I feel we can continue to rely on the default `@babel/preset-env` without overriding the base options (as used by `@openedx/frontend-build` itself).
3. I believe the babel-related config that needs to change lives in the webpack config files, with its usage of `babel-loader` ([example](https://github.com/openedx/frontend-build/blob/e3856b0fab23dd271e6f018103f94bae36080421/config/webpack.dev.config.js#L95C9-L95C50)).

To expand on (3), I believe we'll need to:

1. Update `webpack.dev-stage.config.js` to override the default `babel-loader`'s `exclude` option, e.g.:
    * `config.module.rules[0].exclude = /node_modules\/(?!(lodash-es|@(open)?edx))/;`
2. Repeat in new files:
    * `webpack.dev.config.js`
        * Also ensure using `createConfig('dev')`
    * `webpack.prod.config.js`
        * Also ensure using `createConfig('webpack-prod')`

package.json
```diff
@@ -115,6 +114,7 @@
     "identity-obj-proxy": "3.0.0",
     "jest-canvas-mock": "^2.4.0",
     "jest-localstorage-mock": "^2.4.22",
+    "lodash": "^4.17.21",
```
adamstankiewicz, line 117: [inform] See other related comments, but I believe we can fully remove this `lodash` module from the `package.json`, given it seems the tests do pass without needing its usage in `__mocks__`.

src/components/AdvanceAnalyticsV2/data/utils.js
```diff
@@ -90,16 +89,24 @@ e.g. 3 would mean 3-day moving average.
 */
 const calculateMovingAverage = (timeSeriesData, countKey, period) => {
```
adamstankiewicz, line 90: [curious/consideration] Do you feel there are tests adequately covering the logic in this function? I'm still somewhat concerned about including these changes in a PR primarily about migrating to `lodash-es` given it's a refactor of a reasonably non-trival logic code path.

For example, iff there's a bug in this refactored logic, it'd be nice to be able to revert/rollback just these changes vs. all of the `lodash-es` changes. Maybe split these changes out into their own distinct PR and tag some folks from markhors on it given they primarily own this area and may have additional insights?

brobro10000: I can revert the changes related to this function back to its original usage to instead leverage `.reduce` to handle summation of the moving average vs `sum` and open a seperate PR/ticket that includes this function optimization.

The function was very indicative of a leetcode-esque problem with an already optimized solution so I felt compelled to correct it on the spot.

babel.config.js
brobro10000: Above implementation finally allowed `lodash-es` to not break jest testing. I wonder if we do see an overall benefit of including this package over `lodash`, should an effort be made (over the long term) to migrate to `lodash-es` and include the package and related config changes in the base frontend-build repo config as to not override it.

adamstankiewicz (approved):

PR #1531, opened 2025-06-05
fix: convert cents to USD in requests table

jajjibhai008:
# For all changes

### DIFF
--- a/src/components/learner-credit-management/requests-tab/AmountCell.jsx
+++ b/src/components/learner-credit-management/requests-tab/AmountCell.jsx
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import { formatPrice } from '../data';
 
 const AmountCell = ({ row }) => {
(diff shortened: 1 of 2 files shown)


zamanafzal (approved):

src/components/learner-credit-management/requests-tab/AmountCell.jsx
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import { formatPrice } from '../data';
 
 const AmountCell = ({ row }) => {
-  const formattedContentPrice = formatPrice(row.original.amount);
+  const formattedContentPrice = formatPrice(row.original.amount / 100);
```
sameenfatima78, line 5: shouldn't this also be replaced with `course_price?`

jajjibhai008: Actually in the DataTable the cell name is Amount so i keep it amount here instead of course_price to avoid any confusions.

PR #1581, opened 2025-07-11
feat: add onboarding flows for settings and reporting

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-10360)

https://github.com/user-attachments/assets/7c5e5ba0-a578-43b1-86c4-e7db0b7d0cc6

Customer with Reporting Enabled: https://localhost.stage.edx.org:1991/exec-ed-2u-integration-qa/admin/learners
Customer with Reporting Disabled: https://localhost.stage.edx.org:1991/alc-general/admin/learners

# For all changes

### DIFF
--- a/src/components/ProductTours/AdminOnboardingTours/CustomizeReportsFlow.tsx
+++ b/src/components/ProductTours/AdminOnboardingTours/CustomizeReportsFlow.tsx
@@ -1,0 +1,24 @@
+import { useIntl } from '@edx/frontend-platform/i18n';
+import { CUSTOMIZE_REPORTS_SIDEBAR } from './constants';
+import messages from './messages';
+import { TourStep } from '../types';
+
+interface CreateTourFlowsProps {
+  handleEndTour: (endEventName: string, flowUuid?: string) => void;
+}
+
+const CustomizeReportsFlow = ({
+  handleEndTour,
+}: CreateTourFlowsProps): Array<TourStep> => {
+  const intl = useIntl();
+
+  return [
+    {
+      target: `#${CUSTOMIZE_REPORTS_SIDEBAR}`,
+      placement: 'right',
+      body: intl.formatMessage(messages.viewCustomizeReports),
+      onAdvance: handleEndTour,
+    }];
+};
+
+export default CustomizeReportsFlow;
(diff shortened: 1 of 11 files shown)


src/components/ProductTours/TourCollapsible.tsx
```diff
@@ -80,6 +98,17 @@ const TourCollapsible: FC<Props> = (
     reopenTour(adminUuid);
   };
 
+  const hiddenSteps = useMemo(() => {
```
jesperhodge, line 101: You correctly exclude tabs here that are not enabled.
It would be great to make clear why the steps are excluded, as well.

I was also thinking it would be nice to simplify the logic a bit.
In the current implementation, you have a set of steps, then define a set of hidden steps, depending on props, cache this, and then subtract the list of hidden steps from the list of steps while you are already rendering the component.

Why not just have a `useEffect` depending on the two props that takes the original list with a `.filter` function and just removes the steps if necessary? Save it in a new state and then render that modified list.

That would create a clean separation between data operations and rendering, look super simple and clear, and avoid some extra steps.

src/components/ProductTours/AdminOnboardingTours/CustomizeReportsFlow.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const CustomizeReportsFlow = ({
+  handleEndTour,
+}: CreateTourFlowsProps): Array<TourStep> => {
+  const intl = useIntl();
+
+  return [
+    {
+      target: `#${CUSTOMIZE_REPORTS_SIDEBAR}`,
```
jesperhodge, line 17: `title` is missing

src/components/ProductTours/AdminOnboardingTours/SetUpPreferencesFlow.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const SetUpPreferencesFlow = ({
+  handleEndTour,
+}: CreateTourFlowsProps): Array<TourStep> => {
+  const intl = useIntl();
+
+  return [
+    {
+      target: `#${TOUR_TARGETS.SETTINGS_SIDEBAR}`,
```
jesperhodge, line 17: `title` is missing

jesperhodge (changes requested):

src/components/ProductTours/AdminOnboardingTours/CustomizeReportsFlow.tsx
marlonkeating: Good catch!

PR #1595, opened 2025-07-17
feat: added custom date range option

irfanuddinahmad:
# For all changes


src/components/AdvanceAnalyticsV2.0/AnalyticsFilters.jsx
```diff
@@ -29,14 +29,17 @@ const AnalyticsFilters = ({
   const intl = useIntl();
   const [collapsed, setCollapsed] = useState(false);
   const isProgressOrOutcomesTab = activeTab === 'progress' || activeTab === 'outcomes';
+  const [dateRangeValue, setDateRangeValue] = useState(DATE_RANGE.LAST_90_DAYS);
```
mahamakifdar19, line 32: Since we're using a context provider to persist filter values across all tabs, it's recommended to define `setDateRangeValue` in the `AnalyticsPage` component and also set the default value (90 days) there. The method should then be passed down as a prop, similar to how `setStartDate` and other handlers are handled.

src/components/AdvanceAnalyticsV2.0/AnalyticsFilters.jsx
```diff
@@ -48,6 +51,24 @@ const AnalyticsFilters = ({
...
+    setStartDate(selectedDate);
+    setDateRangeValue(DATE_RANGE.CUSTOM);
+  };
+
+  const handleEndDateChange = (selectedDate) => {
+    setEndDate(selectedDate);
+    setDateRangeValue(DATE_RANGE.CUSTOM);
+  };
+
+  const getInitialStartDate = () => {
```
mahamakifdar19, line 64: Consider turning it into a reusable utility method

src/components/AdvanceAnalyticsV2.0/AnalyticsFilters.jsx
mahamakifdar19, line 32: `dateRangeValue` should be passed down as a prop, following the same pattern as granularity and calculation.

src/components/AdvanceAnalyticsV2.0/AnalyticsFilters.jsx
```diff
@@ -132,9 +161,9 @@ const AnalyticsFilters = ({
                 <Form.Control
                   controlClassName="font-weight-normal analytics-filter-form-controls rounded-0"
                   type="date"
-                  value={startDate || data?.minEnrollmentDate || ''}
+                  value={startDate || getInitialStartDate() || data?.minEnrollmentDate || ''}
```
mahamakifdar19, line 164: I don't think `data?.minEnrollmentDate` is needed here anymore since we now have a default startDate which is last 90 days.

Also, both s`tartDate` and `endDate` need to be updated across all relevant charts and tables. By “everywhere,” I’m referring to components where we're currently using `data?.minEnrollmentDate` as the default start date — for things like CSV downloads and other data displays. Those instances will need to be updated as well to reflect to new default start date (Last 90 days)

References:
- https://github.com/openedx/frontend-app-admin-portal/blob/02c09435d4a8348fa70bbdc2d460842540c5adc0/src/components/AdvanceAnalyticsV2.0/tabs/Engagements.jsx#L169

- https://github.com/openedx/frontend-app-admin-portal/blob/02c09435d4a8348fa70bbdc2d460842540c5adc0/src/components/AdvanceAnalyticsV2.0/tabs/Engagements.jsx#L204

src/components/AdvanceAnalyticsV2.0/AnalyticsFilters.jsx
irfanuddinahmad: No need to pass setDateRangeValue as a prop, since its function is to use the setStartDate and setEndDate props.

irfanuddinahmad: No need.

mahamakifdar19: @irfanuddinahmad Did you get a chance to test it locally? The date range filter value doesn’t persist when switching tabs.

mahamakifdar19 (commented):
LGTM!

mahamakifdar19 (approved):

PR #1627, opened 2025-07-24
feat: adds assignment flow

katrinan029:
# For all changes

https://2u-internal.atlassian.net/browse/ENT-10521

https://github.com/user-attachments/assets/fa8c23cb-1380-4dba-b330-a05ceaffa451


marlonkeating (changes requested):
Seeing some issues when

Orphaned highlight around button when navigating into budget
https://github.com/user-attachments/assets/d366a949-66e4-4817-a9c0-11d0b49a7de9

Not scrolling far enough when transitioning from step 6-7
https://github.com/user-attachments/assets/902b2acc-0afd-4b4d-b79c-4f39ab5e08c1

src/components/ProductTours/AdminOnboardingTours/AdminOnboardingTours.tsx
```diff
@@ -56,13 +83,51 @@ const AdminOnboardingTours: FC<AdminOnboardingToursProps> = ({
     }
   }, [targetSelector]);
 
+  // Handle target setting for both page transitions and step changes
+  // eslint-disable-next-line consistent-return
   useEffect(() => {
```
marlonkeating, line 88: I'm still seeing a lot of issues with the way we're handling the problem of transitioning the tour focus between components that take time to load/render:
- Logic split between multiple files (AdminOnboardingTours.tsx / CheckpointOverlay.tsx)
- Branching logic on specific named steps, outside of where the steps are defined
- Hard-to-follow callback logic between different code branches with nested `setTimeouts` (and divergent `useEffect` return values which necessitates an `eslint-disable-next-line consistent-return`)
- Setting flat timeouts to wait for components to render, when the render time may be shorter or longer than we expect.

Overall, I think this is a case of when a cluster of code smells indicates we're attempting to tackle a problem at the wrong level of abstraction.  Ideally, Paragon's `ProductTour` component would support page/state transitions between checkpoints.  For the present, we may want to consider what is the cleanest and most self-contained 'hacky' way we can accommodate these transitions while waiting for support to be added at the Paragon level.

A simple way might be doing a poll/wait where we check every 100ms if the next target has loaded before we switch focus to it.  Or use something like a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) if we want to get fancy.

